PR #15, opened 2020-07-04
Fixing module startup behavior

mmarkwort:


Toxantron:
The more I think about this, I am questioning the automatic restart in general. In the past 8 years I can't remember a valid case where a module fails to start and without any human interference could be restarted roughly 15ms later. Einstein once said:
> Insanity is doing the same thing over and over and expecting different results.
With that in mind I vote to remove that whole feature including configuration etc.

@dbeuchler and @slawor, any thoughts?

dbeuchler:
Same for me, we started to remove this feature while
we decided to remove module stops in IParallelOperations.

mmarkwort:
I created a new issue https://github.com/PHOENIXCONTACT/MORYX-Platform/issues/20 that addresses the automatic restart feature

src/Moryx.Runtime.Kernel/Modules/ComponentContracts/IModuleFailureStrategy.cs
```diff
@@ -10,9 +10,5 @@ namespace Moryx.Runtime.Kernel
     /// </summary>
     internal interface IModuleFailureStrategy
```
Toxantron, line 11: I think the entire interface is obsolote now

src/Moryx.Runtime.Kernel/Modules/Components/FailureStrategy.cs
```diff
@@ -1,9 +1,6 @@
 // Copyright (c) 2020, Phoenix Contact GmbH & Co. KG
 // Licensed under the Apache License, Version 2.0
 
-using Moryx.Modules;
-using Moryx.Runtime.Modules;
-
 namespace Moryx.Runtime.Kernel
```
Toxantron, line 4: Same here

src/Moryx.Runtime.Kernel/Modules/ComponentContracts/IModuleFailureStrategy.cs
mmarkwort: Ok, I will remove it.

Toxantron: Thanks

mmarkwort: Done

src/Moryx.Runtime.Kernel/Modules/Components/FailureStrategy.cs
mmarkwort: Done

src/Moryx.Runtime.Kernel/Modules/Components/ModuleStarter.cs
```diff
@@ -62,7 +62,11 @@ private void StartModule(IServerModule module)
             }
             else
             {
-                ThreadPool.QueueUserWorkItem(ExecuteModuleStart, module);
+                // Don't try to start modules which initialization has been failed
+                if (module.State != ServerModuleState.Failure)
+                {
+                    ThreadPool.QueueUserWorkItem(ExecuteModuleStart, module);
```
Toxantron, line 68: This seams redundant, but we will address it in #23.

Toxantron:
Thank you for your contribution

PR #16, opened 2020-07-06
Simplify Moryx.Model

Toxantron:


dbeuchler:
Will be reopened for appropiate description

PR #28, opened 2020-08-06
Simplification of Moryx.Model usage with EntityFramework

dbeuchler:
This merge request is used to simplify the complete entity framework usage in moryx. The old implementation strictly hides the `DbContext`. The new implementation should provide the `DbContext` more open and the `UnitOfWork Pattern`

- Removed inheritance of `UnitOfWorkFachtory`
- `IUnitOfWork` provides also the context now
- No naming of data model is required anymore (old constants)
- `ModelSetups` are using `DbContext` now
- Effort InMemory can be used out of the box 
- Foreign model assemblies can be used now with expoded configuration API
- Added several extensions to `DbContext`, `DbSet`, `IQueryable<Entity>`
- Reworked complete documentation
- Adjusted all tests for new APIs
- Old API is kept as it is with some minor adjustments (e.g. `uow.SaveChanges` instead of `uow.Save`)
- Preconfigured DbContext can be created by `IContextFactory`

/cc @mmarkwort may be interested for you because you were the main developer of the Code-First approach


mmarkwort:
Ok, I'll check that

PR #42, opened 2020-09-07
Exclude modules with missing (transient) dependencies

### LINKED ISSUE #22: Do not crash runtime if facade dependency is missing
Currently the runtime crashes if a facade dependency is missing. 

We have two options:

1. The module with missing dependencies should go into the `Failure` state and should never be startable
2. Write log entry and ignore entire module, because it is an design time problem

**Additional context**

````txt
Moryx.Runtime.Kernel.MissingFacadeException: Found no module hosting a facade of type IFoo which was expected by SomeModule.Property
   bei Moryx.Runtime.Kernel.ModuleDependencyManager.FillFacade(Type propType, ImportingProperty importingProperty, IServerModule service)
   bei Moryx.Runtime.Kernel.ModuleDependencyManager.FillDependencies(IEnumerable`1 services)
   bei Moryx.Runtime.Kernel.ModuleDependencyManager.BuildDependencyTree()
   bei Moryx.Runtime.Kernel.ModuleManager.Initialize()
   bei Moryx.Runtime.Kernel.HeartOfGold.<>c.<SetupEnvironment>b__6_0(IInitializable i)
   bei Moryx.Tools.EnumerableExtensions.ForEach[T](IEnumerable`1 enumeration, Action`1 action)
   bei Moryx.Runtime.Kernel.HeartOfGold.SetupEnvironment(EnvironmentInfo env)
   bei Moryx.Runtime.Kernel.HeartOfGold.LoadEnvironment(IRunMode& runMode)
````

### PR DESCRIPTION
**Pull request by Toxantron**
Instead of throwing an exception for a missing facade provide, the dependency is now flagged as missing and the module is no longer available. However the removed module might have provided facades required by other modules and might have been assigned to any facade collections. Non of the modules referencing the broken module can start and are removed

I also noticed that `IsStartDependency` was not supported for collections, so I fixed it.

During the fix I also did a minor refactoring and clean-up of the module manager to cope with the changes. Especially the direct usages of `AllModules` had to be replaced with the reduced collection. I also removed unnecessary fields and overly complex code.

Closes #22.

### CODE REVIEW

File: src/Moryx.Runtime.Kernel/Modules/Components/ModuleDependencyManager.cs [hunk 1]
```diff
@@ -85,130 +81,150 @@ private void BuildFacadeCache(IEnumerable<IServerModule> allServices)
...
+            if (match != null)
             {
-                if (match != null)
-                    importingProperty.Property.SetValue(service, match);
-                return;
+                importingProperty.Property.SetValue(service, match);
+            }
+            else if (importingProperty.Attribute.IsOptional)
+            {
+                _logger.Log(LogLevel.Warning, "Found no plugin hosting a facade of type {0} which is referenced by {1}.{2}",
```
dbeuchler, line 141: It is an old log message but would you change this to "module" insteaf of "plugin"?

src/Moryx.Runtime.Kernel/Modules/Components/ModuleDependencyManager.cs
```diff
@@ -85,130 +81,150 @@ private void BuildFacadeCache(IEnumerable<IServerModule> allServices)
...
+                importingProperty.Missing = true;
             }
-
-            _logger.Log(LogLevel.Error, "Found no plugin hosting a facade of type {0} which is required by {1}.{2}",
-                propType.Name, service.Name, importingProperty.Property);
-            throw new MissingFacadeException(service.Name, importingProperty.Property.Name, propType);
         }
 
-        private IModuleDependency Convert(IServerModule plugin, IDictionary<IServerModule, IModuleDependency> cache)
+        private IReadOnlyList<PluginDependencyBranch> Convert(IReadOnlyList<IServerModule> availableModules, IDictionary<object, IServerModule> facadeProviders)
```
dbeuchler, line 152: Same here: ModuleDependencyBranch?

dbeuchler (approved):
The rest looks great to me!

PR #43, opened 2020-09-11
Fix spelling and include RangeAttribute in validation

Toxantron:
I received feedback about a spelling error in EntryValidation and the idea, to make the value more flexible and also include range validation values.

### DIFF
--- a/src/Moryx/Serialization/DefaultSerialization.cs
+++ b/src/Moryx/Serialization/DefaultSerialization.cs
@@ -93,17 +93,25 @@ namespace Moryx.Serialization
             // Iterate over attributes reading all validation rules
             foreach (var attribute in validationAttributes)
             {
-                if (attribute is MinLengthAttribute)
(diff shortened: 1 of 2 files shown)


dbeuchler (approved):
The renaming must also be fixed within the maintenance web.

Toxantron:
> The renaming must also be fixed within the maintenance web.

I checked and the MaintenanceWeb does currently not include validation AFAIK.

dbeuchler:
> > The renaming must also be fixed within the maintenance web.
> 
> I checked and the MaintenanceWeb does currently not include validation AFAIK.

I think you missed the deserialization: https://github.com/PHOENIXCONTACT/MORYX-MaintenanceWeb/blob/dev/src/Moryx.Runtime.Maintenance.Web.UI/src/modules/models/EntryValidation.ts

Toxantron:
Yes I did :-)

Issue #49, opened 2020-10-15
Supporting CORS in WebHttp


mmarkwort:
Currently the WebHttp host configuration does not support cors. So, requests from clients of other locations than server:port will be blocked.

Toxantron:
I might be wrong, but CORS is only relevant on the client, meaning it will reject responses without the header if script origin and call target don't match. I didn't have any problems and neither does the MaintenanceWeb, so could you elaborate where you had issues? I am not against the change, simply curious what happend.

mmarkwort:
I had some issues with this yesterday. Test it please with swagger or Postman. https://en.wikipedia.org/wiki/Cross-origin_resource_sharing

Toxantron:
Is it possible you were using the Swagger or Postman WebClient? That would explain your CORS issues.

But in case someone else wants to integrate our endpoints in their application, they would run into the same issue.

mmarkwort:
Yes, I write also an openapi.json for the products part and using a swagger tool within VS Code to test it.

Toxantron:
Closed by #50.

PR #50, opened 2020-10-15
Added cors support behavior

mmarkwort:
Solves #49


Toxantron:
Just one question: Is this by any chance code from Stackoverflow? If not, I agree adding CORS.

mmarkwort:
> 
> 
> Just one question: Is this by any chance code from Stackoverflow? If not, I agree adding CORS.

Unfortunately this code comes in Parts from Stackoverflow

Toxantron:
> Unfortunately this code comes in Parts from Stackoverflow

Because Stackoverflow failed to go through with licensing their snippets under MIT license, all of that is still **CC-BY-SA** and incompatible with our Apache License.

Toxantron:
I reduced it to pure CORS implementation, that should not violate any Copyright, as it is an almost empty implementation that only sets CORS headers and should fall below *threshold of originality*.

Toxantron (approved):

Issue #52, opened 2020-10-21
Include REST support in the platform


Toxantron:
**Is your feature request related to a problem? Please describe.**
We migrated the resource and product service and UI to REST, but had to use a couple of workarounds.

**Describe the solution you'd like**
Better REST support in the platform

**Describe alternatives you've considered**
...

**Additional context**
...

PR #58, opened 2020-10-27
Provide base class for web http connector

dbeuchler:


Toxantron (approved):

dbeuchler:
This is also a step for #52

PR #79, opened 2020-11-09
Doc cleanup before release

Toxantron:


docs/articles/Platform/Serialization/PossibleValues.md
```diff
@@ -117,4 +117,17 @@ public StrategyConfig ActiveStrategy { get; set; }
...
+
+This attribute generates possible values from given or derived types found at runtime, either as simple type name or full name. This attribute is useful when creating modules or components, that can be extended with derived types later.
+
+````cs
+[PossibleTypes(typeof(MyBaseType), UseFullName = true)]
+public string ConfiguredType { get; set; }
+
+[DataMember]
+[PossibleTypes(new []{ typeof(Type1), typeof(Type2) })]
+public List<PluginConfig> ActivePlugins { get; set; }
```
dbeuchler, line 132: The `PossibleTypesAttribute` returns a string and not an instance?

Toxantron: Thanks, that was a mistake. I changed it to list of `Type`. But yes, the attribute returns strings.

dbeuchler (approved):

PR #83, opened 2020-11-13
Usage of dotnet instead of msbuild and .net5.0 tfm

### LINKED ISSUE #82: Support .NET 5.0 target framework
After announcing the official release of .NET 5.0 we should also support the target framework. All packages which are targeting .netcoreapp or .netstandard should also target .net5.0

dbeuchler:
This PR contains two changes:

- Replaces all usages of `msbuild` and nearly all of `nuget` with the `dotnet` toolchain.
- Add .net 5.0 TFM to .netstandard and .netcoreapp projects

closes #82


.build/BuildToolkit.ps1
```diff
@@ -27,7 +27,6 @@ $NugetPackageArtifacts = "$ArtifactsDir\Packages";
 . "$PSScriptRoot\Output.ps1";
 
 # Define Tools
-$global:MSBuildCli = "msbuild.exe";
 $global:DotNetCli = "dotnet.exe";
 $global:NugetCli = "nuget.exe";
```
Toxantron, line 31: Do we still need `NugetCli`?

Toxantron (approved):
Looks fine, only wondering if we can eliminate nuget cli reference as well

.build/BuildToolkit.ps1
dbeuchler: yes, eg for install tools (which are not migrated to `dotnet tool`)

PR #93, opened 2021-01-13
Parse and use configured IPAddress in tcp listener

### LINKED ISSUE #89: TcpPortListener does not use configured IpAddress
**Describe the bug**
The `TcpPortListener` is not using the configured ip address of the `TcpListenerConnection`. Instead the static `Any` is used.

````cs
_tcpListener = new TcpListener(new IPEndPoint(IPAddress.Any, Port));
````

### PR DESCRIPTION
**Pull request by Toxantron**
This parse the IPAddress and properly checks conflicts only for registration overlap. This allows users to use different protocols for the same port on different network adapters or IP addresses.

Closes #89.

While doing this, I noticed that part registrations are never released, even when the listener shuts down and that all connections of all modules share the same listener instance. The first might have side effects, while the other can have performance impacts due to locks. I opened #92 for this.

### DIFF
--- a/src/Moryx/Communication/Sockets/Server/TcpServer.cs
+++ b/src/Moryx/Communication/Sockets/Server/TcpServer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading;
 
 namespace Moryx.Communication.Sockets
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Moryx/Communication/Sockets/Server/TcpServer.cs [hunk 1]
```diff
@@ -4,37 +4,68 @@
...
-    internal static class PortMap
+    internal class PortMap
     {
         /// <summary>
         /// Application global map of ports
         /// </summary>
-        private static readonly IDictionary<int, IMessageInterpreter> Map = new Dictionary<int, IMessageInterpreter>();
+        private static readonly List<PortMap> Registrations = new List<PortMap>();
+        
+        public IPAddress Address { get; }
```
dbeuchler, line 22: Can you also add summaries to the new porperties?

dbeuchler (changes requested):

dbeuchler (approved):

Toxantron:
Thanks to @dacky179 who found this, after we didn't notice for 5 years :-D

PR #90, opened 2021-01-05
Use s format specifier for crash log timestamp

dbeuchler:
The crashlog timestamp uses the 12h format without am/pm. We should treat it as bug and change it to the default "s" specifier:

````cs
//    s Format Specifier      de-DE Culture                      2008-10-01T17:04:32
//    s Format Specifier      en-US Culture                      2008-10-01T17:04:32
//    s Format Specifier      es-ES Culture                      2008-10-01T17:04:32
//    s Format Specifier      fr-FR Culture                      2008-10-01T17:04:32
````

PR #97, opened 2021-01-21
Return to a filename compatible format

Toxantron:
We accidentally introduced illegal characters in the timestamp in #90. Instead I restored the old format with proper 24h time.


Toxantron:
I created this PR against `main`, as I consider it a critical error that should be released as a hotfix.

PR #98, opened 2021-01-21
Include Hotfix and bumb version

Toxantron:
Include hotfix #97 and bumb version to `3.0.3`


dbeuchler:
Instead of creating two pull requests I merged the branch in both. In #97 you missed the version which I have added.

PR #101, opened 2021-01-26
Extended message validation

Toxantron:
When implementing another TCP based driver, we discovered the need to distinguish between an initial message (LogOn) and follow-up messages on the same socket. The logon could contain information relevant for assignment while follow-up messages require alternative validations.


dbeuchler:
For now this can be closed, right?

Issue #100, opened 2021-01-21
Set keep alive in TcpTransmission throws PlatformNotSupportedException on Linux/OSX


dbeuchler:
**Describe the bug**
The TcpTransmission throws a PlatformNotSupportedException when using IOControl to set keepAlive on non windows os.
That's because it's only supported on Windows. This IOControl code value is Windows-specific.

https://github.com/dotnet/runtime/issues/25555

This creates also an compiler warning `'IOControlCode.KeepAliveValues' is supported on 'windows'`

1nf0rmagician:
[Here](https://github.com/dotnet/runtime/issues/25555#issuecomment-689476572) is a discussion on that topic from the dotnet runtime repo with some hints to solve this issue

Also [here](https://github.com/PHOENIXCONTACT/MORYX-Core/blob/7b2ddd2bf56708527a1db7fcb809ada156eac98a/src/Moryx/Communication/Sockets/TcpTransmission.cs#L85) is a link to the specific line to be adapted.

1nf0rmagician:
And `Console.Beep(int, int)` is only supported on Windows as well with no signs for a linux implementation coming as the discussion [here](https://github.com/dotnet/runtime/issues/24366) indicates

Toxantron:
Closed by #102

PR #102, opened 2021-01-26
Keep alive on UNIX

Toxantron:
Fix exception for cross platform use of binary connection reported in #100.


Toxantron:
[Cross platform support for keep alive](https://github.com/dotnet/corefx/pull/29963) was added to net.core 3.0 in 2018, but not for net standard. I refactored my old solution to support both, which then closes #100 and let's us enable tcp tests on linux from #286.

dbeuchler (approved):

PR #105, opened 2021-03-18
Added workplan changed event to have the possibility to react on changes

slawor:
It is necessary to have an event about a changed workplan to be able to react on this change.

### DIFF
--- a/src/Moryx/Workflows/API/IWorkplanManagement.cs
+++ b/src/Moryx/Workflows/API/IWorkplanManagement.cs
@@ -1,0 +1,17 @@
+using System;
+
+namespace Moryx.Workflows
+{
+    /// <summary>
+    /// Additional facade interface for components that store and provide workplans
+    /// </summary>
+#pragma warning disable 618
+    public interface IWorkplanManagement : IWorkplans
+#pragma warning restore 618
+    {
+        /// <summary>
+        /// A workplan was changed
+        /// </summary>
+        event EventHandler<Workplan> WorkplanChanged;
+    }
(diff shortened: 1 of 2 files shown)


Toxantron:
Since we have to extend the workplan API anyway. Should we include a `SourceId` for workplans. This could be helpful for future changes when workplans are cloned, versioned, ...
@dbeuchler ?

src/Moryx/Workflows/API/IWorkplanManagement.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Moryx.Workflows
+{
+    public interface IWorkplanManagement : IWorkplans
```
Toxantron, line 5: Can you disable the Deprecated warning for this usage?

Toxantron (changes requested):

dbeuchler:
> Since we have to extend the workplan API anyway. Should we include a `SourceId` for workplans. This could be helpful for future changes when workplans are cloned, versioned, ...
> @dbeuchler ?

yes... source id can be helpfull

slawor:
> Since we have to extend the workplan API anyway. Should we include a `SourceId` for workplans. This could be helpful for future changes when workplans are cloned, versioned, ...
> @dbeuchler ?

Can we add it if we need it in the future? I would like to address only one problem

Toxantron (approved):

Toxantron:
Since the requirement has disappeared, could we do this against future without obsolotes?

PR #108, opened 2021-04-28
ReflectionResolver can map strings on properties with primitive datatypes

MarisaGoergen:

### DIFF
--- a/docs/articles/Platform/Bindings.md
+++ b/docs/articles/Platform/Bindings.md
@@ -105,7 +105,7 @@ class Foo
 }
 
 var foo = new Foo();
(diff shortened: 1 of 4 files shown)


Toxantron (approved):

Toxantron:
@dbeuchler do we consider this minor or patch? We did not change APIs, but whether or not we added a feature depends on what people would expect from the API.

dbeuchler:
> @dbeuchler do we consider this minor or patch? We did not change APIs, but whether or not we added a feature depends on what people would expect from the API.

Puh... in general I agree with you but for me a patch is also okay.

Toxantron:
Patch it is! 😄

PR #110, opened 2021-05-05
Kestrel support added

mmarkwort:


mmarkwort:
Unfortunetly we implemented no CamelCasing in the old REST API . So I had to switch of CamelCasing.

We should discuss if we want to use it later. Currently the WebMaintenance expects no CamelCasing.

Toxantron:
As you probably noticed we restructured your proposal a little bit so far. I still want to thank you a lot for your research and contribution, as it has brought a cross-platform efforts forward a lot.

I also understand that your implementation was the least-invasive approach. During internal discussions we scetched a solution for an API that makes it much easier to support WCF and Kestrel with little need for compiler flags. See Moryx.Communication.Endpoints.

We also noticed a little lifecycle problem in your proposal and are currently working to host controller within the modules container instead of bridging containers.

Toxantron:
I eventually needed to revert some changes to the initial version of @mmarkwort since KestrelHostBuilder does not allow more than one host per port. While we considered distributing endpoints across ports I restored the proposed multiplexer structure. The main difference, besides common API with WCF, is the controller scoped resolution of dependencies from the module providing the endpoint.

I'll try this out with the Maintenance and then prepare as a 3.2 release.

Toxantron (approved):

PR #111, opened 2021-05-05
Feature maintenancekestrelport

mmarkwort:


dbeuchler:
In #122 we currently working on a Maintenance hybrid module which hosts their API by asp.net controllers if used by .net 5.0

PR #141, opened 2022-03-23
Fixed overriding endpoints in EndpointCollector

dbeuchler:
Services were overridden by their route because in asp.net it is possible to host several controllers at the same route.

### DIFF
--- a/src/Moryx.Runtime.Kestrel/EndpointCollector.cs
+++ b/src/Moryx.Runtime.Kestrel/EndpointCollector.cs
@@ -9,7 +9,7 @@ namespace Moryx.Runtime.Kestrel
 {
     internal class EndpointCollector
     {
-        private readonly Dictionary<string, Endpoint> _endpoints = new Dictionary<string, Endpoint>();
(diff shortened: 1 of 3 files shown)


src/Moryx.Runtime.Kestrel/EndpointCollector.cs
```diff
@@ -9,32 +9,24 @@ namespace Moryx.Runtime.Kestrel
...
-        public void AddEndpoint(string address, Endpoint endpoint)
+        public void AddEndpoint(Endpoint endpoint)
         {
             lock (_endpoints)
             {
-                _endpoints[address] = endpoint;
-            }
-        }
-
-        public void RemoveEndpoint(string address)
```
Toxantron, line 33: Did we never use the remove method?

Toxantron (commented):
Could you give me a quick feedback on the `Remove` method.

src/Moryx.Runtime.Kestrel/EndpointCollector.cs
dbeuchler: no 0 references in this internal class

Toxantron (approved):

PR #150, opened 2022-07-18
Add Display attributes to default workplan steps

1nf0rmagician:


1nf0rmagician:
Great input, yes that totally makes sense I'll create a follow up for that 👍

PR #153, opened 2022-08-02
Migrate L1 composition to ASP service collection

Toxantron:
We replace the global Castle Container with the service collection and improve MORYX integration in ASP

/cc @dbeuchler Thought this would interest you


dbeuchler:
Sounds good - thanks for info - if you want to have comments, feel free to let me know 👍

Toxantron:
I might, when it comes to migrating/removing the logger management. You have more experience with that.

PR #154, opened 2022-08-04
Feature/create a new project for kernel endpoints

milmilkat:
Added the base project with only the common controller


src/Moryx.Runtime.Endpoints/Common/HostHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    internal static class HostHelper
+    {
+        /// <summary>
+        /// Returns the amount of physical memory installed
+        /// </summary>
+        /// <returns></returns>
+        public static ulong PhysicalMemory()
+        {
+#if HAVE_SYSTEM_MANAGEMENT
```
Toxantron, line 23: This compiler flag does not make sense anymore. It does not exist in the project and the types are not availble in >=net5

Toxantron (approved):

PR #155, opened 2022-08-04
Feature/added logger to the new project

milmilkat:
Added logger but created a separate MR for you to see if it actually makes sense or not

### DIFF
--- a/src/Moryx.Runtime.Endpoints/Logging/Endpoint/LogController.cs
+++ b/src/Moryx.Runtime.Endpoints/Logging/Endpoint/LogController.cs
@@ -1,0 +1,105 @@
+// Copyright (c) 2020, Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System.Linq;
+using Moryx.Logging;
+using Microsoft.AspNetCore.Mvc;
+using Moryx.Runtime.Endpoints.Logging.Endpoint.Models;
+using Moryx.Runtime.Endpoints.Logging.Endpoint.Request;
+using System.Collections.Generic;
+using Moryx.Runtime.Endpoints.Logging.Endpoint.Response;
+using System;
+using Moryx.Runtime.Logging;
+
+namespace Moryx.Runtime.Endpoints.Logging.Endpoint
+{
+    [ApiController]
+    [Route("loggers")]
+    public class LogController : ControllerBase
+    {
+        private readonly IServerLoggerManagement _loggerManagement;
+        private readonly ILoggingAppender _loggingAppender;
+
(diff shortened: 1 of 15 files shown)


src/Moryx.Runtime.Endpoints/Logging/Endpoint/ILogController.cs
```diff
@@ -0,0 +1,29 @@
...
+using Moryx.Runtime.Endpoints.Logging.Endpoint.Models;
+using Moryx.Runtime.Endpoints.Logging.Endpoint.Request;
+using System.Collections.Generic;
+
+namespace Moryx.Runtime.Endpoints.Logging.Endpoint
+{
+    /// <summary>
+    /// Service contract for logging features of the maintenance.
+    /// </summary>
+    internal interface ILogController
```
Toxantron, line 14: That interface is not relevant anymore.

src/Moryx/Logging/ILoggingComponent.cs
```diff
@@ -8,9 +8,5 @@ namespace Moryx.Logging
     /// </summary>
     public interface ILoggingComponent
     {
-        /// <summary>
-        /// Logger of this component
-        /// </summary>
-        IModuleLogger Logger { get; set; }
```
Toxantron: We can't remove a property from a released interface in a minor.

src/Moryx.Runtime.Endpoints/Logging/Endpoint/LogController.cs
```diff
@@ -0,0 +1,106 @@
...
+                AppenderId = _loggingAppender.AddRemoteLogAppender(request.Name, request.MinLevel)
+            };
+        }
+
+        [HttpGet("appender/{appenderId}")]
+        public ActionResult<IEnumerable<LogMessageModel>> GetMessages([FromRoute] string appenderId)
+        {
+            var appender = int.Parse(appenderId);
+            if (!_loggingAppender.ValidAppender(appender))
+                throw new ArgumentException($"Appender with Id {appenderId} could not be found");
```
Toxantron, line 63: For an unknown id I would prefer a NotFound instead of InternalServerError 500

Toxantron (commented):
I know you we had to remove ParallelOperations, but we need some replacement to prevent once created `RemoteAppenders` from flooding the memory if no one flushes the messages or removes the appenders. I'd say every log appender that has not been checked for more than 30s can be considered dead.

milmilkat:
@Toxantron I am not sure if I could understand you correctly, but take a look and let me know if further ammendents are required.

Toxantron (approved):

PR #156, opened 2022-08-08
Feature/add modules to the new project

milmilkat:


src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        [HttpGet("module/{moduleName}/healthstate")]
+        public ActionResult<ServerModuleState> HealthState([FromRoute] string moduleName)
+        {
+            var module = _moduleManager.AllModules.FirstOrDefault(m => m.Name == moduleName);
+            if (module != null)
+                return module.State;
+
+            Response.StatusCode = (int)HttpStatusCode.NotFound;
+            return ServerModuleState.Failure;
```
Toxantron, line 81: We schould return `NotFound()` here instead of the Legacy code.

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        [HttpGet("module/{moduleName}/notifications")]
+        public ActionResult<IEnumerable<NotificationModel>> Notifications([FromRoute] string moduleName)
+        {
+            var module = _moduleManager.AllModules.FirstOrDefault(m => m.Name == moduleName);
+            if (module != null)
+                return Ok(module.Notifications.Select(n => new NotificationModel(n)));
+
+            Response.StatusCode = (int)HttpStatusCode.NotFound;
+            return Ok(Enumerable.Empty<NotificationModel>());
```
Toxantron, line 92: Same here. We can't set 404 and then return Ok=200

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -0,0 +1,313 @@
...
+        public ActionResult Reincarnate([FromRoute] string moduleName)
+        {
+            var module = GetModuleFromManager(moduleName);
+            if (module == null)
+                throw new ArgumentException(nameof(moduleName));
+
+            new Thread(() =>
+            {
+                _moduleManager.ReincarnateModule(module);
+            }).Start();
```
Toxantron, line 127: This needs to be a background thread and must be disposed after completion. Preferably use ThreadPool or TaskFactory that delegates to the ThreadPool.

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -0,0 +1,313 @@
...
+                var config = GetConfig(module, true);
+                EntryConvert.UpdateInstance(config, request.Config.Root, serialization);
+                _configManager.SaveConfiguration(config, request.UpdateMode == ConfigUpdateMode.UpdateLiveAndSave);
+
+                if (request.UpdateMode == ConfigUpdateMode.SaveAndReincarnate)
+                    // This has to be done parallel so we can also reincarnate the Maintenance itself
+                    new Thread(() =>
+                    {
+                        _moduleManager.ReincarnateModule(module);
+                    }).Start();
```
Toxantron, line 205: See above.

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -0,0 +1,313 @@
...
+                var config = GetConfig(module, false);
+                var configModel = new Config
+                {
+                    Root = EntryConvert.EncodeObject(config, serialization)
+                };
+                return configModel;
+            }
+            catch (Exception ex)
+            {
+                throw new ArgumentException(moduleName, nameof(moduleName), ex);
```
Toxantron, line 185: This would also throw an argument Exception if converting the config fails. I think it should return NotFound if the module does not exist and otherwise whatever went wrong.

Toxantron (changes requested):

Toxantron (approved):

PR #159, opened 2022-08-12
Refactor logging to use ILogger from MS extensions

Toxantron:


src/Moryx.Model/Configuration/ModelConfiguratorBase.cs
```diff
@@ -139,14 +139,14 @@ public virtual async Task<DatabaseMigrationSummary> MigrateDatabase(IDatabaseCon
                 await context.Database.MigrateAsync();
                 result.Result = MigrationResult.Migrated;
                 result.ExecutedMigrations = pendingMigrations;
-                Logger.Log(LogLevel.Info, "Database migration for database '{0}' was successful. Executed migrations: {1}",
+                Logger.Log(LogLevel.Information, "Database migration for database '{0}' was successful. Executed migrations: {1}",
```
milmilkat, line 142: Logger has LogInformation, LogWarning, etc.
I am not sure if changing log level will have the same effect

src/Moryx.Model/DbContextManager.cs
```diff
@@ -19,35 +20,12 @@ namespace Moryx.Model
...
-
-        /// <summary>
-        /// Logger root for this component
-        /// </summary>
-        public ILoggerManagement LoggerManagement { get; set; }
-
-        #endregion
-
-        string ILoggingHost.Name => nameof(DbContextManager);
         private ModelWrapper[] _knownModels;
```
milmilkat, line 25: Could be an enumerable for memory purposes, if it doesn't change many things!

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
```diff
@@ -186,23 +187,7 @@ public ActionResult<DatabaseUpdateSummary> MigrateDatabaseModel(string targetMod
                 throw new ArgumentException("Config values are not valid", nameof(configModel));
```
milmilkat, line 187: Should be returning NotFound instead?

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
```diff
@@ -175,8 +176,8 @@ public ActionResult<InvocationResponse> RestoreDatabase(string targetModel, Rest
             return new InvocationResponse();
         }
 
-        [HttpPost("model/{targetModel}/{migrationName}/migrate")]
-        public ActionResult<DatabaseUpdateSummary> MigrateDatabaseModel(string targetModel, string migrationName, DatabaseConfigModel configModel)
+        [HttpPost("model/{targetModel}/migrate")]
+        public async Task<ActionResult<DatabaseMigrationSummary>> MigrateDatabaseModel(string targetModel, DatabaseConfigModel configModel)
```
milmilkat, line 180: Since this endpoint is changed, the corresponding UI should also change, I am not sure if that is considered or no

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
```diff
@@ -156,7 +156,7 @@ public ActionResult ConfirmWarning([FromRoute] string moduleName)
 
             var notifications = module.Notifications.ToArray();
             foreach (var notification in notifications)
-                notification.Confirm();
+                module.AcknowledgeNotification(notification);
```
milmilkat, line 159: Should we remove all the notification in one step, or they should be acknowledged one by one?

src/Moryx.Runtime.Kernel/Modules/Components/ModuleStarter.cs
```diff
@@ -93,7 +93,7 @@ private void ExecuteModuleStart(object moduleObj)
             }
             catch(Exception ex)
             {
-                _logger.LogException(LogLevel.Error, ex, "Failed to start module {0}", module.Name);
+                _logger.LogError(ex, "Failed to start module {0}", module.Name);
```
milmilkat, line 96: Maybe $ string interpolation could be more readable

src/Moryx.Runtime/Modules/ServerModuleBase.cs
```diff
@@ -84,9 +78,10 @@ void IInitializable.Initialize()
         void IServerModuleStateContext.Initialize()
         {
             // Activate logging
-            LoggerManagement.ActivateLogging(this);
-            LoggerManagement.AppendListenerToStream(ProcessLogMessage, LogLevel.Warning, Name);
-            Logger.Log(LogLevel.Info, "{0} is initializing...", Name);
+            var logger = new ModuleLogger(GetType().Namespace, GetType(), LoggerFactory);
+            logger.SetNotificationTarget(Notifications.AddFromLogStream);
+
+            Logger.Log(LogLevel.Information, "{0} is initializing...", Name);
```
milmilkat, line 84: Not sure $ interpolation could be more readable or not

src/Moryx/Logging/IModuleLogger.cs
```diff
@@ -64,13 +65,8 @@ public interface IModuleLogger : IEnumerable<IModuleLogger>, INamedChildContaine
...
-        /// Active logging level
+        /// Define a target for all messages with a level of <see cref="LogLevel.Warning"/> or higher
         /// </summary>
-        LogLevel ActiveLevel { get; }
-
-        /// <summary>
-        /// Clone this instance to be used on a new object
-        /// </summary>
-        IModuleLogger Clone(Type targetType);
+        void SetNotificationTarget(Action<LogLevel, string, Exception> notificationTarget);
```
milmilkat, line 70: This has been added and not implemented, was it necessary? Or it is because of inheritence?

src/Moryx.DependentTestModule/ModuleController/ModuleController.cs
```diff
@@ -23,8 +24,8 @@ public class ModuleController : ServerModuleFacadeControllerBase<ModuleConfig>,
         [RequiredModuleApi(IsStartDependency = true, IsOptional = false)]
         public ITestModule TestModule { get; set; }
 
-        public ModuleController(IModuleContainerFactory containerFactory, IConfigManager configManager, IServerLoggerManagement loggerManagement) 
-            : base(containerFactory, configManager, loggerManagement)
+        public ModuleController(IModuleContainerFactory containerFactory, IConfigManager configManager, ILoggerFactory loggerFactory)
```
1nf0rmagician, line 27: Could we create something like a MigrationGuide document where we gather the necessary changes for the next major?

Toxantron: Good point

src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the Apache License, Version 2.0
 
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Moryx.Configuration;
 using Moryx.Container;
+using Moryx.Logging;
```
1nf0rmagician, line 8: Is this using required?

src/Moryx.TestTools.UnitTest/DummyLogger.cs
```diff
@@ -138,6 +116,11 @@ public IModuleLogger GetChild(string name, Type target)
         {
             return new DummyLogger();
         }
+
+        public void SetNotificationTarget(Action<LogLevel, string, Exception> notificationTarget)
+        {
+            throw new NotImplementedException();
```
1nf0rmagician, line 122: What was your intention here? Do we still need the DummyLogger test class, if yes should there be an implementation here?

src/Moryx.Model/Configuration/ModelConfiguratorBase.cs
Toxantron: `LogInformation` is just an extension that calls `Log(LogLevel.Information` underneath.

The change was from Moryx.LogLevel to MS.LogLevel, which are similar but not identical.

src/Moryx.Model/DbContextManager.cs
Toxantron: How does using an enumerable change memory footprint? Array is AFAIK the most efficient collection type and implementation of IEnumerable.

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
Toxantron: Good point, but I didn't intent do change the endpoints in this MR. I just fixed it by accident trying to get it to compile again.

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
Toxantron: AFAIK changing to async task does not affect REST signature. It just syntactic sugar in C#, but in the end it's still HTTP Request<->Response.

src/Moryx.Runtime.Endpoints/Modules/Endpoint/ModulesController.cs
Toxantron: In that case it could help, but it would require another method in the module API.

src/Moryx.Runtime.Kernel/Modules/Components/ModuleStarter.cs
Toxantron: Using string interpoltion in logging is discouraged for performance and stability. 

It formats the string `before` passing to the logger, that might not even write it depending on current log level. It could also cause exceptions that are otherwise handled by the logger.

src/Moryx/Logging/IModuleLogger.cs
Toxantron: It should be implemented inside the Module

src/Moryx.Model/DbContextManager.cs
milmilkat: IEnumerable is an unallocated datastructure. A good practice says that what comes from db, just pass it forward to the front end and leave it unallocated. So, as long as we don't say .ToArray or .ToList, etc. The IEnumerable is unallocated and saves memory. I am not sure if this theory is applicable in this case or not

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
milmilkat: I am not talking about the async, its the migrationName which is excluded in the new endpoint and the previous UI was making a request with two URL parameters but the new one only needs targetModel. That is for sure a new endpoint

src/Moryx.Model/DbContextManager.cs
Toxantron: `IEnumerable` is just an interface, it depends on the concrete type underneath. Declaring the field as `IEnumerable` and assigning an array has the same result.

What you describe only applies to implementations of IEnumerable that only allocate when the `Enumerator` is created, which is often the cases for database queries. In that case it is recommended to pass IEnumerable to avoid allocation until it is needed. `_knownModels` on the other hand is the result of iterating the DI container and in that case storing a handful of references is more efficient that reiterating the container for every request.

src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs
Toxantron: Good catch, yes that is breaking.

src/Moryx.TestTools.UnitTest/DummyLogger.cs
Toxantron: Since we still offer the `ModuleLogger`, we also need to provide the `DummyLogger` for those cases.

PR #166, opened 2022-08-30
Reduce IModuleLogger interface

Toxantron:


src/Moryx/Logging/ModuleLogger.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0
 
 using Microsoft.Extensions.Logging;
+using Moryx.Container;
```
1nf0rmagician, line 5: Since you have changed this file anyway, would you take care of the 'missing xml comment' warnings here? Especially the constructor with its parameters might benefit from documentation 😅 (and the SetNotificationTarget method which I struggled with already not knowing what it's for)

1nf0rmagician (changes requested):
The main purpose of this PR is to remove the IModuleLogger interface where possible and to simplify IModuleLogger, isn't it? Thereby, the possibility to set a notification target is now only available when using the actual implementation in ModuleLogger, correct?

src/Moryx/Logging/ModuleLogger.cs
Toxantron: How about I move that to the constructor.

The idea is, that only upon creation the delegate can and must be set.

1nf0rmagician: If it must be set I think the constructor would be a gould place yes. A constructor without the delegate parameter could set an empty one? In that case it hould have a hint in the comment that nothing happens in certain situations 🤔

PR #172, opened 2022-09-26
Introduce module category for WebModuleAttribute

Toxantron:


src/Moryx.Asp.Extensions/ModuleMenuCategory.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Moryx.Asp.Extensions
+{
+    public enum ModuleCategory
```
1nf0rmagician, line 3: License notice and comment on the enum itself would be great 😬

1nf0rmagician (dismissed):

PR #178, opened 2022-11-10
PossibleTypes are now serialized in Entry

MathoMathiasCamara:
PossibleTypes are now serialized in Entry


Toxantron:
Sorry, this is not the fix. Possible values is the responsiblity of `ICustomSerialization`. Please check which serialization is used in the module that causes the issues.

PR #179, opened 2022-11-14
Proposed change for the treatment of null values in the ValueProvider

andreniggemann:
### Summary

This pull request tries to adress some pain point I have encounterd while working with the serialization in MORYX related to default values.

Problems:
1. Using nullable value types does only work when specifing a DefaultValue that is not `null`.
    -  throws a NullReferenceException in the FilterProperties Method of the ValueProviderExecutor
    - Because the DefaultValueProvider does not return "Handled" when the default value is set to `null` the property could be passed to another provider although a DefaultValue is explicitly specified
3. Can't have types without a parameterless comstructor, even on fields that schould not be serialized
    -  Leads to a MethodNotFound Exception because Activator.CreateInstance can't find a paramless ctor.

Proposed solution:
1. ValueProviderExecutor 
    -  Stopping the iteration on a null value
5. DefaultValueProvider
    - Checking the DefaultValueAttribute before trying to create a new instance of the properties type
    - Making the ValueProviderResult indepent of the returned default value beeing null or not

Effect: Nullable properties can be marked with DefaultValue(null) to explicity say that they should not be initialized with any value.

### Review

**Typical tasks**

- [ ] Merge-Request is well described

**Clean code**

### DIFF
--- a/src/Moryx/Configuration/ValueProvider/IValueProviderValidator.cs
+++ b/src/Moryx/Configuration/ValueProvider/IValueProviderValidator.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) 2020, Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System.Reflection;
+
+namespace Moryx.Configuration
+{
+    /// <summary>
+    /// Interface for validators that ensure properties have been filled as expected
+    /// </summary>
+    public interface IValueProviderValidator
+    {
+        /// <summary>
+        /// Checks the property if it was correctly set by a value provider
+        /// </summary>
+        /// <param name="propertyInfo">Property to check</param>
+        /// <param name="target">Object the property is on. Can be used to access the value for instance</param>
+        /// <returns>True if the property was correctly set</returns>
+        bool CheckProperty(PropertyInfo propertyInfo, object target);
+    }
+}
--- a/src/Moryx/Configuration/ValueProvider/PropertyValidationException.cs
+++ b/src/Moryx/Configuration/ValueProvider/PropertyValidationException.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Reflection;
+using System.Runtime.Serialization;
+
+namespace Moryx.Configuration
+{
+    [Serializable]
+    internal class PropertyValidationException : Exception
+    {
+        private PropertyInfo property;
+        private ValueProviderExecutorSettings settings;
+
+        public PropertyValidationException()
+        {
+        }
+
+        public PropertyValidationException(string message) : base(message)
+        {
+        }
+
+        public PropertyValidationException(PropertyInfo property, ValueProviderExecutorSettings settings) 
+            : this($"Failed to provide acceptable value for property {property.Name} on type {property.DeclaringType}")
+        {
+            this.property = property;
+            this.settings = settings;
+        }
+
+    }
+}
--- a/src/Moryx/Configuration/ValueProvider/Providers/ActivatorValueProvider.cs
+++ b/src/Moryx/Configuration/ValueProvider/Providers/ActivatorValueProvider.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) 2020, Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System;
+using System.Reflection;
+
+namespace Moryx.Configuration
+{
+    /// <summary>
+    /// Uses the activator on properties of classes with a default constructor
+    /// </summary>
+    public sealed class ActivatorValueProvider : IValueProvider
+    {
+        /// <inheritdoc/>
+        public ValueProviderResult Handle(object parent, PropertyInfo property)
+        {
+
+            var propertyType = property.PropertyType;
+            var value = property.GetValue(parent);
+
+            if (propertyType.IsClass && value == null && !propertyType.IsAbstract && propertyType != typeof(string))
+            {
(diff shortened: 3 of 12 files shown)


Toxantron:
Thank you very much for reporting your issues as well as proposing a fix, that contains some good ideas, like the return value and out parameter. I also agree we should skip values that were empty and could not be provided as this has evolved from just creating complex config trees to more mixed use cases like domain objects.

However also some remarks:
- It's not the responsibility of the value provider to determine which fields to provide and which not, we defined the filters for that like the `EntrySerializeAttribute`
- Declaring a default Value of `null` is redundant with the type default, it's like declaring a default for int of 0
- Always returning handled breaks the chain as it prevents any following provider from being executed, like the `SharedConfigProvider`

Ideas:
- I am not really a fan, but we could define a NullDefaultFilter to create the behavior of skipping null properties if their default can remain at null, if that helps you. It's isolated, optional and does not break existing code
- If this is related to configs, we shoudl reexamine values that are empty and should not be provided, because what would those config values represent and who sets them.

andreniggemann:
Thanks for the feedback. 

1. There are usecases where we can not add filters, because we don't control the code calling the ValueProviderExecutor. That´s why I wanted to add a solution to the DefaultValueProvider, because it is nearly universally used. One such use case relates to Parameters and the Workflow Editor. A colleague added a Predicate to a parameter class and now the workflow can not be displayed, because Func has no parameterless constructor. Adding a Filter would rely on the library author to add the filter in their code.
2. I would argue that DefaultValue(null) expresses the intent of the programmer, that this property should be intialized to `null`. Therefore it should be handled like any other value one could specify via the DefaultValueAttribute. This would include the DefaultValueProvider returning handled.
3. In my proposed solution Handled is not always returned. It is returned iff the DefaultValueAttribute is specified or a default constructor is present.  This is checked explicitly by one of the test cases.

One other solution I could see, is adding a check for the presence of a default constructor before calling Activator.CreateInstance in the DefaultValueProvider. I had implemented this initially. But did not really like the required reflection based code and thought it might be useful to be more explicit about not initialized values. I can share that code if you want.

One other usecase I had was allowing nullable valuetypes (int?, bool?, TEnum?) in Resources, representing the setup state that the PLC was configured for. Null appeared to be a sensible option on startup before the PLC was send any configuration messages. This use case does not rely on the DefaultValue(null) however, only on skiping empty objects.

On a related note: The DefaultValueProvider appears to have two related but seperate functions. 
1. Reading the DefaultValueAttribute and applying it's value to the property
2. Calling the default constructor on non abstract classes.

Maybe it makes sense to split them into two seperate providers.

Toxantron:
> There are usecases where we can not add filters, because we don't control the code calling the ValueProviderExecutor. That´s why I wanted to add a solution to the DefaultValueProvider, because it is nearly universally used. One such use case relates to Parameters and the Workflow Editor. A colleague added a Predicate to a parameter class and now the workflow can not be displayed, because Func has no parameterless constructor. Adding a Filter would rely on the library author to add the filter in their code.

If a module does not support it, the modules maintainer should take care to provide a filter option. Products, Resources, Recipes and Parameters all support EntrySerialize. So for the Func property `EntrySerialize(Never)` should have resolved the issue.

> I would argue that DefaultValue(null) expresses the intent of the programmer, that this property should be intialized to null. Therefore it should be handled like any other value one could specify via the DefaultValueAttribute. This would include the DefaultValueProvider returning handled.

It is automatically initialized to null by .NET. Again I would argue that requires a filter to skip non-providable properties and especially to make sure it does not stop/crash on empty properties.

> In my proposed solution Handled is not always returned. It is returned iff the DefaultValueAttribute is specified or a default constructor is present. This is checked explicitly by one of the test cases.

My mistake, you are right.

> One other usecase I had was allowing nullable valuetypes (int?, bool?, TEnum?) in Resources, representing the setup state that the PLC was configured for. Null appeared to be a sensible option on startup before the PLC was send any configuration messages. This use case does not rely on the DefaultValue(null) however, only on skiping empty objects.

Yeah, I agree. I'd prefer it not being in the provider but rather a filter or in the executor.

> On a related note: The DefaultValueProvider appears to have two related but seperate functions.

I agree.

andreniggemann:
I'm sorry for the late response but I had little time last week.

> If a module does not support it, the modules maintainer should take care to provide a filter option. Products, Resources, Recipes and Parameters all support EntrySerialize. So for the Func property EntrySerialize(Never) should have resolved the issue.

image
For what it's worth, the concrete issue ssems to be that the workflow tries to serialize the task using TaskStepSerialization.
For that it tries to create Prototypes and falls back on to the default serialization which uses the DefaultValueProvider with the DefaultSettings. Therefore checks that might be in the library get skipped.

> It is automatically initialized to null by .NET. Again I would argue that requires a filter to skip non-providable properties and especially to make sure it does not stop/crash on empty properties.

Are you sure relying on filters to make sure the providers do not crash is the best approach? Providers are executed sequentially, so it could always be the case that a subsequent provider can handle a property. For one provider to fail the entire series, nevermind the whole program seems ill adviced. Therefore 
a) each filter should try to detect properties it can not handle and skip them.
b) the executor should catch provider errors and mark them as skipped.

If failing when some properties can't be supplied is a desired feature, for example to ensure null safety later in the program there might be the need for a third concept, next to filters and providers. The validator.
Validators could be employed the same way filters are, requiering unanimous consent that a property was filled correctly.

Toxantron:
Sorry, we will take care of this in the next major. I rebased you changes on future, will do some modifications and would like some feedback afterwards in #206

PR #185, opened 2022-11-18
Clean up for major release

slawor:
- Removed obsoletes
- Removed netstadard target
- Removed not supported features
- Removed unnecessary ClassDisplayAttribute
- Removed some todos


dbeuchler:
If you remove the .net standard target it is completlely not possible anymore to use the moryx libraries in .NET Framework or .net core <6.0 projects. Why is it removed? Also for reusing the libraries without the runtime context, the users are forced to use .net6. As I understand the libraries are currently not using any new api which was introduced in .net6, therefore it cost no sugar.

Toxantron:
@dbeuchler good point for reusable libraries to keep downwards compatible

Toxantron (approved):

PR #191, opened 2022-11-25
Fix packaging of static web assets for net6 projects

1nf0rmagician:
Revert previous attempt to fix the package generation. And replace target with "before" trigger with a simple conditional target. This way, the wwwroot directory is again included as staticwebassets within the identically named folder within the nuget package. 

However, the microsoft typescript team mentions that it would be intentional to include the wwwroot directory in the content and contentfiles directory in the package. This is known to cause problems with Blazor (and obviously also RCLs). Here the discussion for future reference, maybe the way we pack the static web assets might change based on the outcome

- https://github.com/microsoft/TypeScript/issues/42907

The current behaviour causes issues as soon as you try to start a referencing start project, e.g. MORYX-Abstraction layer as of https://github.com/PHOENIXCONTACT/MORYX-AbstractionLayer/commit/4f35e1edf412cb48831a95869c8966a1404c61b5


.build/BuildToolkit.ps1
```diff
@@ -74,7 +74,7 @@ function Invoke-Initialize([string]$Version = "1.0.0", [bool]$Cleanup = $False)
     }
 
     if (-not $env:MORYX_BUILD_CONFIG) {
-        $env:MORYX_BUILD_CONFIG = "Debug";
+        $env:MORYX_BUILD_CONFIG = "Release";
```
seveneleven, line 77: Is this supposed to be changed?

.github/workflows/build-and-test.yml
```diff
@@ -17,7 +17,7 @@ env:
   MORYX_OPTIMIZE_CODE: "false"
   MORYX_BUILD_CONFIG: "Release"
   MORYX_BUILDNUMBER: ${{github.run_number}}
-  dotnet_sdk_version: '6.x'
+  dotnet_sdk_version: '7.x'
```
seveneleven, line 20: 6 should be fine

src/Moryx.Maintenance.Web/Moryx.Maintenance.Web.csproj
```diff
@@ -1,5 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
+﻿<Project Sdk="Microsoft.NET.Sdk.Razor" InitialTargets="BuildUI">
```
seveneleven, line 1: Is this still needed?

.build/BuildToolkit.ps1
1nf0rmagician: Commited by mistake 👍

.github/workflows/build-and-test.yml
1nf0rmagician: Is actually necessary, otherwise the wwwroot directory is copied in the staticwebassets as well as content and contentfiles directories within the package 🤔

src/Moryx.Maintenance.Web/Moryx.Maintenance.Web.csproj
1nf0rmagician: And this is required as well, has propably something todo with the point in time when msbuild checks if the directory exists, but never confirmed that 🤔

PR #192, opened 2022-11-25
`wwwroot` will not be included in `Maintenance.Web` package

seveneleven:


Toxantron:
Don't we need the wwwroot folder as it contains the compiled JS or am I missing something?

1nf0rmagician:
Yeah, I'm trying to solve the same problem with #191 

This solves the issue that the content of wwwroot is in the contentfiles of the package and thus pops up in the wwwroot directory of the referencing application. However, the content of wwwroot should be copied to the staticwebassets folder within the package, as it is done on the dev branch 🤔
So unfortunately this is not the solution to our problem 😐

1nf0rmagician:
Solved with #191

PR #197, opened 2022-12-15
Maintenance UI produces several errors on Startup

MathoMathiasCamara:
removed onUpdateApplicationLoad, onUpdateSystemLoad, onUpdateApplicationInfo, onUpdateHostInfo function call in componentDidMount()


Toxantron:
They were used for the landing page that displayed host information. Most of those functions are windows only/not supported on >net core and we also removed the entire page showing them.

1nf0rmagician:
Great thank you for the clarification 👍😊 I'll merge it then

PR #198, opened 2022-12-19
Rename Workflow to WorkplanInstance

1nf0rmagician:
In order to clarify the meaning of the current `Workflow ` class within the context of `Workplans` we rename the following classes
- `IWorkflow` -> `IWorkplanInstance`
- `Workflow` -> `WorkplanInstance`
-  `SimpleWorkflow` -> `SimpleWorkplanInstance`
- `IWorkflowEngine` -> `IWorkplanEngine`
- `WorkflowEngine` -> `WorkplanEngine`
- `WorkflowFactory` -> `WorkplanInstanceFactory`
- `SubworkflowTransition` -> `SubworkplanTransition`
- `WorkflowValidation` -> `WorkplanValidation`
- `SubworkplanStep` -> `SubworkplanStepBase`
- `SubworkflowStep` -> `SubworkplanStep`


src/Moryx/Workflows/API/Instance/ITokenHolder.cs
```diff
@@ -12,7 +12,7 @@ namespace Moryx.Workplans
     public interface ITokenHolder
     {
         /// <summary>
-        /// Workflow unique id of this holder. Same id can be present in another workflow
+        /// Workplan-unique id of this holder. Same id can be present in another workplan
```
MarisaGoergen, line 15: this is probably for a workplan instance

Toxantron (approved):

src/Moryx/Workflows/API/Instance/ITokenHolder.cs
Toxantron: Actually worlplan ist fine here, as all instances of the same workplan have the same Ids for the steps, it is unique within the plan, not within all plans.

PR #205, opened 2023-01-09
Update/nunit console runner version

seveneleven:


Toxantron:
@seveneleven this is no longer relevant, right?

PR #206, opened 2023-01-10
Fix multiple issues with the value provider

Toxantron:
Rebased version of #179


src/Moryx/Configuration/ValueProvider/Providers/DefaultValueProvider.cs
```diff
@@ -7,72 +7,25 @@
...
-                value = ProvideDefaultValue(property);
-                if (value == null)
-                    return ValueProviderResult.Skipped;
-
-                property.SetValue(parent, value);
-
-                return ValueProviderResult.Handled;
-            }
-
+            if (activatorValueProvider.Handle(parent, property) == ValueProviderResult.Skipped)
```
Toxantron, line 25: This returns `Skipped` if the first call returns `Handled`

PR #208, opened 2023-01-12
Added confirmation dialog to database delete action

MathoMathiasCamara:
Refers to merge request : https://github.com/PHOENIXCONTACT/MORYX-MaintenanceWeb/pull/40


1nf0rmagician:
@VladPxC you might want to have a look here. Starting with our .net5 support wemoved the maintenance to be part of MORYX-Core. Your change makes a lot of sense to me so we would apply it here as well 👍

1nf0rmagician (approved):

PR #216, opened 2023-01-26
Clean up package references and update copyright notice

1nf0rmagician:
First commit adds missing package references to the build.targets file and consolidates versions.

Second commit only updates year of the copyright notice


Directory.Build.targets
```diff
@@ -12,28 +12,33 @@
   <ItemGroup>
     <!--3rd party dependencies-->
     <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.11.0" />
```
MarisaGoergen, line 14: `ctrl + k + d` to make the formatting pretty?

Directory.Build.targets
```diff
@@ -12,28 +12,33 @@
...
+	<PackageReference Update="Moq" Version="4.16.1" />
+	<PackageReference Update="NUnit" Version="3.13.2" />
+	<PackageReference Update="NUnit3TestAdapter" Version="4.1.0" />
 
     <PackageReference Update="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Update="CommandLineParser" Version="2.8.0" />
-
-    <PackageReference Update="Microsoft.EntityFrameworkCore" Version="6.0.0" />
+	  
+	<FrameworkReference Update="Microsoft.AspNetCore.App" Version="6.0.0"/>
```
MarisaGoergen, line 22: does it make sense to have one varible for the net versions in order to know what belongs together?

Directory.Build.targets
1nf0rmagician: It si correctly formated on my machine 😲

src/Moryx.Resources.Management/Resources/ResourceProxyBuilder.cs
```diff
@@ -48,11 +48,8 @@ public void PrepareBuilder()
                 return;
 
             // Check if we already created the assembly and the module builder
-#if HAVE_APPDOMAIN_DYNAMICASSEMBLY
```
MarisaGoergen, line 51: why don't we need that anymore

src/StartProject.Asp/StartProject.Asp.csproj
```diff
@@ -6,7 +6,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-      <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="5.0.13" />
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="6.0.0" />
```
MarisaGoergen, line 9: should we add this also to the standard targets file, since it also depends on the net version?

1nf0rmagician: Yes we could do this, howevr its only purpose is to simplify development with razor pages, there is no actual reason to have it in a real application. I wanted to remove it first but decided otherwise for some reason 🤷‍♂️

src/Moryx.Resources.Management/Resources/ResourceProxyBuilder.cs
1nf0rmagician: The line below that does exactly this but for >.net5. It was forgotten when the net framework target was removed 🙃

Directory.Build.targets
1nf0rmagician: yeah sure 😅

Directory.Build.targets
seveneleven: Tabs vs spaces

1nf0rmagician: Mhh, I thought we had a file somewhere that alignes these settings for all of us 🤔

seveneleven: But this either is not part of the repo or GitHub isn't aware. The root cause here is, that some lines are indented by spaces, some by tabs and some by a mix of both.

1nf0rmagician: unified that ✅

PR #217, opened 2023-01-30
Add Platform intialization to service collection extensions

1nf0rmagician:
With this PR we remove the old RuntimePlatform class and the `PlatformType` enum. As MORYX is now integrated in asp.net core with both UI and server modules hosted within the same runtime, there is no need to host and differentiate multiple MORYX runtimes. The changes also include
- Making the Platform class non-abstract
- Moving the static initialization from RuntimePlatform to Platform
- Adding the Initilization of the class to the `AddMoryxKernel` service collection extension method
- Adding tests for the service collection extensions
- Uncommenting tests for the `ModuleManager`
- Updating the migration guide
-


src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs
```diff
@@ -40,7 +46,7 @@ public static void AddMoryxKernel(this IServiceCollection serviceCollection)
         public static void AddMoryxModules(this IServiceCollection serviceCollection)
         {           
             // Find all module types in the app domain
-            var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
```
Toxantron, line 43: Are you sure about this? This casued issues in the past.

1nf0rmagician: `assembly.GetExportedTypes()` throws a System.NotSupportedException if called on a dynamic assembly. Without it will throw the exception, if that is desired as a safeguard/notification for developers then we need to discuss it. Otherwise it will just prevent the exception to be thrown (which was required for me to run the test)

Toxantron: I am sorry, I misread the change. My mind switched old and new, so I thought you removed the change. Still slightly confused why there are already dynamic assemblies on system boot, but I'll accept it.

Toxantron (approved):

PR #219, opened 2023-02-02
Add method to UOW in order to link entities to business objects

MarisaGoergen:
Add method to UOW in order to link entities to business objects. When uow.SaveChanges() is called, in all linkes business objects the id will be updated


src/Moryx.Model/Repositories/UnitOfWork.cs
```diff
@@ -52,6 +54,11 @@ private IRepository GetRepository(Type api)
             return instance;
         }
 
+        public void LinkEntityToBusinessObject (IPersistentObject businessObject, IEntity entity)
```
1nf0rmagician, line 57: Please provide a comment for public members 🙈

src/Moryx.Products.Management/Implementation/RecipeStorage.cs
```diff
@@ -198,7 +198,7 @@ public static WorkplanEntity ToWorkplanEntity(IUnitOfWork uow, Workplan workplan
                 var reference = referenceRepo.Create((int)WorkplanReferenceType.NewVersion);
                 reference.Source = workplanEntity;
                 reference.Target = workplanEntity = workplanRepo.Create(workplan.Name, workplanEntity.Version + 1, (int)workplan.State);
-                EntityIdListener.Listen(workplanEntity, workplan);
+
```
1nf0rmagician, line 201: Don't we need to link them here now? (Also for all other changes)

MarisaGoergen: Since the EntityIdListener didn't work and the programm itself works, it should be fine

MarisaGoergen: Since the EntityIdListener didn't work and the programm itself works, it should be fine

PR #221, opened 2023-02-03
Fix entity id link in products and resources modules

1nf0rmagician:
Products and Resources modules were not using the newly integrated option to link `IPersistetObject` object to corresponding entities when they were created. The custom way of the modules caused issues, thereby the new way is used now in both modules.


src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -300,8 +300,8 @@ public void Save(Resource resource)
                 var newResources = new HashSet<Resource>();
 
                 var saveResult = ResourceEntityAccessor.SaveToEntity(uow, resource);
-                var entity = saveResult.Item1;
-                if (saveResult.Item2)
+                var entity = saveResult;
+                if (entity.Id == 0)
```
Toxantron, line 304: Careful, this does not work with InMemory

PR #223, opened 2023-02-10
SqLite Databases can not be erased from the maintenance

MathoMathiasCamara:
EfCore locks the file once the connection is etablish, therefore throwing an exception if you try to delete the file. refer to this [git issue](https://github.com/dotnet/efcore/issues/26580) for more info.


src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
```diff
@@ -67,8 +68,10 @@ public override DbContextOptions BuildDbContextOptions(IDatabaseConfig config)
         }
 
         private static string BuildConnectionString(IDatabaseConfig config)
-        {
-                return config.ConnectionSettings.ConnectionString;
+        {  
+           return  config.ConnectionSettings.ConnectionString.Contains("Pooling") ?
```
seveneleven, line 72: I'd prefer to use the connection string builder to set this up, instead of concatenating strings.

1nf0rmagician:
Wouldn't it be better for us to keep the pooling active (to improve connection speed) and instead call clearpool() before we try to delete the database? As there is only one function that could trigger a database deletion from the UI this should be possible right? 🤔

src/Moryx.Model.Sqlite/SqliteDatabaseConfig.cs
```diff
@@ -47,7 +47,7 @@ public override string Database
         }
 
         /// <inheritdoc />
-        [DataMember, Required, DefaultValue("Data Source=.\\db\\<DatabaseName>.db")]
+        [DataMember, Required, DefaultValue("Data Source=.\\db\\<DatabaseName>.db;Pooling=false")]
```
seveneleven: Instead of making this option public/a general setting, you could try to use it only for deletion

src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
```diff
@@ -68,7 +69,9 @@ public override DbContextOptions BuildDbContextOptions(IDatabaseConfig config)
 
         private static string BuildConnectionString(IDatabaseConfig config)
         {
-                return config.ConnectionSettings.ConnectionString;
+            var connectionStringBuilder = new SqliteConnectionStringBuilder(config.ConnectionSettings.ConnectionString);
+            connectionStringBuilder.Pooling = false;
```
seveneleven, line 73: As mentioned above, why not doing this only for the case of deleting a database?

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+
+        [Test]
+        public async Task SqliteDeleteDatabaseShouldWork()
+        {
+            string databaseName = "TestDeleteProductDatabase";
+            string datasource;
+            SqliteDatabaseConfig dbConfig;
+            SqliteModelConfigurator configurator;
+            InitializeSqliteTools(databaseName, out datasource, out dbConfig, out configurator);
```
seveneleven, line 65: why aren't these lines part of the `[SetUp]`?

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Moryx.Model.Tests
+{
+    [TestFixture]
+    public  class SqliteTests
+    {
+        private DbContextManager _dbContextManager;
+        private DatabaseConfigUpdateService _databaseConfigUpdateService;
```
seveneleven, line 22: These are not used below

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            var result = await configurator.TestConnection(dbConfig);
+            Assert.AreEqual(result, TestConnectionResult.ConnectionOkDbDoesNotExist);
+
+            bool isCreated = await configurator.CreateDatabase(dbConfig);
+
+            Assert.IsTrue(isCreated);
+            Assert.IsTrue(File.Exists(datasource));
+
+            //remove the database
+            await configurator.DeleteDatabase(dbConfig);
```
seveneleven, line 54: I consider cleaning things up part of a `[TearDown]`

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        public async Task SqliteCreateDatabaseShouldWork()
+        {
+            string databaseName = "TestCreateProductDatabase";
+            string datasource;
+            SqliteDatabaseConfig dbConfig;
+            SqliteModelConfigurator configurator;
+            InitializeSqliteTools(databaseName, out datasource, out dbConfig, out configurator);
+
+            var result = await configurator.TestConnection(dbConfig);
+            Assert.AreEqual(result, TestConnectionResult.ConnectionOkDbDoesNotExist);
```
seveneleven, line 46: Here, expected and actual values are switched. This could lead to confusing outputs. Same probably goes for following calls of `AreEqual()`

src/Tests/Moryx.Model.Tests/SqliteTests.cs
seveneleven: Also, 
* you might want to make sure, the database did not exist before executing this test. 
* you could use a different database filename to make sure, all tests are 'isolated'

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        private static ConfigManager CreateConfigManager()
+        {
+            var configManager = new ConfigManager
+            {
+                ConfigDirectory = ""
+            };
+            return configManager;
+        }
+
+        private static void InitializeSqliteTools(string databaseName, out string datasource, out SqliteDatabaseConfig dbConfig, out SqliteModelConfigurator configurator)
```
seveneleven: * I'd remove the `out` parameters 
* and make the `configurator` a class member, since it seems to be the 'SUT'

src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
1nf0rmagician: I agree, is there a reason I can't see? 🙈

src/Tests/Moryx.Model.Tests/SqliteTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        private SqliteDatabaseConfig dbConfig;
+        private SqliteModelConfigurator configurator;
+        private string datasource;
+
+        [SetUp]
+        public void Setup()
+        {
+            string databaseName = "TestDatabase";
+            datasource = $".\\db\\{databaseName}.db";
+            string connectionString = $@"Data Source={datasource};Pooling=false";
```
1nf0rmagician, line 30: Adapt the connection string here, if you change the general pooling setting to be disable on database deletion only

src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
MathoMathiasCamara: we are using the File.Remove() to delete the database. we are not using a query like we do in Postgres to delete the database therefore we don't have access to SqlConnection. refer to [sqlite delete drop database ](https://www.tutlane.com/tutorial/sqlite/sqlite-delete-drop-database#:~:text=Like%20most%20relational%20database%20systems,and%20it%20is%20case%2Dsensitive.)

seveneleven: We might rethink/rework how to create/delete databases, or which conditions should apply to consider a database being created/deleted.

src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
```diff
@@ -38,6 +38,11 @@ protected override DbCommand CreateCommand(string cmdText, DbConnection connecti
         /// <inheritdoc />
         public override Task DeleteDatabase(IDatabaseConfig config)
         {
+            using (var sqliteConnection = (SqliteConnection)CreateConnection(config))
```
1nf0rmagician, line 41: You created the sqliteconnection here but never used it. ClearAllPools is a static method you don't need a surrounding using statemant for 🙃

src/StartProject.Asp/Config/Moryx.Products.Model.ProductsContext.DbConfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "ConfigState": "Generated",
-  "Host": "localhost",
-  "Port": 5432,
-  "Database": "Framework6_ProductsContext",
-  "Username": "postgres",
-  "Password": "postgres"
+  "ConnectionSettings": {
```
1nf0rmagician, line 3: I like that we habe one db config for sqlite and one for postgres by default. However, please rename the database to Framework6_ProductsContext. 
And is there a reason you deleted all the other configs that were checked in?

MathoMathiasCamara: i changed the configs a lot , so i don't even remember what was there initially.

src/Moryx.Model.Sqlite/SqliteModelConfigurator.cs
```diff
@@ -72,7 +75,8 @@ public override DbContextOptions BuildDbContextOptions(IDatabaseConfig config)
 
         private static string BuildConnectionString(IDatabaseConfig config)
         {
-                return config.ConnectionSettings.ConnectionString;
+            var connectionStringBuilder = new SqliteConnectionStringBuilder(config.ConnectionSettings.ConnectionString);
+            return connectionStringBuilder.ToString();
```
seveneleven, line 79: I'd assume, this should be `connectionStringBuilder.ConnectionString`. But fine, if this works.

MathoMathiasCamara: i will revert the chages here because i am not using it for "Pooling=False" anymore.

seveneleven (approved):

1nf0rmagician (approved):

PR #231, opened 2023-03-07
ControlCabinetType is not creatable using the DemoImporter

MathoMathiasCamara:
New ProductType that has List<ProductPartLink> can't be created using the DefaultImporter in the Demo.


MarisaGoergen (approved):

Toxantron:
In the past we required ProductType definitions to declare all lists. Because this will only prevent the exception in one place and it could still occur in any other place. If we want to fix it, the storage must populate the collection instead of skipping on null.

PR #250, opened 2023-04-11
Get all resources when parameter of GetDetailsBatch is empty

MarisaGoergen:
Get all resources when parameter of GetDetailsBatch is empty


1nf0rmagician (dismissed):

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
```diff
@@ -59,7 +59,7 @@ public ActionResult<ResourceModel[]> GetDetailsBatch([FromQuery] long[] ids)
         {
             var converter = new ResourceToModelConverter(_resourceTypeTree, _serialization);
 
-            if (ids is null)
+            if (ids is null || ids.ToList().Count == 0)
```
Toxantron, line 62: This should not be a to list, but compare length please.

1nf0rmagician (approved):

PR #251, opened 2023-04-11
Stop exposing `internal` methods on serialization

seveneleven:
Invoking `internal` methods from the UI led to exceptions. So, they won't be provided to the API anymore.

### DIFF
--- a/src/Moryx/Serialization/EntrySerializeSerialization.cs
+++ b/src/Moryx/Serialization/EntrySerializeSerialization.cs
@@ -52,7 +52,7 @@ namespace Moryx.Serialization
         /// <inheritdoc />
         public override IEnumerable<MethodInfo> GetMethods(Type sourceType)
         {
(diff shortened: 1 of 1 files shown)


1nf0rmagician:
I'm not sure if this is the right solutions as the behaviour hasn't changed lately and never caused issues. But if it is the right solution @Toxantron should we also fix this in the last major?

1nf0rmagician (approved):

1nf0rmagician:
Discussion with Christian and Marisa lead to the conclusion: We fix the bug here to not reproduce filters that we would need for other serializations as well. (Assuming this bug was circumvented in all prior major versions by just using public methods we do not backport the change)

Toxantron:
Explizitly removing it here instead of fixing the invocation seems wrong. I am not sure if it is used somewhere, but exposing internal means that developers can have "Debug" functions with internal objects. If the method has to be public, so does every of its arguments.

Toxantron:
@seveneleven this can be closed right?

seveneleven:
Has been solved with https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/254

PR #253, opened 2023-05-05
Catch exception on missing license

MathoMathiasCamara:
Problem:

- When starting the application, the `endpoint.Mapcontrollers()` call raises a TargetInvocationException when there is a module without a license:

`System.Reflection.TargetInvocationException: 'Exception has been thrown by the target of an invocation.'
WupiException: Der CmContainer-Eintrag wurde nicht gefunden, Fehler 200.`
This should be catched or at least give a hint to the module in question.

Solution:
- Added a subscription to the AppDomain unhandled Exception event to collect all the license missing exception before the web host is built.
in case there is an exception we build and execute the license exception-page host instead of the default host configured by the user.
- Added a new Extension method 'StartAllMoryxModules' for the IServiceProvider.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -11,5 +11,6 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+	<add key="3rdParty" value="https://packages-ctvc.europe.phoenixcontact.com/nuget/3rdParty/" />
   </packageSources>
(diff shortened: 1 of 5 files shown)


Toxantron:
I agree we need to catch missing license exceptions and display them properly. But the Framework is Open Source and will remain independent of our commercial code and specifically Wibu.

PR #282, opened 2023-05-31
Make SQLite Test OS independent

1nf0rmagician:
The path of the test database used windows seperators, this is replaced now by `Path.Combine`


1nf0rmagician:
Already resolved with #286

PR #283, opened 2023-06-01
resource reference md file added

MathoMathiasCamara:


docs/articles/Core/ResourceReference.md
```diff
@@ -0,0 +1,25 @@
+# ResourceReference
+
+`ResourceReference` is an attribute that links to resources together.
```
MarisaGoergen, line 3: maybe `several` resources, since it can also be used for a list (see channels in IMessageDriver)

MarisaGoergen (changes requested):

MarisaGoergen (approved):

PR #305, opened 2023-07-20
Remove dependency on Castle Windsor

Toxantron:
We have noticed from [Windsor Milestone 6](https://github.com/castleproject/Windsor/issues/621) that the project is in need of a maintainer and preventing us in the future with NET8.

One proposal is to remove Windsor and rebuild to SC only.

Alternatively we could join Windsor and support the project.


Toxantron:
With the release of [Castle.Windsor v6.0.0](https://github.com/castleproject/Windsor/releases/tag/v6.0.0) this is no longer necessary.

PR #307, opened 2023-07-21
Update Castle.Windsor and clean-up container structure

Toxantron:
With the [latest release](https://github.com/castleproject/Windsor/releases/tag/v6.0.0) of Castle.Windsor we no longer need to replace anything and just update and clean-up for MORYX 8.


Toxantron:
Summary:
- I removed the inheritance tree caused by the global/local split
- I removed the registrator and everything associated and put it directly in the container wrapper
- All overloads of registration in container and registrator were replaced by extensions
- LoadComponents and installers were replaced by extensions
- All `Resolve` overloads were replaced by extensions
- Dependency overrides in the registration were replaced with sub resolves in the `MoryxFacility`

This means that our wrapper of Castle.Windsor has gotten much smaller **and** all MORYX specific code and behavior is now located in a MoryxFacility as intended by Castle architecture.

1nf0rmagician:
Since there was a change in the Resources module I guess there might be some changes in this MR that should be reflected in the Migration guide for MORYX 8. Can you confirm that @Toxantron and if yes add the respective parts? 😬

PR #308, opened 2023-07-26
Remove AddDefaultValueSerializer when loading resources from the DB

MarisaGoergen:
The DefaultValueAttributeProvider.Handle function replaces system default values with the costum default values set using the DefaultValueAttribute. The consequence was that setting the system default value wasn't possible anymore, since it gets overwritten when loading the resource  from the database. NewtonsoftJson is able to set the costum default value directly when populating an object since Version 12.0.1.

### DIFF
--- a/src/Moryx.Resources.Management/Resources/ResourceEntityAccessor.cs
+++ b/src/Moryx.Resources.Management/Resources/ResourceEntityAccessor.cs
@@ -76,11 +76,6 @@ namespace Moryx.Resources.Management
             if (ExtensionData != null)
                 JsonConvert.PopulateObject(ExtensionData, Instance, JsonSettings.Minimal);
 
-            // Read everything else from defaults
-            ValueProviderExecutor.Execute(Instance, new ValueProviderExecutorSettings()
-                .AddFilter(new DataMemberAttributeValueProviderFilter(false))
-                .AddDefaultValueProvider());
-
             return Instance;
         }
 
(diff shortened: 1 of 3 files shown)


MarisaGoergen:
Is the `DefaultValueAttributeProvider` still neccessary? @Toxantron

Toxantron:
> Is the `DefaultValueAttributeProvider` still neccessary? @Toxantron

Yes, we still need it (and hopefully use it) to populate config prototypes when sending them to the UI.

src/Moryx.Resources.Management/Resources/ResourceEntityAccessor.cs
```diff
@@ -78,8 +78,7 @@ public Resource Instantiate(IResourceTypeController typeController, IResourceGra
 
             // Read everything else from defaults
             ValueProviderExecutor.Execute(Instance, new ValueProviderExecutorSettings()
-                .AddFilter(new DataMemberAttributeValueProviderFilter(false))
-                .AddDefaultValueProvider());
+                .AddFilter(new DataMemberAttributeValueProviderFilter(false)));
```
Toxantron, line 81: I am wondering whether we need this call at all now. Does it still do anything without a registered provider? Does it populate `null` objects for example?

MarisaGoergen:
That's what `DefaultValueHandling.IgnoreAndPopulate` is for.

Toxantron:
> That's what `DefaultValueHandling.IgnoreAndPopulate` is for.

That's what I had in mind, just wanted to check.

src/Moryx.Resources.Management/Resources/ResourceEntityAccessor.cs
MarisaGoergen: Really is not needed anymore, since it should do nothing

Toxantron (approved):

PR #313, opened 2023-08-01
Added GetEnum method to get Display.Name value or use enum value.

MathoMathiasCamara:
The value of the enum was used as the name. Now added a method to check if there is a Display attribute on the enum, if yes use the display value otherwise use the old behavior.

### DIFF
--- a/src/Moryx.AbstractionLayer/Tasks/TaskStep.cs
+++ b/src/Moryx.AbstractionLayer/Tasks/TaskStep.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0
 
 using System;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.Serialization;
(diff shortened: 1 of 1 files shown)


src/Moryx.AbstractionLayer/Tasks/TaskStep.cs
```diff
@@ -66,6 +67,23 @@ private static OutputDescription[] DescriptionsFromEnum(Type enumType)
...
+        /// Get the enum name. If Display attribute is applied on the enum value
+        /// the display value get used otherwise the enum value is used.
+        /// </summary>
+        /// <param name="value"> value of the enum</param>
+        /// <param name="enumType">type of the enum</param>
+        /// <returns></returns>
+        private static string GetEnumName(object value, Type enumType)
+        {
+            var enumMembers = enumType.GetMembers();
+            var enumValueInfo = enumMembers?.FirstOrDefault(x => x.DeclaringType == enumType &&
```
Toxantron, line 80: Please don't use null checks for every invocation if the argument should not be null. Each `?` adds 3 invisible lines to your source code.

Toxantron (approved):

Toxantron:
I forgot, could you add a UnitTest for that behavior (in a separate PR)

PR #319, opened 2023-08-08
Fix time conversion

MarisaGoergen:
With EF Core 6 and Npgsql 6 DateTimes can only be saved in UTC format. In order to fix that for all possible modules, the UOW will now check for DateTimes before it calles `DbContext.SaveChanges()`.

Additionally DateTimes where forgotten in the `EntryConvertGenerator`. Because of that it wasn't possible for Resources and Products to have DateTime properties. These properties would always be set to the default value.

### DIFF
--- a/src/Moryx.Model/Annotations/DateTimeKindAnnotation.cs
+++ b/src/Moryx.Model/Annotations/DateTimeKindAnnotation.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
@@ -26,7 +27,7 @@ namespace Moryx.Model.Annotations
         private static readonly ValueConverter<DateTime, DateTime> UtcConverter =
             new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Utc));
         private static readonly ValueConverter<DateTime, DateTime> LocalConverter =
-            new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Local));
+            new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => ConvertToLocal(v));
         private static readonly ValueConverter<DateTime, DateTime> UnspecifiedConverter =
            new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Unspecified));
 
@@ -38,6 +39,12 @@ namespace Moryx.Model.Annotations
                 return TimeZoneInfo.ConvertTimeToUtc(dateTime, TimeZoneInfo.Local);
         }
        
+        private static DateTime ConvertToLocal(DateTime timeUtc)
+        {
+            var result = TimeZoneInfo.ConvertTimeFromUtc(timeUtc, TimeZoneInfo.Local);
+            result = DateTime.SpecifyKind(result, DateTimeKind.Local);
+            return result;
+        }
 
         /// <summary>
         /// Adds the date time kind converter to the model builder.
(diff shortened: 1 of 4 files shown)


src/Moryx.Model/Repositories/UnitOfWork.cs
```diff
@@ -65,6 +66,23 @@ public void SaveChanges()
         {
             try
             {
+                var entries = DbContext.ChangeTracker.Entries();
```
seveneleven, line 69: I would extract this into a separate method

src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
```diff
@@ -133,17 +133,31 @@ public static object ToObject(Type type, string value, IFormatProvider formatPro
 			else if (type.IsEnum)
             {
                 result = Enum.Parse(type, value);
-            }
+            }else if(type == typeof(DateTime))
```
seveneleven, line 136: Formatting

Toxantron:
Do we have any idea, how this will impact performance. I would prefer something that more precisely targets the date time properties instead of iterating over all properties on all entities during saving.

Thinking about it: Maybe we just make it mandatory for user code to do the conversion, right at the time when needed. Yes, this will be more work, but seams like the proper solution. We already have Storage classes for most models, so that should be an isolated issue.

Toxantron:
@dbeuchler that is even better. Less redundant code and it does not iterate all properties of all entities on every write.

src/Moryx.Model/Annotations/DateTimeKindAnnotation.cs
```diff
@@ -23,13 +24,20 @@ public static class DateTimeKindAnnotation
...
+            new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Utc));
         private static readonly ValueConverter<DateTime, DateTime> LocalConverter =
-            new ValueConverter<DateTime, DateTime>(v => v, v => DateTime.SpecifyKind(v, DateTimeKind.Local));
-
+            new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Local));
         private static readonly ValueConverter<DateTime, DateTime> UnspecifiedConverter =
-            new ValueConverter<DateTime, DateTime>(v => v, v => DateTime.SpecifyKind(v, DateTimeKind.Unspecified));
+           new ValueConverter<DateTime, DateTime>(v => ConvertToUtc(v), v => DateTime.SpecifyKind(v, DateTimeKind.Unspecified));
+
+        private static DateTime ConvertToUtc(DateTime dateTime)
```
dbeuchler, line 33: Why is everything converted to UTC? In my sample if provided an option to attribute your entity properties: More flexible for the developer, he can decide what is necessary

Toxantron:
> With EF Core 6 and Npgsql 6 DateTimes can only be saved in UTC format.

We found that, do you have different information?

dbeuchler:
> > With EF Core 6 and Npgsql 6 DateTimes can only be saved in UTC format.
> 
> We found that, do you have different information?

Both is possible, but it has to be defined as I understand. In the [docs](https://www.npgsql.org/doc/types/datetime.html#:~:text=Starting%20with%206.0%2C%20Npgsql%20maps,will%20throw%20an%20exception%2C%20etc.) I found just 

> Starting with 6.0, Npgsql maps UTC DateTime to timestamp with time zone , and Local/Unspecified DateTime to timestamp without time zone` 

and

>  In versions prior to 6.0 (or when Npgsql.EnableLegacyTimestampBehavior is enabled), reading a timestamp with time zone returns a Local DateTime instead of Utc. 


> In versions prior to 6.0 (or when Npgsql.EnableLegacyTimestampBehavior is enabled), DateTime is always written as timestamp without time zone.

I am using the `Npgsql.EnableLegacyTimestampBehavior` in connection to the extension. But currently I am also not too deep in topic. 

For this MR I would say providing all options is a good idea but for the postgres integration a custom behaviour could be applied.

Toxantron:
Additionally: We had a couple of issues in the past when analyzing the database we had different timestamps for Created vs ActivityStarted with timezone gaps.

I think that's why we decided to have all timestamps as UTC, but maybe we take the annotation and just apply it to our databases first.

dbeuchler:
> Additionally: We had a couple of issues in the past when analyzing the database we had different timestamps for Created vs ActivityStarted with timezone gaps.
> 
> I think that's why we decided to have all timestamps as UTC, but maybe we take the annotation and just apply it to our databases first.

Yes I know this problem - using `DateTime.UtcNow` at every point would be a good starting point and additionally apply the annotation to all MORYX models sounds good. Forcing UTC for every model, on the other hand is not a good idea, this should be the decision of the developer what is required for his use-case.

Toxantron (approved):

PR #320, opened 2023-08-10
LoadVersions only returns the current version

MathoMathiasCamara:
Currently **LoadVersions** only returns the current version

Updated **LoadVersions** to collect both targetReferences and sourceReferences and concatenate them. Return all the versions ordered by version.

LoadVersions

### DIFF
--- a/src/Moryx.Products.Management/Implementation/RecipeManagement.cs
+++ b/src/Moryx.Products.Management/Implementation/RecipeManagement.cs
@@ -10,6 +10,7 @@ using Moryx.Container;
 using Moryx.Model;
 using Moryx.Model.Repositories;
 using Moryx.Products.Model;
+using Moryx.Tools;
 using Moryx.Workplans;
 
 namespace Moryx.Products.Management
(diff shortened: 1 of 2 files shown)


src/Moryx.Products.Management/Implementation/RecipeManagement.cs
```diff
@@ -88,31 +89,37 @@ public IReadOnlyList<Workplan> LoadVersions(long workplanId)
         {
             using var uow = ModelFactory.Create();
 
-            var versions = new List<Workplan>();
             var repo = uow.GetRepository<IWorkplanRepository>();
-            do
+            var currentWorkplan = repo.GetAll()
```
Toxantron, line 93: **Never** call `GetAll()` on a database. If you want to fetch data, filter and concat it, that do the majority of the work in Linq.

src/Moryx.Products.Management/Implementation/RecipeManagement.cs
```diff
@@ -90,8 +90,10 @@ public IReadOnlyList<Workplan> LoadVersions(long workplanId)
             using var uow = ModelFactory.Create();
 
             var repo = uow.GetRepository<IWorkplanRepository>();
-            var currentWorkplan = repo.GetAll()
-                .FirstOrDefault(entity => entity.Id == workplanId);
+            var currentWorkplan = repo.GetByKey(workplanId);
+
+            if (currentWorkplan == null) new List<Workplan>();
```
Toxantron, line 95: I think there is a `return` missing.

Toxantron (changes requested):

Toxantron:
The initial implementation always assumed the latest version is given, but from there it recursively went all the way back (do-while loop). Your current implementation looks at sources and targets, but only one level.

Toxantron:
I restored and fixed the LinQ approach but also found another issue in the process. The [model definition](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Products.Model/ProductsContext.cs#L64) of WorkplanReferences.Target => Workplan.SourceReferences is inverted and really confusing. 

image

Toxantron (approved):

PR #324, opened 2023-08-11
Provider of DB is now selected by default

enis-pixel:
By default, the SqliteModelConfigurator is selected.

### DIFF
--- a/src/Moryx.Model/DbContextManager.cs
+++ b/src/Moryx.Model/DbContextManager.cs
@@ -42,7 +42,7 @@ namespace Moryx.Model
 
                     var configuratorType = !string.IsNullOrEmpty(config.ConfiguratorTypename)
                         ? Type.GetType(config.ConfiguratorTypename)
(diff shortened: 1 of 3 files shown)


src/Moryx.Model/DbContextManager.cs
```diff
@@ -65,6 +65,16 @@ public DbContextManager(IConfigManager configManager, ILoggerFactory loggerFacto
             }
         }
 
+        private Type DefaultConfigurator()
+        {
+            var sqliteModelConfigurator = ReflectionTool.GetAssemblies()
+                .FirstOrDefault(x => x.GetName().Name.Contains("Moryx.Model.Sqlite"))
+                .GetTypes()
```
1nf0rmagician, line 72: FirstOrDefault can return `null`, i.e. you need to put a `?` before the next function call and account for `sqliteModelConfigurator` being null in the following 😊

1nf0rmagician (approved):

PR #326, opened 2023-08-14
Unresolved dependencies are removed from the dependency list of available modules

MathoMathiasCamara:
Added **MissingState**, **MissingServerModule** and **MissingModuleDependency**. **MissingServerModule** now represent module that can't be resolved, it contains a property **RepresentedService** that holds the type of the Server that is missing. And has a static state **Missing**.

**Final result**:
commandCenterMissingDependency


Toxantron (approved):
I dont like the solution, the whole fake module and the conditions are "bad taste" but also the most efficient solution on the current verion. Maybe we can improve it for MORYX 8.

Issue #136, opened 2022-01-04
TextBindingResolverFactory crashes on null template


Toxantron:
**Describe the bug**
When invoking the `TextBindingResolverFactory` with a `null` template string, it crashes.

**Expected behavior**
Binding a null, empty or binding less string, should simply return the original value on each call.

dbeuchler:
already fixed by #332

PR #332, opened 2023-08-21
When invoking the TextBindingResolverFactory with a null template string, it crashes.

MathoMathiasCamara:
Added a null check in the **Create** method, to check if the value is null, returns **NullResolver(originalValue)**. Added a test case for Null and wrong string format.
[Issue link](https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/136)

### DIFF
--- a/src/Tests/Moryx.Tests/Bindings/BindingResolverExtensionTests.cs
+++ b/src/Tests/Moryx.Tests/Bindings/BindingResolverExtensionTests.cs
@@ -125,6 +125,19 @@ namespace Moryx.Tests.Bindings
             Assert.AreEqual(5 - removeCount, CountChainLinks(_resolver));
         }
 
+        [TestCase(null, Description = "should return null value")]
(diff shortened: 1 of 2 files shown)


src/Tests/Moryx.Tests/Bindings/BindingResolverExtensionTests.cs
```diff
@@ -125,6 +125,18 @@ public void CallRemoveRemovesItemFromChain(int removeCount)
             Assert.AreEqual(5 - removeCount, CountChainLinks(_resolver));
         }
 
+        [TestCase(null, Description = "should return null value")]
+        [TestCase("This is a wrong binding", Description = "Should return original string value")]
```
Toxantron, line 129: Your invalid binding string should reference a property on `data` that does not exist or is a typo of the property.

For example `"This text used {Branch.Nmae} on data which was not found"`.

Toxantron (approved):

PR #334, opened 2023-08-23
Add possibility to exlude parts

enis-pixel:
By default, the LoadType method in the ProductMangementFacade returns the parts. Now there is an additional property called NoPartLinks in the ProductQuery class, which can be set to true to speed up the queries and exclude the parts.

### DIFF
--- a/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
+++ b/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
@@ -46,7 +46,9 @@ namespace Moryx.AbstractionLayer.Products.Endpoints
                 HasWorkplans = typeof(IWorkplanRecipe).IsAssignableFrom(recipeType)
             };
         }
-        public ProductModel ConvertProduct(IProductType productType, bool flat)
+
+        /// <param name="includeParts">Determines whether parts should be included</param>
+        public ProductModel ConvertProduct(IProductType productType, bool flat, bool includeParts = true)
         {
             
             // Base object
@@ -77,7 +79,8 @@ namespace Moryx.AbstractionLayer.Products.Endpoints
             converted.Recipes = recipes.Select(ConvertRecipe).ToArray();
 
             // Parts
-            ConvertParts(productType, properties, converted);
+            if (includeParts)
+                ConvertParts(productType, properties, converted);
 
             return converted;
         }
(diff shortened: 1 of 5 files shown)


src/Moryx.Products.Management/Implementation/Storage/ProductStorage.cs
```diff
@@ -353,6 +353,28 @@ public IReadOnlyList<IProductType> LoadTypes(ProductQuery query)
             else if (query.RecipeFilter == RecipeFilter.WithoutRecipes)
                 productsQuery = productsQuery.Where(p => p.Recipes.Count == 0);
 
+            // Exclude parts
+            if (query.NoPartLinks)
+            {
```
Toxantron, line 358: What's the point of this? The ProductsQuery never includes PartLinks in the response. It only returns the head type for the query arguments and then you need to call LoadType on each.

Copying all properties onto a new object within a LinQ query seems like bad taste. And it would break the selectors 5 lines below.

If there is a runtime issue with proxies or lazy loading, check the UnitOfWork settings at the top or the conversion at the bottom.

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
```diff
@@ -46,7 +46,9 @@ public RecipeDefinitionModel ConvertRecipeType(Type recipeType)
                 HasWorkplans = typeof(IWorkplanRecipe).IsAssignableFrom(recipeType)
             };
         }
-        public ProductModel ConvertProduct(IProductType productType, bool flat)
+
+        /// <param name="includeParts">Determines whether parts should be included</param>
+        public ProductModel ConvertProduct(IProductType productType, bool flat, bool includeParts = true)
```
Toxantron, line 51: The only difference I see between `bool flat` (meaning no properties, recipes or parts) and the new argument is to load 2/3 of the default page but not the part links. This also would have to be a new method as it would break a public API.

src/Moryx.Products.Model/Entities/ProductTypeEntity.cs
```diff
@@ -28,7 +28,7 @@ public class ProductTypeEntity : ModificationTrackedEntityBase
 
         public virtual ICollection<ProductTypePropertiesEntity> OldVersions { get; set; }
 
-        public virtual ProductTypePropertiesEntity CurrentVersion { get; protected internal set; }
```
Toxantron, line 31: The current version is updated automatically. For which requirement do you set it manually?

Toxantron (changes requested):
If there is a runtime issue with part links, it probably caused by UoW settings or accidental navigation on proxies. It also might just be an issue of SQLite wich is not optimized for this purpose. Can you confirm faulty joins on SQL systems like postgres?

Issue #245, opened 2020-06-30
Remove ResourceWrapper


Toxantron:
- Remove wrapper and state machine to reduce overhead per resource instance. 
- Combine Initialize and Start of ResourceManager, pass resources from one step to the next. 
- Split running and failed resources into two collections

PR #335, opened 2023-08-25
Move IPublicResource To IResource

MathoMathiasCamara:
Removed all IPublicResource references, and used IResource instead

related to  https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/245


Toxantron:
Isn't this a duplicate?

Issue #238, opened 2020-11-26
Remove IPublicResource restriction from ResourceManagement facade


dbeuchler:
**Is your feature request related to a problem? Please describe.**
I wanted to write a server module which should interact with drivers and I came really fast to an end. Only `IPublicResources` can be selected by the ResourceManagement facade. 

**Describe the solution you'd like**
The API should be more open and all types of resources should be selectable (e.g. drivers). 

**Describe alternatives you've considered**
My alternative is that I do not write a module. Instead I have written another resource to use the driver.

PR #336, opened 2023-08-25
Remove wrapper

MathoMathiasCamara:
refer to https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/245

- Remove wrapper and state machine to reduce overhead per resource instance.
- Combine Initialize and Start of ResourceManager, pass resources from one step to the next.


src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -97,11 +99,20 @@ private enum ResourceStartupPhase
         /// </summary>
         private readonly object _fallbackLock = new();
 
+        private List<IResource> _failedResources = new();
+        private List<IResource> _runningResources = new();
```
Toxantron, line 103: This collection is a duplicate of the graph, just less efficient.

src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -113,26 +124,57 @@ public void Initialize()
...
+            // start resources
+            _startup = ResourceStartupPhase.Starting;
+            Parallel.ForEach(Graph.GetAll(), StartResource);
+            _startup = ResourceStartupPhase.Started;
+        }
+
+        private void ResourceFailed(Resource resource, Exception e)
+        {
+            //populate the failed resources list for tracking
+            lock (_failedResources)
```
Toxantron, line 146: You only lock one collection, but work on both.

src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -113,26 +124,57 @@ public void Initialize()
...
+            {
+                if(_runningResources.Any(x => x.Id == resource.Id))
+                    _runningResources.Remove(resource);
+
+                if(!_failedResources.Any(x => x.Id == resource.Id))
+                _failedResources.Add(resource);
```
Toxantron, line 152: Resources can not switch between the two and iterating the entire list (`Any`) for every resource will kill boot times for large applications (>5000 resources).

src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -150,8 +192,21 @@ private void LoadResources(ICollection<ResourceEntityAccessor> allResources)
...
+                RegisterEvents(resource);
+        }
+
+        private void ResourceStarted(IResource resource)
+        {
+            lock (_runningResources)
+            {
+                //check if the resource exist in failed resources
+                if (_failedResources.Any(x => x.Id == resource.Id))
+                    _failedResources.Remove(resource);
```
Toxantron, line 220: Same as above.

src/Moryx.Resources.Management/Resources/IResourceManager.cs
```diff
@@ -31,5 +31,11 @@ internal interface IResourceManager : IInitializablePlugin
         /// Raised when the capabilities have changed.
         /// </summary>
         event EventHandler<ICapabilities> CapabilitiesChanged;
+
+
+        /// <summary>
+        /// Initialize and start all the resources available
+        /// </summary>
+        void InitializeAndStart();
```
Toxantron, line 39: Nobody calls this new method.

Toxantron (changes requested):
When removing `IPublicResource` you forgot a couple of lines that only applied to the differentiation of the two. Read over the code again, identify what happend before and bring it down to `IResource` only. The idea was a simplification, not just replacing all usages.

When removing the wrapper, consider the resource graph, the lifecycle and the standard case. Keep the list of failed resources and move any failed (should be only a few if any) in there. Take a look at the conditions for `GetResources` and make sure, that failed resources are filtered out, e.g. `graph.Where(...).Except(_failedResources)...`

Also check if the behavior of resources proxies still works correctly as the ProxyBuilder uses reflection and System.Emit so breaking changes might not be obvious.

src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Castle.Core.Resource;
```
Toxantron, line 8: This reference does not belong here

src/Moryx.Resources.Management/Resources/ResourceManager.cs
```diff
@@ -150,8 +196,22 @@ private void LoadResources(ICollection<ResourceEntityAccessor> allResources)
...
+        /// <summary>
+        /// Handles the resource started by adding it to the list of running resources and removes it from failed resources
+        /// </summary>
+        /// <param name="resource">Started resource</param>
+        private void HandleResourceStarted(IResource resource)
+        {
+            lock (_failedResources)
+            {
+                //check if the resource exist in failed resources
+                if (_failedResources.FirstOrDefault(x => x.Id == resource.Id) is not null)
```
Toxantron, line 212: If is not necessary. Just call `Remove`, it returns false if no entry was found.

But in fact: We talked about reducing the life cylce, so a resource can not be failed before initialize and it will never reach Start if it failed in Initialize.

We only need the failed collection when responding to requests on the facade.

src/Moryx.Resources.Management/Resources/ResourceProxy.cs
```diff
@@ -17,6 +19,12 @@ internal abstract class ResourceProxy : IResource
         /// </summary>
         private IResourceTypeController _typeController;
 
+        public event EventHandler<ICapabilities> CapabilitiesChanged
+        {
+            add { Target.CapabilitiesChanged += value; }
```
Toxantron, line 24: Generally a good idea, but you can not forward the registration as it would pass the non-proxy object as sender to the listener.

Toxantron (changes requested):

1nf0rmagician:
Closes #238

PR #341, opened 2023-09-14
If you put a EntrySerialize over a Resource itself and also use it in one of the properties, there will be a StackOverflow Exception

MathoMathiasCamara:
- Added a check to ignore self reference properties
- Added a check to ignore EntrySerializeAttribute from overriding the attribute on the base class.


src/Moryx/Serialization/EntryConvert/EntryConvert.cs
```diff
@@ -277,6 +277,17 @@ public static Entry EncodeObject(object instance, ICustomSerialization customSer
                 try
                 {
                     value = property.GetValue(instance);
+
+                    //check if this property is a self reference (Descriptor)
+                    Type propertyType = value?.GetType();
+                    var flag = propertyType == instanceType;
```
MarisaGoergen, line 283: Is a vlaue check also needed?

src/Moryx/Serialization/EntryConvert/EntryConvert.cs
```diff
@@ -271,12 +271,27 @@ public static Entry EncodeObject(object instance, ICustomSerialization customSer
             var filtered = customSerialization.GetProperties(instance.GetType());
             foreach (var property in filtered)
             {
+                //exclude property of the same type as the object we are encoding
+                var propertyType = property.PropertyType;
+                if (propertyType == instanceType) continue;
```
MarisaGoergen, line 276: if you do it like this, it's not possible to edit Properties of the same type as the instance in the UI

MarisaGoergen (changes requested):
The UI shouldn't show even the resource graph if you put entry serialize on the class.

src/Moryx/Serialization/EntryConvert/EntryConvert.cs
MathoMathiasCamara: If you don't do this it lead to stack overflow cause you endup with a circular relation. class->property->class->property...etc

MathoMathiasCamara: If you have a suggestion on how to handle the circular relationship, i am listening.

MarisaGoergen:
Why don't we just declare that it's not allowed to use `EntrySerialize` over classes? @Toxantron @dbeuchler

dbeuchler:
> Why don't we just declare that it's not allowed to use `EntrySerialize` over classes? @Toxantron @dbeuchler 

I have not read the full merge request but: the EntrySerializeAttribute was developed to define it on classes, methods and properties. If you write it on a class, the rules will be applied for the whole class except the member is modified by the attribute itself. You can also declare "Never" on a the class and just add members explicitly.

See line 123: https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx/Serialization/EntrySerializeSerialization.cs

As I remember: The Resource-Type declares EntrySerialize(Never). This should not be overridden by the developer. https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.AbstractionLayer/Resources/Resource.cs

Toxantron:
> As I remember: The Resource-Type declares EntrySerialize(Never). This should not be overridden by the developer.

We should defenitly prohibit the override. For MORYX 8 we could even forbid it on the level of attribute declaration. There is an option "AllowMultiple = false".

src/Moryx/Serialization/DefaultSerialization.cs
```diff
@@ -34,7 +34,12 @@ public DefaultSerialization()
         /// <see cref="ICustomSerialization"/>
         public virtual IEnumerable<PropertyInfo> GetProperties(Type sourceType)
         {
-            return sourceType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
+            var properties = sourceType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
+            var hasEntrySerializeAttribute = sourceType.GetCustomAttribute(typeof(EntrySerializeAttribute)) != null;
+            //ignore the entry serialize attribute on a class
+            return hasEntrySerializeAttribute ?
+                properties.Where(x => x.GetCustomAttributes(typeof(EntrySerializeAttribute)).Count() > 0) :
```
Toxantron, line 41: I think this method get's called for all layers. We can not prohibit it in general, as this would break multiple usages down the tree.

src/Moryx/Serialization/EntryConvert/EntryConvert.cs
Toxantron: This will not prevent recursion anyway. You could still do "A->B->A->B"...

src/Moryx/Serialization/DefaultSerialization.cs
Toxantron: Also, this type of serialization shoudl not check for `EntrySerialize` at all. There is a dedicated implementation with a more sophistacated check.

Toxantron:
Please keep in mind, there is a dedicated [EntrySerializeSerialization](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx/Serialization/EntrySerializeSerialization.cs). We should not change the behavior of the DefaultSerialization.

Instead we should isolate this either as a double declaration check in EntrySerializeSerialization and just checking for EntrySerizalize in ResourceSerialization.

Toxantron (changes requested):
Please keep in mind, there is a dedicated [EntrySerializeSerialization](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx/Serialization/EntrySerializeSerialization.cs). We should not change the behavior of the DefaultSerialization.

Instead we should isolate this either as a double declaration check in EntrySerializeSerialization and just checking for EntrySerizalize in ResourceSerialization.

src/Moryx/Serialization/DefaultSerialization.cs
MathoMathiasCamara: @Toxantron  Are you referring to the `ResourceSerialization` or what?

src/Moryx/Serialization/EntrySerializeSerialization.cs
```diff
@@ -120,6 +120,12 @@ private static IEnumerable<PropertyInfo> EntrySerializeAttributeFilter(Type sour
                 .Where(pm => pm.Mode.HasValue && pm.Mode == EntrySerializeMode.Never)
                 .Select(pm => pm.Property).ToArray();
 
+            //Class: Always
+            // if there are Always properties return those properties only
+            if ((sourceTypeMode.HasValue && sourceTypeMode.Value == EntrySerializeMode.Always) &&
```
Toxantron, line 125: While this might fix the Resource error, it is a behavior change compared to the block below. Considering that the base class, already declares "Never", I propose that we only check for duplicate declaration of the attribute and give preference to the initial declaration, assuming that the base class should define the behavior.

Toxantron (changes requested):
Instead of changing the behavior, just improve the code that reads the class property. This should give preference to the base class declaration over the incorrect override.

For MORYX 8 I suggest we declare the attribute to prohibit duplicate declaration.

Toxantron (approved):

PR #345, opened 2023-10-10
Add API to fluently initialize resources

seveneleven:
`ResourceInitializerExtensions` provide a fluent API to create code for resource initialization in a compact way.


1nf0rmagician (approved):
I like it 👍

PR #351, opened 2023-10-29
Improve (tidy up) commandcenter navigation

seveneleven:
### Move module specific navigation from left menu to module pane

Left menus module navigation only navigates to modules. I.e. modules have
their own navigation bar to reach overview, configuration and console
sections.

This eliminates many clicks for navigating/expanding the tree on the left
navigation

Before
image
After
image

### Improve left navigation menu for modules and databases

* Icons have been removed to gain more space for identifiers/names
* Clicking menu items will navigate to targets, no matter where they've
  been clicked. Worked only with the link text before.
* Database
  * Context name and namespace have been separated to put more emphasize
    on the context name while keeping the namespace without wasting space
    for it (and risking ugly word wrap).

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/Moryx.CommandCenter.Web/src/common/components/Menu/RoutingMenu.tsx
+++ b/src/Moryx.CommandCenter.Web/src/common/components/Menu/RoutingMenu.tsx
@@ -22,12 +22,13 @@ class RoutingMenu extends React.Component<RouteComponentProps<{}> & MenuProps, {
         if (this.props.onActiveMenuItemChanged != null) {
             this.props.onActiveMenuItemChanged(menuItem);
         }
+        this.props.history.push(menuItem.NavPath);
     }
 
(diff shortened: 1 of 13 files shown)


Toxantron:
I like this a lot. Should we release it as a minor (6.3) oder rather just merge into future as an appetizer.

Toxantron (approved):

1nf0rmagician:
> I like this a lot. Should we release it as a minor (6.3) oder rather just merge into future as an appetizer.

Since we promised features to be released for MORYX 6 up until May 2024 and since it does not require a bit of extra work to make in minor conform, I would release it as a minor 🤔

Toxantron:
> I would release it as a minor 🤔

Glad you think so, already merged

PR #392, opened 2024-03-05
Update/ci package feed

seveneleven:
Paths to package feeds have changed. With that, `README.md` was updated.

Package feed and version info moved to top...
image

... and have been removed from Package List:

image


1nf0rmagician (dismissed):

1nf0rmagician:
The future badge shows now result, is it still possible to differentiate the labels if we only use one feed? If not we should merge the badges i guess 😊

seveneleven:
> The future badge shows now result, is it still possible to differentiate the labels if we only use one feed? If not we should merge the badges i guess 😊

Actually, it's the badge for the dev branch, which doesn't work properly (The screenshot was taken, when there was no future package on the feed). But with that, it's hard to get the latest dev version now, since future is the most recent version now. 

I'd just remove both for now, until we have a good solution for that

1nf0rmagician (dismissed):

1nf0rmagician (approved):

1nf0rmagician (approved):

PR #432, opened 2024-06-19
Add integration test tools for MORYX modules

1nf0rmagician:
Adding a `MoryxTestEnvironment` as well as a tutorial on how to use it for integration testing of modules.

### DIFF
--- a/docs/tutorials/HowToTestAModule.md
+++ b/docs/tutorials/HowToTestAModule.md
@@ -11,16 +11,16 @@ The following example shows a setup for the `IShiftManagement` facade interface.
 private ModuleConfig _config;
 private Mock<IResourceManagement> _resourceManagementMock;
 private Mock<IOperatorManagement> _operatorManagementMock;
-private MoryxTestEnvironment<IShiftManagement> _env;
+private MoryxTestEnvironment _env;
 
 [SetUp]
 public void SetUp()
 {
     ReflectionTool.TestMode = true;
     _config = new();
-    _resourceManagementMock = MoryxTestEnvironment<IShiftManagement>.CreateModuleMock<IResourceManagement>();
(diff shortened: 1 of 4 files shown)


Toxantron:
I removed the class generic as it makes things unncessarily complicated. It was only used to extract the module from the service collection via its facade registration. Since modules are [registered under their type](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs#L57) this could be simplified.

1nf0rmagician:
I like it, adjusted the tutorial to match the non generig class

docs/tutorials/HowToTestAModule.md
```diff
@@ -0,0 +1,45 @@
+# Setup a test environment for integration tests of a module
+
+In order to test a module in its lifecycle with its respective facade we offer the `Moryx.TestTools.IntegrationTest`.
+The package brings a `MoryxTestEnvironment<T>`.
+With this class you can first create mocks for all module facades your module dependents on using the static `CreateModuleMock<FacadeType>` method.
```
seveneleven, line 5: Not in sync with the code?

src/Moryx.TestTools.IntegrationTest/MoryxTestEnvironment.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// </summary>
+    /// <typeparam name="T">Type of the facade to be tested.</typeparam>
+    public class MoryxTestEnvironment
+    {
+        private readonly Type _moduleType;
+
+        public IServiceProvider Services { get; private set; }
+
+        /// <summary>
+        /// Creates an <see cref="IServiceProvider"/> for integration tests of moryx. We prepare the
```
seveneleven, line 33: 'It' prepares?

Toxantron (approved):

seveneleven (commented):
I like it!

PR #433, opened 2024-07-17
NUnit Test Tools

andreniggemann:
## General idea

Given that test are often run in CICD Pipelines it is important to be able to interpret the test failures given the test reports.
To test complex behaviors properly we often have multiple assertions in a single test.

Unless a message is manually specified on an assert, the test report the report will have no meaningful details about the condition that failed and it's possible you have to rely on the line number to find the failed assertion.
Manual messages have further drawbacks like copy paste mistakes or not being maintained once the condition is changed.

The idea to solve this is to create a drop in replacement for NUnit `Assert.That` that gives more detailed test results.
It relies on the `CallerArgumentExpression` Attribute to get include the expression being tested in the test report.

The only scenario I know off where this is not a drop in replacement, is when the Assert has a message and format parameters.

## Changed behavior for functions/delegates
One maybe controversial, but in my eyes very helpful, aspect, is the handling of functions being passed into the test expression.
The test will implicitly Assert that the function does not throw and then check the return value with the given constraints.
This is helpful in `Assert.Multiple` blocks because later assertions can still be evaluated.

I think @seveneleven had a good analogy for this:
Given a text input field with validation logic it is very annoying if only the first broken rule is shown. 
We have to fix one rule at a time, only to discover the next broken rule. 

Given this functionality we only need to wrap expressions that can fail into a lambda and still check all the rules at the same time.


seveneleven (approved):
I like this feature a lot!

1nf0rmagician (approved):

PR #434, opened 2024-07-18
load type from database before raising TypeChange event

MathoMathiasCamara:
TypeChanged event does not provide "singleton" types when saving via UI

- load product type form database before raising the TypeChangedEvent
- update how to set references if a part was: `added` `modified` and `removed`

### DIFF
--- a/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
+++ b/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
@@ -270,16 +270,21 @@ namespace Moryx.AbstractionLayer.Products.Endpoints
                     continue;
 
                 var match = unused.Find(r => r.Id == partModel?.Id);
+                // new partlink
                 if (match == null)
                 {
                     match = (IProductPartLink)Activator.CreateInstance(elemType);
+                    match.Product = _productManagement.LoadType(partModel.Product.Id);
                     value.Add(match);
                 }
+                //modified reference
+                else if (match.Product.Id != partModel.Product.Id)
+                    match.Product = _productManagement.LoadType(partModel.Product.Id);
                 else
+                 //part removed
                     unused.Remove(match);
 
                 EntryConvert.UpdateInstance(match, partModel.Properties);
-                match.Product = _productManagement.LoadType(partModel.Product.Id);
             }
 
             // Clear all values no longer present in the model
--- a/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
+++ b/src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
@@ -258,7 +258,7 @@ namespace Moryx.AbstractionLayer.Products.Endpoints
         private void UpdateCollection(IList value, IEnumerable<PartModel> parts)
         {
             // Track which part links are still represented by the models
-            var unused = new List<IProductPartLink>(value.OfType<IProductPartLink>());
+            var oldParts = new List<IProductPartLink>(value.OfType<IProductPartLink>());
             // Iterate over the part models
             // Create or update the part links
             var elemType = value.GetType().GetInterfaces()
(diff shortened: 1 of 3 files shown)


src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
```diff
@@ -270,16 +270,21 @@ private void UpdateCollection(IList value, IEnumerable<PartModel> parts)
                     continue;
 
                 var match = unused.Find(r => r.Id == partModel?.Id);
```
dacky179, line 272: partmodel is never null

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
```diff
@@ -270,16 +270,21 @@ private void UpdateCollection(IList value, IEnumerable<PartModel> parts)
...
                     match = (IProductPartLink)Activator.CreateInstance(elemType);
+                    match.Product = _productManagement.LoadType(partModel.Product.Id);
                     value.Add(match);
                 }
+                //modified reference
+                else if (match.Product.Id != partModel.Product.Id)
+                    match.Product = _productManagement.LoadType(partModel.Product.Id);
                 else
+                 //part removed
                     unused.Remove(match);
```
dacky179, line 285: please reconsider name "unused": a removed part will be removed from the unused list and then we remove all remaining in the unused list to remove them from the value list? it making me headache to read this methode. ore maybe i do not unterstand it and it needs comments...

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
```diff
@@ -269,22 +269,27 @@ private void UpdateCollection(IList value, IEnumerable<PartModel> parts)
...
                 {
-                    match = (IProductPartLink)Activator.CreateInstance(elemType);
-                    value.Add(match);
+                    oldPartMatch = (IProductPartLink)Activator.CreateInstance(elemType);
+                    oldPartMatch.Product = _productManagement.LoadType(partModel.Product.Id);
+                    value.Add(oldPartMatch);
                 }
+                //modified reference
+                else if (oldPartMatch.Product.Id != partModel.Product.Id)
+                    oldPartMatch.Product = _productManagement.LoadType(partModel.Product.Id);
```
dacky179, line 282: Sorry if its a double post: cannot find my posts. 
Remove from oldParts list otherwise changed instance would be deleted at the end. I think this code was never executed otherwise i cannot explain why this never causes problems...

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
```diff
@@ -269,22 +269,27 @@ private void UpdateCollection(IList value, IEnumerable<PartModel> parts)
...
                 else
-                    unused.Remove(match);
+                    //part removed
```
dacky179, line 284: better comment: 
// existing unchanged partlink: do not delete at the end

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
dacky179: @Toxantron, did we ever replace a product of a partlink without creating a new partlink?

Toxantron: Good question. I don't think so, but that is only the current definition of the endpoint. There is nothing preventing someone from replacing a reference on an exisiting part link through the facade.

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
Toxantron: I agree

Toxantron (approved):

src/Moryx.AbstractionLayer.Products.Endpoints/ProductConverter.cs
dacky179: This was not fixed: _**Remove from oldParts list otherwise changed instance would be deleted at the end.**_ @MathoMathiasCamara

Toxantron: Damn, you are right.

PR #441, opened 2024-08-26
Port Fix repository validation that failes for inverted type list to MORYX 8

Toxantron:
Port #437


1nf0rmagician:
Duplicate #442

PR #460, opened 2024-09-13
Avoid NullReferenceExceptions in ValueProviderExecutor

slimberg91:
We add a `Tracing` to our activities. This information is also stored at some processes. When accessing the process information of a running/recreated order via the order ui, the processes list stays empty. This happens due to a NullReferenceException when recreating the tracings of the processes. The problem is, that the `Tracing` class contains nullable DateTime properties. When prototypes will be created, may by `Activator.CreateInstance()`, then nullable properties will contain the value `null` which has leaded to the exception when calling FilterProperties for that null value. The newer code will avoid this by null-checking. I applied the new code here as a backport.


Toxantron (approved):

Toxantron:
Thank you for your contribution. Please note, that MORYX 3 will reach it's end of life in 2025.

PR #476, opened 2024-11-07
Fix exception on creating a bidirectional link to a new resource

1nf0rmagician:
When creating a bidirectional relation the missing properties on the relation to be created caused an exception when creating the backlink.

Also fixes display name and description serialization on resource types in the controller

### DIFF
--- a/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
+++ b/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
@@ -197,7 +197,7 @@ namespace Moryx.AbstractionLayer.Resources.Endpoints
                 Creatable = node.Creatable,
                 Name = node.Name,
                 BaseType = baseType?.Name,
-                DisplayName = resType.GetDisplayName(),
+                DisplayName = resType.GetDisplayName() ?? Regex.Replace(resType.Name, @"`\d", string.Empty),
                 Description = resType.GetDescription(),
                 // Convert resource constructors
                 Constructors = node.Constructors.Select(ctr => EntryConvert.EncodeMethod(ctr, Serialization)).ToArray()
--- a/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
+++ b/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
@@ -76,7 +76,7 @@ namespace Moryx.AbstractionLayer.Resources.Endpoints
                 Description = current.Description,
 
                 // Use simplified type reference
-                Type = current.ResourceType()
+                Type = current.ResourceType(),
             };
 
             // Set partial flag or load complex properties depending on details depth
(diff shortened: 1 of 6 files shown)


1nf0rmagician:
FYI: @dacky179

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
```diff
@@ -197,14 +197,8 @@ internal ResourceTypeModel ConvertType(IResourceTypeNode node, ResourceTypeModel
...
                 BaseType = baseType?.Name,
-
-                // Read display name of the type otherwise use type short name
-                DisplayName = resType.GetCustomAttribute<DisplayNameAttribute>(false)?.DisplayName ??
-                              Regex.Replace(resType.Name, @"`\d", string.Empty),
-
-                // Read description of the type
-                Description = resType.GetCustomAttribute<DescriptionAttribute>(false)?.Description,
-
+                DisplayName = resType.GetDisplayName(),
```
Toxantron, line 200: What about the Fallback value based on TypeName?

src/Moryx.Resources.Management/Resources/ResourceLinker.cs
```diff
@@ -404,14 +404,22 @@ private static void UpdateRelationEntity(ResourceEntity resource, ResourceEntity
         {
             if (att.Role == ResourceReferenceRole.Source)
             {
+                relEntity.SourceId = referencedResource.Id;
                 relEntity.Source = referencedResource;
+                relEntity.TargetId = resource.Id;
                 relEntity.Target = resource;
                 relEntity.SourceName = att.Name;
+                // When saving a reference we filter existing references by name, if the reference is traversed from the oposite direction it is not included
+                relEntity.TargetName = att.Name;
```
Toxantron, line 413: Wait, the attributes name will not be the SourceName and the TargetName.

Toxantron (changes requested):
I don't see "the fix". I trust you replicated the issue in a test and fixed it, I just don't understand how setting SourceName **and** TargetName to the same value fixes it. It's only because your Test case uses the same name from both directions, but that is not real world usage.

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
1nf0rmagician: That is what the `ResourceTypeModel.Name` property is for, I don't think we should duplicate that information, when the UI want's to use the default value thats alright, but setting it in the endpoint seems not right for me 🤔

Toxantron: Be that as it may, isn't it still a breaking change. If the UI used the display value reyling on the server it now shows empty

1nf0rmagician: Mhh true, but I don't like that we did this in the first place... 😑

Toxantron (approved):

PR #481, opened 2024-11-19
Add missing type constraint on `ResourceInitializerExtensions.Configure` method

### LINKED ISSUE #475: ResourceInitializerExtensions provide extensions for object
**Describe the bug**

The public `ResourceInitializerExtensions` adds extensions to object instead of limit to `Resource`. A `where` clause should be added: `where TResource : Resource`

````
public static class ResourceInitializerExtensions
{
    /// <summary>
    /// Allows to configure a <see cref="Resource"/> with method chaining
    /// </summary>
    /// [...]
    public static TResult Configure<TResult>(this TResult configurable, Action<TResult> onConfigure)
    {
        onConfigure(configurable);
        return configurable;
    }
[...]
}
````

1nf0rmagician:
Resolves #475


1nf0rmagician:
True, I went with the assumption that this is actually a bug, i.e. unintended and hence would allow for such a change. If we can't/don't want to change it in this way, we should just close #475, shouldn't we?

PR #483, opened 2024-11-19
Allow DeadEndValidator to pass on workplans with disconnected connectors

### LINKED ISSUE #104: DeadEndValidation of workflow fails if connector exists but is not connencted to input nor output
**Describe the bug**
The DeadEnd validation of the workflow fails if a connector exists in the model but is not connected to input nor output.

**To Reproduce**
1. Add connector to workflow (programatically)
2. Create a valid workflow but do not use connector
3. Validate workflow

**Expected behavior**
The unused connector should be ignored

1nf0rmagician:
Resolves #104


Toxantron:
I think we should limit it to "dangling" intermediate connectors if the rest of the workplan is still complete. Unused entries, especially Start create a DeadEnd by definition.

PR #487, opened 2024-11-21
Add `FunctionResult<TResult>` to allow functions return 'proper' results

seveneleven:
Add generic `FunctionResult<TResult>` that allows functions to always return a proper result, that either contains a valid value or an error and helps exercising error handling. Tests and comments should be quite descriptive in how it works. 

This PR is the essence of multiple similar but more specialized implementations I'd like to keep from spreading. While I see already a few more things to add, I'd like to hold them back until there will be a use case for them.

It bears some discussion, in speacial for 

* Naming
* Is it in the right place?
* Did I miss an important(!) feature?

### DIFF
--- a/src/Moryx/Tools/FunctionResult.cs
+++ b/src/Moryx/Tools/FunctionResult.cs
@@ -1,0 +1,246 @@
+﻿// Copyright (c) 2023, Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System;
+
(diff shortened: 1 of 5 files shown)


1nf0rmagician (dismissed):
I like it, took me a moment to appreciate it. But I think a good example for a use case would be the IdentifierProvider module. I already created all the response types, but making them inherit from this as a base class could improve the usability of the types in code 👍🤗

1nf0rmagician (approved):

PR #493, opened 2024-11-26
Fix test steps in build pipeline

seveneleven:
Test projects now reference `coverlet.collector` to create test coverage reports.

Version 5.8.0 is used since newer versions result in build errors because some dependency to `Newtonsoft.Json` couldn't be resolved.

Notice, that 5.8 does not exist and gets resolved to 6.0.0 instead. But using 6.0.0 gave the same dependency issues described above.

Providing an output path doesn't seem to work as before, so test files get copied manually now.

The artifacts plugin was configured to report an error if not test results were found: No test results found does either mean there are not tests in the project, so no need to run this step anyway or that for some reason those results could not be generated which is considered an error

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Upload test results
         uses: actions/upload-artifact@v4
         with:
-          name: test-results
+          name: integration-test-results
           path: artifacts/Tests/
           if-no-files-found: error
           retention-days: 1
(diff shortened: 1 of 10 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -105,6 +107,7 @@ jobs:
         with:
           name: test-results
```
1nf0rmagician, line 108: i guess this needs to be something like `integration-test-results` as the pipeline complains that an artifact with that name already exists 🙈

PR #498, opened 2024-12-03
Extend construct method for constructor invocations on resources

1nf0rmagician:
Before, we created new resources using reflection and returned them always unsafed. This does not allow resources to act on the resource graph within their constructor methods, however. Hence, we added a conditional invocation to the resource modification interface, which creates a new resource in the resource graph with all required initializations. The behaviour of the endpoint is thereby extended with the option to create such resources with their respective constructors.

### DIFF
--- a/src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
+++ b/src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
@@ -18,6 +18,7 @@ using Moryx.Runtime.Modules;
 using Moryx.Runtime.Container;
 using Moryx.Configuration;
 using System.Runtime.Serialization;
+using System.ComponentModel.DataAnnotations;
 
 namespace Moryx.AbstractionLayer.Resources.Endpoints
 {
@@ -33,8 +34,8 @@ namespace Moryx.AbstractionLayer.Resources.Endpoints
         private readonly IResourceTypeTree _resourceTypeTree;
         private readonly ResourceSerialization _serialization;
 
-        public ResourceModificationController(IResourceManagement resourceManagement, 
-            IResourceTypeTree resourceTypeTree, 
+        public ResourceModificationController(IResourceManagement resourceManagement,
+            IResourceTypeTree resourceTypeTree,
             IModuleManager moduleManager,
             IServiceProvider serviceProvider)
         {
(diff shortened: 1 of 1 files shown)


src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
```diff
@@ -144,30 +145,48 @@ public ActionResult<Entry> InvokeMethod(long id, string method, Entry parameters
...
+        private ActionResult<ResourceModel> Construct(string type, MethodEntry method)
+        {
+            try
+            {
+                var id = _resourceManagement.Create(_resourceTypeTree[type].ResourceType, r => EntryConvert.InvokeMethod(r, method, _serialization));
+                return GetDetails(id);
+            }
+            catch (Exception e)
+            {
+                if (e is ArgumentException or SerializationException or ValidationException)
```
seveneleven, line 184: Would it be possible to validate the input at the beginning?

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
```diff
@@ -187,11 +206,11 @@ public ActionResult<ResourceModel> Save(ResourceModel model)
                     resourcesToSave.Skip(1).ForEach(id => _resourceManagement.Modify(id, r => true));
                 });
 
-                return Ok(GetDetails(id));
+                return GetDetails(id);
```
seveneleven, line 209: Why did you remove the explicit `Ok` here? I like the readability aspect of it.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
1nf0rmagician: Unfortunately that is not so easy. The validation logic is nested in the resource management, we would either need to pull that out of it and only use it there and here, or provide a method on the facade. But that is not part of this PR, I just added the catch block here because I stumbled upon the issue by accident. This way you at least get the correct response code 🙈

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
1nf0rmagician: Yeah that is why I added it in a recent commit, problem is GetDetails is already a controller action that returns Ok(resourceModel), i.e. wrapping it in another Ok() call causes the response object to look like this
```json
{
    result: null,
    value: // the actual resource model that should be serialized
}
```
This is not detected by the ui compiler or swagger, but causes errors when saving resources for example.

seveneleven (approved):

PR #501, opened 2024-12-05
Port: Extend construct method for constructor invocations on resources

1nf0rmagician:
Ports #498


src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceManagementController.cs
```diff
@@ -142,30 +143,48 @@ public ActionResult<Entry> InvokeMethod(long id, string method, Entry parameters
...
+        private ActionResult<ResourceModel> Construct(string type)
         {
-            var resource = (Resource)Activator.CreateInstance(_resourceTypeTree[type].ResourceType);
-            if (resource is null)
+            Resource resource;
+            try
+            {
+                resource = (Resource)Activator.CreateInstance(_resourceTypeTree[type].ResourceType);
+            }
+            catch (Exception)
```
seveneleven, line 159: Apart from the fact, that I wonder why NotFound is returned for something that presumably did not exist (i.e. wasn't asked for), and what I wouldn't change within this PR: Wouldn't it be better to distinguish errors here and return a 500 or better reason for everything that does not fall into the 'not found' category?

1nf0rmagician: Yes that would definitively be better, I just saw the if-null statement and thought, well this will never happen because a cast of null will through an exception, but then I concentrated on the other construct method 🙈

seveneleven (approved):

PR #510, opened 2024-12-10
Filter generic base types when seaching for the resource proxy type

1nf0rmagician:
When inheriting from a generic base type without adding additional interfaces, a resource can still have a generated proxy. However, in that case the proxy builder would assume the generic base type to be the best possible fit, which cannot be proxified.

Also, adds pre-set package file versions from major version.

__Note__: As described in #343 we currently don't plan to support generic resource proxies, this only prevents falsely assumed generic types

### DIFF
--- a/src/Moryx.Resources.Management/Resources/ResourceTypeController.cs
+++ b/src/Moryx.Resources.Management/Resources/ResourceTypeController.cs
@@ -264,17 +264,19 @@ namespace Moryx.Resources.Management
         /// </summary>
         private void ProvideProxyType(Type resourceType)
         {
-            // Step 1: Find the least specific base type that offers the same amount of interfaces
+            // Step 1: Find the least specific base type that offers the same amount of interfaces and is not a generic itself
             // ReSharper disable once AssignNullToNotNullAttribute -> FullName should be not null
             var targetType = _typeCache[resourceType.ResourceType()];
             var linker = targetType;
 
             var interfaces = RelevantInterfaces(linker);
-            // Move up the type tree until the parent offers less interfaces than the current linker
-            while (linker.BaseType != null && interfaces.Count == RelevantInterfaces(linker.BaseType).Count)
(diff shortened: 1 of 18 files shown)


src/Moryx.Resources.Management/Resources/ResourceTypeController.cs
```diff
@@ -264,17 +264,17 @@ private ResourceProxy InstantiateProxy(string typeName, Resource instance)
...
             var linker = targetType;
 
             var interfaces = RelevantInterfaces(linker);
-            // Move up the type tree until the parent offers less interfaces than the current linker
-            while (linker.BaseType != null && interfaces.Count == RelevantInterfaces(linker.BaseType).Count)
-            {
+            // Move up the type tree until the parent offers less interfaces than the current linker, is abstract or a generic
+            while (linker.BaseType != null && linker.BaseType.Creatable 
+                && !linker.BaseType.ResourceType.IsGenericType 
+                && interfaces.Count == RelevantInterfaces(linker.BaseType).Count)
```
seveneleven, line 276: I'd like to keep brackets here or use some other 'tool' as it's hard to reason where the block starts (and where the condition ends)

src/Tests/Moryx.Resources.Management.Tests/Mocks/ResourceWithGenericMethod.cs
```diff
@@ -57,4 +57,10 @@ public void RaiseEvent()
             throw new NotImplementedException();
         }
     }
+
+    public interface IGenericBaseResourceInterface : IResource { }
+
+    public class GenericBaseResource<T> : Resource, IGenericBaseResourceInterface { }
+
+    public class NonGenericInheritingResource : GenericBaseResource<object> { }
```
seveneleven, line 65: InheritingFromGenericResource?

seveneleven (dismissed):

seveneleven (approved):

Toxantron (approved):

PR #512, opened 2024-12-18
Port: Filter generic base types when seaching for the resource proxy type

1nf0rmagician:
Ports #510

### DIFF
--- a/.build/Common.props
+++ b/.build/Common.props
@@ -1,16 +1,16 @@
 <Project>
 
   <PropertyGroup>
-    <AssemblyVersion>6.0.0</AssemblyVersion>
+    <AssemblyVersion>8.0.0</AssemblyVersion>
     <AssemblyVersion Condition="'$(MORYX_ASSEMBLY_VERSION)'!=''">$(MORYX_ASSEMBLY_VERSION)</AssemblyVersion>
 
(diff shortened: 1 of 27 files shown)


src/Moryx.AbstractionLayer/Drivers/Camera/ICameraDriver.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Moryx.AbstractionLayer.Drivers;
```
seveneleven, line 1: I wonder how this makes it into this PR, that's already merged 🤔

.build/Common.props
```diff
@@ -1,16 +1,16 @@
 <Project>
 
   <PropertyGroup>
-    <AssemblyVersion>0.0.0</AssemblyVersion>
```
Toxantron, line 4: Here this should be 8 now, right?

src/Moryx.AbstractionLayer/Drivers/Camera/ICameraDriver.cs
1nf0rmagician: We found this to happen from time to time, Github just recognized changes from older merges that actually don't change anything🙈

Toxantron (approved):

PR #516, opened 2025-01-15
Implement new .NET8 specific attributes to entry-format validation

marierin:


src/Moryx.CommandCenter.Web/src/modules/models/EntryValidation.ts
```diff
@@ -1,12 +1,16 @@
...
+*/
+
+export default class EntryValidation {
+    public minimum: number;
+    public maximum: number;
+    public regex: string;
+    public isRequired: boolean;
+    public isPassword: boolean;
+    public deniedValue: string[];
+    public allowedValue: string[];
```
1nf0rmagician, line 13: For this there is already a possible Values attribute. We should make the generated entry equivalent between using AllowedValues and our Custom PossibleValues

src/Moryx/Serialization/DefaultSerialization.cs
```diff
@@ -1,322 +1,322 @@
...
-
-        /// <see cref="ICustomSerialization"/>
-        public virtual string[] PossibleElementValues(Type memberType, ICustomAttributeProvider attributeProvider)
-        {
-            var elementType = EntryConvert.ElementType(memberType);
-            return PossibleValues(elementType, attributeProvider);
-        }
-
-        /// <see cref="ICustomSerialization"/>
-        public virtual EntryValidation CreateValidation(Type memberType, ICustomAttributeProvider attributeProvider)
```
1nf0rmagician, line 95: I would have expected changes in this method, might be however that they are just hidden behind the encoding change and I overlooked them

src/Moryx.CommandCenter.Web/src/modules/models/EntryValidation.ts
1nf0rmagician, line 1: As you know, the command center is a web project building on your changes and hence it is not an angular project it unfortunately does not share the entry editor with our commercial web projects. 
Therefore, it would be good if you would check that the new validation options are used in the UI, i.e. search the place where the other validation (errors) are depicted and add the new ones there too.

1nf0rmagician (commented):
Two additional remarks
1. Please update the documentation. Fell free to not only add your new additions but also correct any issues you encountered when using the documentation to understand the Entry concept. As I told you at the beginning, it would be nice if the next person getting into it could understand what it is and what it can be used for without to much additional help from us personally.
2. Did you decide against the new `Base64String` validation attribute for a specific reason? Coincidently, we could probably make good use of that validation attribute with the new File System we want to provide😬

1nf0rmagician:
One more comment to add: Please check [this site](https://weblogs.asp.net/ricardoperes/net-8-data-annotations-validation), it explains the two different attributes to validate collection length and string length. I think it would be alright to share the field in the validation class, as we also have the `IsCollection` flag on the entry. Bt you need to make sure in the serialization (and on the consuming site as well) to ignore the wrong attributes and to check for the right things when validating

src/Moryx/Serialization/DefaultSerialization.cs
marierin: Yes, I actually did edit this method as well, but I have no idea where the changes are now... so I'll do it again.

src/Moryx.CommandCenter.Web/src/modules/models/EntryValidation.ts
marierin: I think that the validation is not used in the command center.

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
...
-- Key { Identifier = "Blah", Name = "Bob" }
-- Value { Current = "Foo" }
-- SubEntries
--- Entry-A
---- Key { Identifier = "ValueA", Name = "ValueA" }
---- Value { Current = "10" }
--- Entry-B
---- Key { Identifier = "ValueB", Name = "ValueB" }
---- Value { Current = "SomeString", Default = "Hello" }
+The PackagingResource object as a JSON-Script:
```
1nf0rmagician, line 80: 'as a serialized entry tree'

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
...
+                "type": "int",
+                "unitType": null,
+                "current": "",
+                "default": null,
+                "possible": null,
+                "isReadOnly": false
+            },
+            "validation": {
+                "isRequired": true,
+                "minLength": 7,
```
1nf0rmagician, line 112: I think you changed this to minimum and maximum didn't you? 😊

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
...
+},
+{
+	"identifier": "MaterialType",
+	"displayName": "Marterial",
+	"description": null,
+	"value": {
+		"type": "String",
+		"unitType": null,
+		"current": null,
+		"default": "abc",
```
1nf0rmagician, line 150: This does not match the example above 🙈

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
 Example class:
 
 ````cs
-public class Root
+public class PackagingResource
```
1nf0rmagician, line 57: I like the new example a lot! Could you please just change the property names to the c#conventions, i.e. PascalCase.
You could also remove the DataMember attributes as they are relevant only for storing the properties to the data base which is not necessarily part of these docs 😊

src/Moryx/Configuration/PossibleValuesSerialization.cs
```diff
@@ -65,21 +67,59 @@ public override EntryPrototype[] Prototypes(Type memberType, ICustomAttributePro
...
+            if (validationAttribute != null)
+            {
+                object[] allowed = validationAttribute.Values;
+                validation.AllowedValues = Array.ConvertAll(allowed, item => item.ToString());
+            }
+
+            IEnumerable<string> values;
+            // Possible values for primitive collections only apply to members
+            if (valuesAttribute == null || IsPrimitiveCollection(memberType))
+            {
```
1nf0rmagician, line 89: You could simplify you changes to this method as follows:
```c# 
var validationAttribute = attributeProvider.GetCustomAttribute<AllowedValuesAttribute>();
var allowedValues = validationAttribute?.Values.Select(v => v.ToString()) ?? Enumerable.Empty<string>() ;
return allowedValues.Concat(base.PossibleValues(memberType, attributeProvider));
```
I think that would be all you needed. Please check, however, whether these lines would be better placed in the DefaultSerialization class, as the PossibleValues method there is called from here. I think it should be better placed there 😊
Of course you could then move the precompiler condition to the other class and leave this method fully unchanged

1nf0rmagician (changes requested):

src/Moryx/Serialization/DefaultSerialization.cs
```diff
@@ -103,30 +111,38 @@ public virtual EntryValidation CreateValidation(Type memberType, ICustomAttribut
...
-                }
-                else if (attribute is RangeAttribute rangeAttribute)
+                switch (attribute)
                 {
-                    validation.Minimum = Convert.ToDouble(rangeAttribute.Minimum);
-                    validation.Maximum = Convert.ToDouble(rangeAttribute.Maximum);
-                }
-                else if (attribute is RegularExpressionAttribute regexAttribute)
-                    validation.Regex = regexAttribute.Pattern;
-                else if (attribute is StringLengthAttribute strLength)
```
1nf0rmagician, line 121: Am I missing something or are the `RegularExpressionAttribute` and the `StringLengthAttribute` missing? 🤔

src/Moryx/Serialization/DefaultSerialization.cs
```diff
@@ -73,6 +74,13 @@ public virtual EntryPrototype[] Prototypes(Type memberType, ICustomAttributeProv
         /// <see cref="ICustomSerialization"/>
         public virtual string[] PossibleValues(Type memberType, ICustomAttributeProvider attributeProvider)
         {
+#if NET8_0
+            var validationAttribute = attributeProvider.GetCustomAttribute<AllowedValuesAttribute>();
+            var allowedValues = validationAttribute?.Values.Select(o => o.ToString()) ?? Enumerable.Empty<string>();
+            if (allowedValues.All(value => !string.IsNullOrEmpty(value)))
```
1nf0rmagician, line 80: What is the intention behind this if? From my understanding this would cause the allowed values to be used only if the empty string is not part of the AllowedValues, is that intended? 
Also this method will throw a NullRef exception, if someone in a .net 8 application uses the possible values attribute for a list of primitive types or an enum.
Also the PossibleValues attribute is ignored, if the AllowedValues attribute is used (This is the only thing i'm fine with, if you intended it)

This is what i would expect, but please check if i missunderstood something 🙈
```
if( allowedValues is null)
{
    return allowedValues?.Distinct().ToArray();
}
```

1nf0rmagician (changes requested):

1nf0rmagician:
Also please check the unit test, they should be resolved by the changes necessary. But you might want to check them if not

Directory.Build.targets
```diff
@@ -18,8 +18,8 @@
   <!-- Package versions for package references across all projects -->
   <ItemGroup>
     <!--3rd party dependencies-->
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-    <PackageReference Update="Moq" Version="4.20.69" />
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="17.*" />
+    <PackageReference Update="Moq" Version="4.*" />
```
dbeuchler, line 22: I would not use a generic version here. Make it explict.

Directory.Build.targets
```diff
@@ -18,8 +18,8 @@
   <!-- Package versions for package references across all projects -->
   <ItemGroup>
     <!--3rd party dependencies-->
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-    <PackageReference Update="Moq" Version="4.20.69" />
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="17.*" />
```
dbeuchler, line 21: Same here - i would use an explict version here. In the past there were a lot of changes between the SDKs, also in minors.

docs/articles/Core/Serialization/EntryFormat.md
dbeuchler, line 57: Root sounds better than this

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
...
 {
-    [DisplayName("Bob")]
-    public Foo Blah { get; set; }
+    [EntrySerialize]
+    [isRequiered]
```
dbeuchler, line 60: Isnt it `RequiredAttribute`?

docs/articles/Core/Serialization/EntryFormat.md
```diff
@@ -54,33 +54,109 @@ The recursive `SubEntries` structure represents properties of a `EntryValueType.
...
+    public Product Product_1 { get; set; }
```
dbeuchler, line 61: No underscores in property-names

src/Moryx.CommandCenter.Web/src/modules/models/EntryValidation.ts
```diff
@@ -1,12 +1,13 @@
...
+ * Licensed under the Apache License, Version 2.0
+*/
+
+export default class EntryValidation {
+    public minimum: number;
+    public maximum: number;
+    public regex: string;
+    public isRequired: boolean;
+    public isPassword: boolean;
+    public deniedValue: string[];
```
dbeuchler, line 12: Do you know why this is not plural here? `deniedValues`

PR #519, opened 2025-02-06
Using Path.Combine for creating backup-path in DatabaseController

### LINKED ISSUE #509: ModelSetups are not listed on linux-environment
**Describe the bug**
When using MORYX in linux environment (docker in case, base is alpine), the ModelSetups are not listed. My assumption is that the problem is the hardcoded directory-separator in [DatabaseController.cs](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Runtime.Endpoints/Databases/Endpoint/DatabaseController.cs)

Fix proposal: 
- Use `Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)` for the execution directory
- Append "Backups" by `Path.Combine`

Maybe it has also something todo with my used `SupportedFileRegex` of `MyFile_(?<si>\d+)_(?<ietfTag>.*).xlsx` but I could not explay why this might be a problem

dbeuchler:
This should fix #509


1nf0rmagician (commented):
Love it! Thank you 😊

1nf0rmagician (approved):

PR #529, opened 2025-03-11
Add Flags enum entry

jsonBackup:
- Added the Flags entry to the EntryUnitType enum to represent unit types with the Flags attribute.
- Updated the GetUnitTypeByAttributes method to check if the property is an enum and has the Flags attribute.


1nf0rmagician:
FYI: @dbeuchler, @Toxantron, @dacky179

PR #530, opened 2025-03-11
Add Flags enum entry

jsonBackup:


1nf0rmagician (changes requested):
As mentioned in the other PR, you only need this one toward dev. But please before you merge it bump the Version file to a new minor version so that we don't forget and accidently create a patch version 🙂

VERSION
```diff
@@ -1 +1 @@
-8.2.0
+8.3.0
```
1nf0rmagician, line 1: As we are already on 8.3.1 with the latest release it appears the file was outdated even before you changed it. The correct version is 8.4.0 🙂

1nf0rmagician (changes requested):

1nf0rmagician (approved):

PR #533, opened 2025-04-09
Cleanup/code guidelines

dbeuchler:
This PR collects several changes which cleans up the code. Some of the changes are breaking and documented in migration guide.


Toxantron:
That looks like a lot of good changes, you think you can resolve the conflicts, so we can merge this?

Issue #246, opened 2023-03-28
Missing feature to execute resource initializer from facade


dbeuchler:
**Is your feature request related to a problem? Please describe.**
We want to auto install our application and want to execute a resource initializer from outside the resource management module.
There is no possibility to execute an initializer on the ResourceManagement-Facade (compare to execute product importer).

**Describe the solution you'd like**
A new method `ExecuteInitializer(IResourceIntializer initializer);`

**Describe alternatives you've considered**
There is a workaround to get the `ResourceGraph` by reflection.

````cs
var resourceInitializer = new ClipXResourceInitializer();
var resourceGraph = ResourceManagement.GetType().GetProperty("ResourceGraph").GetValue(ResourceManagement, null) as IResourceGraph;
var resources = resourceInitializer.Execute(resourceGraph);
resources.ForEach(r => resourceGraph.Save(r));
````

**Additional context**

PR #536, opened 2025-04-16
Removed resource initialiers from ModuleConfig

dbeuchler:
The configured resource initializers are not used in the module config anymore. It makes no sense to configure them.

Ideas for next major: 
- Preconfigure initializer an execute by console
- More important: Execute resource initializer from facade #246


1nf0rmagician (approved):
Neat 👍

PR #538, opened 2025-05-12
Add a shared file system to the kernel

Toxantron:


src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs
```diff
@@ -87,6 +93,19 @@ public static IConfigManager UseMoryxConfigurations(this IServiceProvider servic
             return configManager;
         }
 
+        /// <summary>
+        /// Use moryx file system and configure base directory
+        /// </summary>
+        /// <returns></returns>
+        public static IMoryxFileSystem UseMoryxFileSystem(this IServiceProvider serviceProvider, string path)
```
1nf0rmagician, line 100: Should this have a default path which points to a directory next to, not within the apps directory?

src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs
```diff
@@ -87,6 +93,19 @@ public static IConfigManager UseMoryxConfigurations(this IServiceProvider servic
...
+        {
+            var fileSystem = serviceProvider.GetRequiredService<LocalFileSystem>();
```
1nf0rmagician, line 102: While the method states MORYX file system we actually use the LocalFileSystem. I don't know really what would be better here, just wanted to say it feels of 😅

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+        private readonly ILogger _logger;
+
+        private readonly List<MoryxFileTree> _ownerTrees = new List<MoryxFileTree>();
+
+        public LocalFileSystem(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger(nameof(LocalFileSystem));
+        }
+
+        public void SetBasePath(string basePath = "fs")
```
1nf0rmagician, line 28: Isn't this somehow duplicated in the service collection extension `UseMoryxFileSystem`? Also it only allows for directories within the apps directory which is not recommended by .net 🙈

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+        {
+            _fsDirectory = Path.Combine(Directory.GetCurrentDirectory(), basePath);
+            _ownerFilesDirectory = Path.Combine(_fsDirectory, "owners");
+        }
+
+        public void LoadTrees()
```
1nf0rmagician, line 34: Who calls this? Should the FileSystem maybe be an `IHostedService` to have a start lifecycle hook which loads the trees? 🤔

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices.ComTypes;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Moryx.Runtime.Kernel.FileSystem
+{
+    internal class LocalFileSystem : IMoryxFileSystem
```
1nf0rmagician, line 15: Would it be a benefit from your perspective to propperly wrap all file system interactions in propper try-catch blocks, logging errors with clear indicators what went wrong and maybe what to do to solve it? 🤔

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+            }
+            return existingTree;
+        }
+
+        private MoryxFileTree ReadExtensibleTree(MoryxFileTree tree)
+        {
+            // Read tree from hash
+            var path = HashPath.FromHash(tree.Hash).FilePath(_fsDirectory);
+            if (!File.Exists(path))
+                throw new FileNotFoundException(path);
```
1nf0rmagician, line 162: since this is also called in the `LoadTrees` method I would expect that the error will be confusing for people using it. I imagine someone booting up the application and for some reason an owner file is missing. I think that could be better explained than this 🙈

1nf0rmagician: Doesn't the ReadAllFiles method also throw the exception, when called on a non existing file? 😅 i.e. is there any benefit in this?

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+            var ownerFiles = Directory.EnumerateFiles(_ownerFilesDirectory);
+            foreach (var ownerFile in ownerFiles)
+            {
+                var treeHash = File.ReadAllText(ownerFile);
+                var ownerTree = new MoryxFileTree
+                {
+                    FileName = ownerFile,
+                    Hash = treeHash,
+                };
+                ReadExtensibleTree(ownerTree);
```
1nf0rmagician, line 46: What is the goal in reading all trees ones? we don't store them when it's successful and there is also no specific error handling if we can't read them, is there? 🤔

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        public MoryxFile GetFile(string hash)
+        {
+            return _ownerTrees.FindFile(hash);
+        }
+
+        public async Task<string> WriteAsync(MoryxFile file, Stream content)
+        {
+            if (file is MoryxFileTree fileTree)
```
1nf0rmagician, line 57: Shouldn't we throw ArgumentNullExceptions early, if e.g the file is null?

1nf0rmagician: Same goes for other method implementing the interface, `OpenStream` would run in a `NullRefException` for example🙈

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
```diff
@@ -0,0 +1,278 @@
...
+            {
+                foreach (var line in tree.Files.Select(FileToLine))
+                    sw.WriteLine(line);
+                await sw.FlushAsync();
+            }
+
+            tree.Hash = await StreamToDiskAsync(stream);
+
+            // Update parent or owner file
+            if (tree.ParentTree == null)
```
1nf0rmagician, line 79: Did we conclude the discussion what happens if within a single owner tree multiple files are added/trees updates? 🤔

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
Toxantron: Well the extension calls the method. Technically I might work with currentDir/../file-system.

However we could also use an absolute path.

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
Toxantron: Good point

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
Toxantron: We do store them in the owner trees and thereby everyhing below is stored recursively. Seemed easier during runtime to maintain the trees in memory.

1nf0rmagician: Ahh, I see now 👍

src/Moryx.Runtime.Kernel/FileSystem/LocalFileSystem.cs
1nf0rmagician: Ohh sorry, I apparently wasn't fully awake yet 🙈 

I think absolute or `currentDir/../file-system` could work. I guess checking the user input would be a good thing and then either taking the absulute patch given or creating the directory with the relative path 🤔

PR #541, opened 2025-05-21
Another attempt to build RC packages

Toxantron:


1nf0rmagician:
@Toxantron is this still something that should be used for rc builds in the future or did you find another way which you did not pursue yet and where we could/should take over? 🤔

PR #542, opened 2025-06-13
Draft: add gRPC extensions an configuration class

MathoMathiasCamara:
Given that we are using gRPC in more than one project/repos. @seveneleven proposed to have a package for the common components to avoid duplication in future project.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -49,6 +49,8 @@
 
     <PackageReference Update="Castle.Windsor" Version="6.0.0" />
 
+	<PackageReference Update="Grpc.Core" Version="2.46.6" />
+	<PackageReference Update="Grpc.Net.Client" Version="2.71.0" />
   </ItemGroup>
 
 </Project>
--- a/src/Moryx/Grpc/ClientConfig.cs
+++ b/src/Moryx/Grpc/ClientConfig.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) 2025, Phoenix Contact GmbH & Co. KG
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.ComponentModel;
+
+namespace Moryx.Grpc;
+
+/// <summary>
+/// Configuration for the gRPC client.
+/// </summary>
+public class ClientConfig
+{
+    /// <summary>
+    /// Host of the gRPC server
+    /// </summary>
+    /// <example>localhost</example>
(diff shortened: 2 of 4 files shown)


src/Moryx/Grpc/ClientConfig.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </summary>
+    /// <example>50021</example>
+    [Description("Port of the gRPC server"), DefaultValue(50021)]
+    public int Port { get; set; }
+
+
+    /// <summary>
+    /// Whether to use TLS or not
+    /// </summary>
+    [Description("Wheter to use TLS or not"), DefaultValue(true)]
```
dbeuchler, line 33: `DefaultValue` = `true` is not a good idea because you never cannot set it to false

src/Moryx/Moryx.csproj
```diff
@@ -15,7 +15,10 @@
 
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" />
-    <PackageReference Include="Microsoft.Extensions.Logging"/>
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+	  
+	<PackageReference Include="Grpc.Net.Client" />
+	<PackageReference Include="Grpc.Core" />
```
dbeuchler, line 21: MORYX does not have any package references except dotnet and Newtonsoft (should be replaced by System.Text.Json later). If this should really part of the MORYX-Framework, this should be part of a separate package to keep the dependencies as minimal as possible

Directory.Build.targets
```diff
@@ -49,6 +49,8 @@
 
     <PackageReference Update="Castle.Windsor" Version="6.0.0" />
 
+	<PackageReference Update="Grpc.Core" Version="2.46.6" />
+	<PackageReference Update="Grpc.Net.Client" Version="2.71.0" />
```
dbeuchler, line 53: Is 2.46/2.71 really necessary, or is 2.0.0 also enough? Let the developers decide which version is necessary.

dbeuchler (commented):
I'm questioning whether it is truly necessary for MORYX to provide these kinds of wrapper functions. We had similar functionality in the past but removed most of it, primarily because wrapping tends to obscure important details and limits the ability to apply custom adjustments. This can lead to reduced transparency and flexibility in implementation.

PR #550, opened 2025-07-17
Make TimerId of ScheduleExecution thread safe.

dacky179:
Handle problem that two threads getting the same timer id.
Make sure a non periodical execution is stopped even also on exceptions.
image

### DIFF
--- a/src/Moryx/Threading/ParallelOperations.cs
+++ b/src/Moryx/Threading/ParallelOperations.cs
@@ -128,11 +128,8 @@ namespace Moryx.Threading
         /// </summary>
         public int ScheduleExecution<T>(Action<T> operation, T userState, int delayMs, int periodMs, bool criticalOperation) where T : class
         {
-            int id;
-            lock (_runningTimers)
-            {
-                id = ++_lastTimerId;
-            }
+            // thread safe increment of timer id
+            int id = Interlocked.Increment(ref _lastTimerId);
             
             var timer = new Timer(new NonStackingTimerCallback(state =>
             {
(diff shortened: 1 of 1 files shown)


src/Moryx/Threading/ParallelOperations.cs
```diff
@@ -128,19 +128,27 @@ public int ScheduleExecution<T>(Action<T> operation, T userState, int delayMs, i
         /// </summary>
         public int ScheduleExecution<T>(Action<T> operation, T userState, int delayMs, int periodMs, bool criticalOperation) where T : class
         {
-            var id = ++_lastTimerId;
+            int id;
+            lock (_runningTimers)
```
seveneleven, line 132: Wouldn't you want to create a separate lock object here for synchronizing the `_lastTimerId`?

dacky179: To avoid deadlock situation and keep it simple I did not create a second object for that single usage.

seveneleven (dismissed):

Toxantron:
FYI: Best practise for these cases is to use Interlocked.Increment(_timerId). Instead of blocking all threads coming through (even though irrelevant here), it turns the increment into an atomic thread safe iteration. Take a look at the [ParallelOperationsQueue](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx/Threading/ParallelOperationsQueue.cs)

dacky179:
One more time you teached me 👍  . Thanks.

seveneleven (approved):

PR #551, opened 2025-07-23
Include users management

jsonBackup:


seveneleven (approved):

Toxantron:
Sorry, I fixed the shift management via force push. @1nf0rmagician and I spoke about dropping the old UserManagement for MORYX 10, as it was basically replaced by operators.

PR #564, opened 2025-08-04
Fix/port health state of notification source facade

dacky179:
Port [#](https://github.com/PHOENIXCONTACT/MORYX-AbstractionLayer/pull/227)


seveneleven (approved):
Nice!

PR #566, opened 2025-08-07
Fix pagination for notifications in `Moryx.CommandCenter.Web`

DosenSuppe:
Notifications are no longer infinitely added below one another. 
Added Material-UI `TablePagination` to the table in `Notifications.tsx`. 
Changing `rowsPerPage` only updates the current page if it exceeds the max page number.

image
image
Screenshot 2025-08-07 152355

### DIFF
--- a/src/Moryx.CommandCenter.Web/src/modules/container/Notifications.tsx
+++ b/src/Moryx.CommandCenter.Web/src/modules/container/Notifications.tsx
@@ -7,7 +7,9 @@ import Table from "@mui/material/Table";
 import TableBody from "@mui/material/TableBody";
 import TableCell from "@mui/material/TableCell";
 import TableContainer from "@mui/material/TableContainer";
+import TableFooter from "@mui/material/TableFooter";
 import TableHead from "@mui/material/TableHead";
+import TablePagination from "@mui/material/TablePagination";
 import TableRow from "@mui/material/TableRow";
 import * as React from "react";
 import NotificationModel from "../models/NotificationModel";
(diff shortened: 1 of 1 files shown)


src/Moryx.CommandCenter.Web/src/modules/container/Notifications.tsx
```diff
@@ -40,19 +75,27 @@ export class Notifications extends React.Component<NotificationsPropModel> {
...
                         }
                     </TableBody>
+                    <TableFooter>
+                        <TablePagination
+                            count={messages.length}
+                            page={page}
+                            onPageChange={handlePageChange}
+                            rowsPerPage={rowsPerPage}
+                            onRowsPerPageChange={handleChangeRowsPerPage}
+                            rowsPerPageOptions={[5, 10, 25]}
```
seveneleven, line 94: Where did you get these values from? I'd vote for something like 25, 50, 100, as 5 and 10 seem to be quite low.

src/Moryx.CommandCenter.Web/src/modules/container/Notifications.tsx
```diff
@@ -17,15 +19,48 @@ interface NotificationsPropModel {
...
+}
+
+export class Notifications extends React.Component<NotificationsPropModel, NotificationPaginationState> {
 
     constructor(props: NotificationsPropModel) {
         super(props);
-
+        this.state = {
+            page: 0,
+            rowsPerPage: 5
```
seveneleven, line 33: You could introduce constants for all the numeric values you used.

seveneleven (changes requested):
Great work! It would be great if you had some screenshots or similar to show what it looks like.

PR #574, opened 2025-08-14
Fix configuring multiple plugins via the Command Center

DosenSuppe:
The format of the identifiers for created entry prototypes has changed:
	FROM: 	"CREATED"
	TO:		"CREATED{`CreatedPrototypesCount`}"

The Back-End has already been adjusted to support this format: https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/048b037


seveneleven (approved):
Good job 👍

PR #584, opened 2025-08-25
Fix ConnectionString forwarded improperly

DosenSuppe:
Problem:
ConnectionString not forwarded properly whilst configuring.

Solution:
Save button is disabled until the database connection test is completed.

Also added missing outline to the save button.

### DIFF
--- a/src/Moryx.CommandCenter.Web/src/databases/container/DatabaseModel.tsx
+++ b/src/Moryx.CommandCenter.Web/src/databases/container/DatabaseModel.tsx
@@ -48,6 +48,7 @@ interface DatabaseModelStateModel {
     selectedBackup: string;
     testConnectionPending: boolean;
     testConnectionResult: TestConnectionResult;
+    busy: boolean;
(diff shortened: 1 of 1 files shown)


1nf0rmagician (changes requested):
While your changes are reasonable, they unfortunately do not tackle the actual issue. 

Please reread the issue your are referencing and check the provided code reference as to were the improper/not-usage of the connection string for the database test occurs (which is in the .net code of the command center). The issue is specifically that the connection string is not used at all for the connection test but a default user instead. Disabling the save button will not help here, instead it would actually limit the users options to still modify the config even though we don't detect that the database is available.

Moreover, @MathoMathiasCamara whats your opinion on just disabling the save button while the connection test is running? I would have thought a small loading icon where we later show the connected/disconnected icons would be more user friendly, maybe with a hover text explaining that the connection to the db is currently tested.

1nf0rmagician (approved):

PR #597, opened 2025-09-10
FIX: Allow numbers in TextBindingResolver

slimberg91:
In cases where propertynames contain a number, e.g. "Comment1", they will be excluded from the resolver. This should fix the problem.

image

### DIFF
--- a/src/Moryx/Bindings/TextBindingResolverFactory.cs
+++ b/src/Moryx/Bindings/TextBindingResolverFactory.cs
@@ -14,7 +14,7 @@ namespace Moryx.Bindings
         /// Regex for binding.
         /// It matches strings of type "Product.Id" or "Recipe".
         /// </summary>
-        private const string BindingRegex = @"\{(?<binding>[A-Z][a-z][0-9]+(\.?[A-Z]\w+(?:\[\w+\])?)*)(?<hasFormat>:(?<format>\w+))?\}";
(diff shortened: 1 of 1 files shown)


src/Moryx/Bindings/TextBindingResolverFactory.cs
```diff
@@ -14,7 +14,7 @@ public static class TextBindingResolverFactory
         /// Regex for binding.
         /// It matches strings of type "Product.Id" or "Recipe".
         /// </summary>
-        private const string BindingRegex = @"\{(?<binding>[A-Z][a-z]+(\.?[A-Z]\w+(?:\[\w+\])?)*)(?<hasFormat>:(?<format>\w+))?\}";
+        private const string BindingRegex = @"\{(?<binding>[A-Z][a-z][0-9]+(\.?[A-Z]\w+(?:\[\w+\])?)*)(?<hasFormat>:(?<format>\w+))?\}";
```
1nf0rmagician, line 17: I think this is what you were aiming for
```
@"\{(?<binding>([A-Za-z_][A-Za-z0-9_]*)(\[\d+\])?(\.([A-Za-z_][A-Za-z0-9_]*)(\[\d+\])?)*)?(?<hasFormat>:(?<format>\w+))?\}"
```
Your suggestion would only match anything starting with capital letter, followed by lower letter and a set of numbers, e.g. `Co1`. I took the freedom to build upon your suggestion and allow for any valid C# Property names, followed by the brackets and formating already allowed before 😊

1nf0rmagician (changes requested):

PR #616, opened 2025-09-17
Add support for stateBaseType

MathoMathiasCamara:
The new feature adds support for Type, since the generic version is doing practically the same thing. Now on top of the generic type the `StateBase.Create` method overload support a Type as a parameter.

### DIFF
--- a/src/Moryx/StateMachines/StateBase.cs
+++ b/src/Moryx/StateMachines/StateBase.cs
@@ -99,11 +99,21 @@ namespace Moryx.StateMachines
             where TStateBase : StateBase
         {
             var stateBaseType = typeof(TStateBase);
+            Create(stateBaseType, context, initialKey);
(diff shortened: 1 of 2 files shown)


src/Moryx/StateMachines/StateBase.cs
```diff
@@ -99,11 +99,21 @@ internal static void Create<TStateBase>(IStateContext context, int? initialKey)
             where TStateBase : StateBase
```
1nf0rmagician, line 99: Correct me if I missunderstand something here, but this line makes already sure, that `StateBase` is assignable from `var stateBaseType = typeof(TStateBase);`, doesn't it? 🤔  
And I see no other changes in your PR aside from that, or am I overlooking something?

1nf0rmagician (approved):

PR #628, opened 2025-09-23
Simplify driver APIs

1nf0rmagician:


dbeuchler:
I have another one: 

 `IMarkiungLaserDriver`

````cs
        void RequestLastError();
        void RequestLastWarning();

        event EventHandler<NotificationResponse> ErrorOccured;
        event EventHandler<NotificationResponse> WarningOccured;
````

.. are always not implemented in my laser drivers (Foba, ACI) - it was just for an implementation of a specific manufacturer. Errors/Warnings should always be a response of the actions (Mark, SetMArkingFile) - other state changes should be raised by the driver-state-change. 

Are there other votes for removal?

dbeuchler:
closed after consent described here: #765

PR #640, opened 2025-09-24
StateClass now inherit the fields (constant State keys)  from the parent

MathoMathiasCamara:
This PR is a continuation of the #616 PR, previously when you inherited from an abstract state base class you lost all the constants protected fields  that define the different states `StateMap`. 

```c#
public abstract class MyGeneralStateBase : StateBase<MyContext> 
{
 [StateDefinition(typeof(StartState))]
 protected const int StateStart = 100;

 //constructor
  .....
}

public abstract class MySpecificStateBase :  MyGeneralStateBase 
 {
  // Here i need to redeclare it again
 [StateDefinition(typeof(StartState))]
 protected new const int StateStart = 100;

 [StateDefinition(typeof(StopState))]
 protected const int StateStop = 101;

 //constructor
  .....
}
``` 

Now all the constant fields of the parent are inherited and taking into account in the `StateMap` :

```c#
public abstract class MyStateBase : StateBase<MyContext> 
{
   [StateDefinition(typeof(StartState))]
   protected const int StateStart = 100;
}

public abstract MySpecificStateBase :  MyGeneralStateBase 
{
   [StateDefinition(typeof(StopState))]
   protected const int StateStop = 101;

   //constructor
    .....
}
```

### DIFF
--- a/src/Moryx/StateMachines/StateBase.cs
+++ b/src/Moryx/StateMachines/StateBase.cs
@@ -168,7 +168,7 @@ namespace Moryx.StateMachines
         /// </summary>
         internal static IEnumerable<FieldInfo> GetStateFields(Type stateBaseType)
         {
-            var stateFields = from field in stateBaseType.GetFields(BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public)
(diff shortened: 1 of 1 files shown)


1nf0rmagician:
@dbeuchler do you have an opinion on this?
@Toxantron I remember you saying at one point in the past you had made bad experience with inheriting state machines, but you did not mention specifics. If you are interested take a look (also at the referenced PR)

From my pov this change looks good 👍

dbeuchler (approved):

PR #663, opened 2025-10-02
Add split button to 'create dialog'

enis-pixel:
## Summary
This PR replaces the old `+`button with a split button that exposes two actions:
- Create (default): Behaves as `Add + Create` under the hood. It adds the current operation (if valid) and immediately creates all queued operations. If the form is invalid but there are already queued operations, it still creates those.
- Add: Only stages the current operation to the list (no create).

The right-hand arrow only selects which action the left button executes. Selection doesn't execute. This streamlines the common case (one order + one operation) while keeping the rare multi-operation flow via `Add`.

## Screenshots
image
image
image
image


src/Moryx.Orders.Web/src/app/dialogs/create-dialog/create-dialog.component.ts
```diff
@@ -237,4 +249,33 @@ export class CreateDialogComponent implements OnInit {
     }
     if (!failed) this.dialog.close();
   }
+
+  primaryActionLabel = computed(() => {
```
1nf0rmagician, line 253: fields belong at the beginning of the class to the others 😊

src/Moryx.Orders.Web/src/app/dialogs/create-dialog/create-dialog.component.ts
```diff
@@ -217,8 +225,12 @@ export class CreateDialogComponent implements OnInit {
         onlySelf: true,
       });
     }
-  }
 
+    this.operationNumberFormControl.reset('');
+    this.operationNumberFormControl.markAsPristine();
```
1nf0rmagician, line 230: this and the following line should be redundant, see https://angular.dev/api/forms/FormControl -> reset

Also check line 208, that shouldn't be necessary anymore either

src/Moryx.Orders.Web/src/app/dialogs/create-dialog/create-dialog.component.ts
```diff
@@ -237,4 +249,33 @@ export class CreateDialogComponent implements OnInit {
...
+
+  isSelected     = (a: Action) => this.primaryAction() === a;
+
+  private async performAction(a: Action): Promise<void> {
+    if (a === 'ADD_CREATE') {
+      if (this.canAddOperation()) this.addOperation();
+      if (this.operations().length > 0) await this.create();
+      return;
+    }
+    if (this.canAddOperation()) this.addOperation();
```
1nf0rmagician, line 279: either make this if-else and remove the return or use the early return and remove the second if, this is half of both options 😅

PR #669, opened 2025-10-05
Access instructor directly from endpoint

Toxantron:


Toxantron:
Still need to fix the tests.

dbeuchler:
Nice – definitely one of the most unnecessary modules we've ever had 👍 

Could we also harmonize the Namespaces `VisualInstructions` vs. `WorkerSupport`?

src/Moryx.ControlSystem.WorkerSupport/Components/InstructionProcessor.cs
```diff
@@ -1,56 +0,0 @@
...
-            // Apply content pattern
-            if (!string.IsNullOrEmpty(_config.ContentReplacement))
-            {
-                // Fill content with content regex
-                if (_contentRegex != null && !string.IsNullOrEmpty(item.Content))
-                    item.Content = _contentRegex.Replace(content, _config.ContentReplacement);
-                // If replacement was configured but no regex, we apply the preview regex
-                else if (_previewRegex != null && !string.IsNullOrEmpty(item.Preview))
-                    item.Content = _previewRegex.Replace(preview, _config.ContentReplacement);
-            }
```
dbeuchler, line 42: You have dropped the support for replacing regex - I don't really know, but was it an unused feature? I remember features like `Instruction.Content = "<guid>|Master"` which was then replaced by `http://server/media/...` - maybe we could move that to the `VisualInstuctor` itself?

dbeuchler (commented):
What do you think about bringing back a `WorkerSupport` as a resource which is a wrapper around all `VisualInsturctors` - this could also do the content-replacement

dacky179:
I though having a Workersupport module is a good thing.
We had ideas to show instructions from other modules than the resource management. If you remove the central module, would this still be possible? (without using a resource as gateway)

Toxantron:
I started harmonizing the namespaces and package names to VisualInstructions instead of AssemblyInstructions and WorkerSupport.

Toxantron:
> don't like it; Its a restriction that one display can only have one source for a instructor.
On one machine its ok to have a dedicated instructor for order instructions and the next one wants to show on same client like a resource did.

You misunderstood: Thats not my proposal, that was the current implementation of the worker support module. That's why it consider it useless, because the whole idea of VisualInstructions is to display and run sth on a specific screen.

> I would expect that a central module manages all clients and what they are showing.

Maybe. In it's current form it was merely a proxy created for business purpose and not technical reasons.

> In my opinion the WorkerSupport Module might be a very simple module, but it still has it's use cases.

It might be, right now I don't see any downside of putting the endpoint on the resources and removing it. MORYX does not loose any features and actually becomes a lot more resilient to instructor changes at runtime.

dbeuchler:
>MORYX does not loose any features and actually becomes a lot more resilient to instructor changes at runtime.

To finalize that, @1nf0rmagician @seveneleven and me discussed this PR again and the results are:

- We keep the WorkerSupport-Module for MORYX 10
- We harmonize the namespaces for WorkerSupport/VisualInstructions
- We keep the Content-Replacement Feature

I tagged the changes, for maybe later use, but this PR will not be merged.

PR #765, opened 2025-11-04
Rework of Driver APIs

dbeuchler:
Things for MORYX 10:

- [x] Replace Driver-callback mechanism by async-Tasks
- [x] Make StateClassification extensible for other drivers
- [x] Rework `IMarkingLaserDriver` and remove unused stuff (Warning/Error-events+requests)
- [x] IAxesController should accept arrays of instructions, for moving multiple axes at once

Points for discussion

- [x] Do we really need a driver for a weight scale? Couldn't this be represented as IInputDriver<double>
- [x] Do we need a scanner driver or could they be reduced to IInputDriver<string> or at least be combined with RfidAntenna?
- [x] Merge with #628 - won't do!


Toxantron:
@dbeuchler and @1nf0rmagician sorry for not being aus involved recently as I planned to. Could you eloborate why you keep the generic interfaces? Given the little benefit and the import regarding proxies, I don't fully understand your decision.

1nf0rmagician (approved):

PR #769, opened 2025-11-05
Removed analytics and replaced by external modules

dbeuchler:
- Analytics module was doing nothing and the web module will be replaced as described here #728
- Added support for external modules in shell - will be embedded with `<embed>`

image

````json
"Shell": {
  "ExternalModules": [
    {
      "Route": "example",
      "Title": "Example",
      "Url": "http://www.example.com",
      "Icon": "globe"
    }
  ]
}
````

### CODE REVIEW

File: src/Moryx.Launcher/Pages/External.cshtml.cs
Comment (1nf0rmagician) [inline src/Moryx.Launcher/Pages/External.cshtml.cs]: Can be removed

File: src/Moryx.Launcher/ShellNavigator.cs [hunk 1]
```diff
@@ -72,32 +73,51 @@ public async Task<IReadOnlyList<WebModuleItem>> GetWebModuleItems(HttpContext co
...
             }
 
-            var webModules = compiledPageActionDescriptors.Select(cpad => CreateWebModuleItem(cpad))
-                .OfType<WebModuleItem>().ToList();
+            // Load modules
+            var modules = compiledPageActionDescriptors.Select(CreateWebModuleItem)
+                .Where(m => m != null).ToList<ModuleItem>();
+
+            // Load external modules
+            var externalModuleConfigs = _configuration.GetSection("Shell:ExternalModules").Get<ExternalModuleConfig[]>();
```
1nf0rmagician: Do you want to split this into a separate method and reuse it in the cshtml of the page? Then we don't have the configuration string in two places

1nf0rmagician: And do we need to update the config during runtime, or should we get it once during startup and store it for the runtime? 🤔

dbeuchler: Adding to the appsettings and refresh the page is nice but I also affects the loading time of the whole application. Will move it to one-time load.

src/Moryx.Launcher/Pages/External.cshtml.cs
dbeuchler: I can do that. VS/Rider creates the file by default as "code-behind"

PR #770, opened 2025-11-05
Improve MQTT support

andreniggemann:
Bugfixes:

* The change in ConnectingToBrokerState prevents an application crash that were not uncommon during debugging

Cleanup:

* Mostly typos in the string resources or non equal punctuation
* Don't use the obsolete Payload Method and use ReadOnlySequence instead. To avoid unnecessary array copies methods deserializing the data have breaking signature changes
* Remove Newtonoft.Json in favor of System.Text.Json

Features:

* Add the option to add custom topics as the User by removing internal access modifiers from MqttTopic Serialize and Deserialize
* Add Retain information for Publishing and Receiving messages
* Support Mqtt5 response topics
* Support unsubscribing from Topics, by removing or changing the resource
* Support changing the broker without restarting the Resource Management, by a) providing a Reconnect method and b) handling changes to the relevant properties
* Support diagnostic tracing of message contents, before and after deserialization

### DIFF
--- a/src/Moryx.Drivers.Mqtt/MqttDriver.cs
+++ b/src/Moryx.Drivers.Mqtt/MqttDriver.cs
@@ -40,7 +40,7 @@ public class MqttDriver : Driver, IMessageDriver
     /// </summary>
     public IParallelOperations ParallelOperations { get; set; }
 
-    internal static readonly System.Diagnostics.ActivitySource _activitySource = new System.Diagnostics.ActivitySource("Moryx.Drivers.Mqtt.MqttDriver");
+    internal static readonly ActivitySource _activitySource = new ActivitySource("Moryx.Drivers.Mqtt.MqttDriver");
 
     #region EntrySerialize
 
@@ -68,10 +68,10 @@ public class MqttDriver : Driver, IMessageDriver
     /// </summary>
     [EntrySerialize, DataMember, DefaultValue("127.0.0.1")]
     [Display(Name = nameof(Strings.MqttDriver_BrokerUrl), Description = nameof(Strings.MqttDriver_BrokerUrl_Description), ResourceType = typeof(Strings))]
-    public string BrokerURL
+    public string BrokerUrl
     {
-        get => _brokerURL;
-        set => ConfigChange(ref _brokerURL, value);
+        get => _brokerUrl;
+        set => ConfigChange(ref _brokerUrl, value);
     }
 
     private void ConfigChange<T>(ref T field, T value) where T : IEquatable<T>
@@ -218,7 +218,7 @@ public class MqttDriver : Driver, IMessageDriver
     public bool HasChannels => Channels.Count > 0;
     internal DriverMqttState State => (DriverMqttState)CurrentState;
     private IMqttClient _mqttClient;
-    private string _brokerURL;
+    private string _brokerUrl;
     private string _identifier;
     private int _port;
     private int _scheduledReconnect;
@@ -320,7 +320,7 @@ public class MqttDriver : Driver, IMessageDriver
         _clientId = $"{System.Net.Dns.GetHostName()}-{Id}-{Name}";
         var optionsBuilder = new MqttClientOptionsBuilder()
             .WithClientId(_clientId)
-            .WithTcpServer(BrokerURL, Port)
+            .WithTcpServer(BrokerUrl, Port)
             .WithTlsOptions(new MqttClientTlsOptions() { UseTls = UseTls })
             .WithCleanSession(!ReconnectWithoutCleanSession);
 
@@ -655,8 +655,14 @@ public class MqttDriver : Driver, IMessageDriver
             var result = _mqttClient.UnsubscribeAsync(Identifier + subscribedTopic).Result;
         }
         // Ignore exceptions that can occure during shutdown
-        catch (ObjectDisposedException) { }
-        catch (MqttClientDisconnectedException) { }
+        catch (ObjectDisposedException)
+        {
+            // Ignore
+        }
+        catch (MqttClientDisconnectedException)
+        {
+            // Ignore
+        }
         catch (Exception ex)
         {
             Logger.LogError(ex, "Failed to unsubscribe from topic {topic}", subscribedTopic);
--- a/src/Moryx.Drivers.Mqtt/MqttTopic.cs
+++ b/src/Moryx.Drivers.Mqtt/MqttTopic.cs
@@ -17,6 +17,7 @@ using System.ComponentModel.DataAnnotations;
 using System.Diagnostics;
 using Moryx.Drivers.Mqtt.Messages;
 using System.Buffers;
+using System.Text.Json;
 
 namespace Moryx.Drivers.Mqtt
 {
@@ -442,7 +443,7 @@ namespace Moryx.Drivers.Mqtt
 
                     if (TraceDecodedMessage && span is not null)
                     {
-                        span.AddTag("message.decoded", System.Text.Json.JsonSerializer.Serialize(msg));
+                        span.AddTag("message.decoded", JsonSerializer.Serialize(msg));
                     }
                 }
             }
--- a/src/Moryx.Drivers.Mqtt/MqttTopics/MqttTopicJson.cs
+++ b/src/Moryx.Drivers.Mqtt/MqttTopics/MqttTopicJson.cs
@@ -8,6 +8,7 @@ using Moryx.Serialization;
 using Moryx.Drivers.Mqtt.Properties;
 using System.Text.Json;
 using System.Buffers;
+using System.Text.Json.Serialization;
 
 namespace Moryx.Drivers.Mqtt.MqttTopics
 {
@@ -51,7 +52,7 @@ namespace Moryx.Drivers.Mqtt.MqttTopics
         protected override byte[] Serialize(object payload)
         {
             var options = GetSystemTextJsonOptions();
-            return System.Text.Json.JsonSerializer.SerializeToUtf8Bytes(payload, options);
+            return JsonSerializer.SerializeToUtf8Bytes(payload, options);
         }
 
         /// <inheritdoc />
@@ -63,14 +64,14 @@ namespace Moryx.Drivers.Mqtt.MqttTopics
 
         private JsonSerializerOptions GetSystemTextJsonOptions()
         {
-            var options = new System.Text.Json.JsonSerializerOptions()
+            var options = new JsonSerializerOptions()
             {
-                PropertyNamingPolicy = Format == JsonFormat.camelCase ? JsonNamingPolicy.CamelCase : null,
+                PropertyNamingPolicy = Format == JsonFormat.CamelCase ? JsonNamingPolicy.CamelCase : null,
             };
             options.Converters.Clear();
             if (EnumsAsStrings)
             {
-                options.Converters.Add(new System.Text.Json.Serialization.JsonStringEnumConverter());
+                options.Converters.Add(new JsonStringEnumConverter());
             }
 
             return options;
--- a/src/Moryx.Drivers.Mqtt/MqttTopics/JsonFormat.cs
+++ b/src/Moryx.Drivers.Mqtt/MqttTopics/JsonFormat.cs
@@ -15,6 +15,6 @@ public enum JsonFormat
     /// <summary>
     /// Format object keys as camelCase
     /// </summary>
-    camelCase = 1
+    CamelCase = 1
 }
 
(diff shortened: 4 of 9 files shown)


src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -58,21 +59,46 @@ public class MqttDriver : Driver, IMessageDriver<object>
...
+            return;
+        }
+        field = value;
+        Reconnect();
+    }
+    [EntrySerialize]
+    [Description("Executes reconnect. Maybe helpful for troubleshooting")] // TODO: Localize
+    public void Reconnect()
+    {
+        if (State is null || State.Classification != StateClassification.Running) return;
```
dbeuchler: new line with braces

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -314,11 +371,20 @@ public void Send(object message)
...
+            // Search by identifier if set
+            if (message is IIdentifierMessage identifierMessage && !string.IsNullOrEmpty(identifierMessage.Identifier))
+                topics = Channels.Where(t => t.TopicType != TopicType.SubscribeOnly && t.Matches(identifierMessage.Identifier)).ToList();
+            else
+                topics = Channels.Where(t => t.TopicType != TopicType.SubscribeOnly && (t.MessageType?.IsInstanceOfType(message) ?? false)).ToList();
+
+        }
+        catch (Exception ex)
+        {
+            Logger?.LogError(ex, "Failed to find matching topic, because of an error");
```
dbeuchler: Logger should not be null

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -447,6 +554,25 @@ private static MqttQualityOfServiceLevel ConvertStringToQoS(string qos)
...
+            if(!_mqttClient.IsConnected)
+            {
+                return;
+            }
+            var result = _mqttClient.UnsubscribeAsync(Identifier + subscribedTopic).Result;
+        }
+        // Ignore exceptions that can occure during shutdown
+        catch (ObjectDisposedException) { } 
+        catch (MqttClientDisconnectedException) { }
+        catch (Exception ex)
```
dbeuchler: Format

dbeuchler (changes requested):
- Missing summaries in public members
- Several one-liners, bad readability of long lines. Separate them

src/Moryx.Drivers.Mqtt/MqttTopics/MqttTopicJson.cs
```diff
@@ -9,9 +9,17 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 using Moryx.Drivers.Mqtt.Localizations;
+using System;
+using System.Text.Json;
 
 namespace Moryx.Drivers.Mqtt.MqttTopics
 {
+
+    public enum SerializerSelection
```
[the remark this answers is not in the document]
dbeuchler: I used some time and have done some research - there are some reddit articles and other web-sites which say that Newtonsoft.Json should not be used anymore. System.Text.Json is faster, more flexible and the asp.net default. We should not introduce more code which must be migrated when removing the dependency in the future. I don't know how complex the migration will be to do it in MORYX10 or 12 - I will do some tests in the next days.

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -208,36 +271,91 @@ internal async Task Connect(bool firstConnect = true)
...
             if (firstConnect)
+            {
                 Logger.Log(LogLevel.Warning, "Error while connecting to Broker: {message}", e.Message);
+            }
+
             State.TriedConnecting(false);
         }
     }
 
+    /// <summary>
```
1nf0rmagician: would you put these properties next to the other props?

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -434,19 +620,58 @@ internal async Task SubscribeTopicAsync(string topic)
...
+        catch (MqttClientDisconnectedException) { }
+        catch (Exception ex)
+        {
+            Logger.LogError(ex, "Failed to unsubscribe from topic {topic}", subscribedTopic);
+        }
+    }
+
+    internal void DelayedConnectionAttempt()
+    {
+        ParallelOperations.ScheduleExecution(() => Connect(false).ContinueWith(t =>
```
1nf0rmagician: Do you want to remember the operation id here to cancel it in the OnStop of the driver?

src/Moryx.Drivers.Mqtt/MqttTopic.cs
```diff
@@ -348,52 +409,85 @@ internal override Task OnSend(object payload)
...
             topic = MqttDriver.Identifier + topic;
-            return MqttDriver.OnSend(new MqttMessageTopic(ResponseTopic, topic), msg);
+            return MqttDriver.OnSend(new MqttMessageTopic(ResponseTopic, topic, payload is IRetainAwareMessage ram ? ram.Retain ?? Retain : Retain), msg);
         }
 
-        internal override void OnReceived(string receivedTopic, byte[] messageAsBytes)
+        internal override void OnReceived(string receivedTopic, ArraySegment<byte> messageAsBytes, string? responseTopic, bool retain)
         {
-            if (MessageType != null)
+            TMessage msg;
```
1nf0rmagician: Is it intended that you have the Recieve Activity in the driver and an extra parsing here?

src/Moryx.Drivers.Mqtt/MqttTopics/MqttTopicJson.cs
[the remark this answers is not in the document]
1nf0rmagician: I would agree with @dbeuchler here, unless we have a specific demand to introduce this dependency for this package I would prefer keeping it out. You could make one commit removing it, then it would be an easy thing to bring the support in, if it becomes necessary for whatever reason 🙈

src/Moryx.Drivers.Mqtt/Properties/Strings.de.resx
```diff
@@ -117,7 +117,7 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="MqttDriver_BrokerUrl_Description" xml:space="preserve">
+  <data name="BROKER_URL_DESCRIPTION" xml:space="preserve">
```
dbeuchler: Someting is wrong here. Use `MqttDriver_BrokerUrl_Description`

src/Moryx.Drivers.Mqtt/MqttTopic.cs
```diff
@@ -119,8 +141,21 @@ public string Identifier
         [Display(Name = nameof(Strings.MqttTopic_ResponseTopic), Description = nameof(Strings.MqttTopic_ResponseTopic_Description), ResourceType = typeof(Strings))]
         public string ResponseTopic { get; set; }
 
+        /// <summary>
+        /// Marks if messages send by this topic should be marked as retained by default.
+        /// Messages can implment IRetainAware to overwrite the default.
+        /// </summary>
+        [EntrySerialize, DataMember, DefaultValue(false)]
```
dbeuchler: Please do not use default for bool. True cannot work and false is already the default.

src/Moryx.Drivers.Mqtt/MqttTopics/SerializerSelection.cs
```diff
@@ -0,0 +1,20 @@
...
+
+/// <summary>
+/// Selects which serializer library should be used for json serialization
+/// </summary>
+public enum SerializerSelection
+{
+    /// <summary>
+    /// Use newtonsoft.json for serialization
+    /// </summary>
+    NewtonsoftJson,
```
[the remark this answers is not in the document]
andreniggemann: I switched the defaults to point to System.Text.Json and marked Newtonsoft as obsolete.

I think @1nf0rmagician mentioned that it might be a good idea to keep it in this merge request and remove it later.
This would mean the logic is not lost due to squashing of the commits incase anyone might need it later.

Additionally I would like to add that we don't know if anyone relies on Newtonsoft features so that it might fall under the backward compatibility exemption of the ADR.

src/Moryx.Drivers.Mqtt/MqttTopic.cs
```diff
@@ -157,10 +196,21 @@ public TopicChanged(string oldTopic, string newTopic)
         /// </summary>
         public Type MessageType { get; set; }
 
+        /// <summary>
+        /// Output the type name of the resolved MessageType.
+        /// This allows the user to check if the Type resolution has succeeded.
+        /// </summary>
+        [EntrySerialize, ReadOnly(true)]
```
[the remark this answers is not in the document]
1nf0rmagician: Readonly actually requires a parameter if I'm not mistaken 😅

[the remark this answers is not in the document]
dbeuchler: We already investigated it. ReadOnly is not evaluated when the property has no public setter. Therefore it is redundant and must not be declared. @andreniggemann will remove the attribute.

src/Moryx.Drivers.Mqtt/MqttDriver.cs
andreniggemann: I know included this in a slightly modified form. 
I placed it in OnDisconnect which is called by OnStop. This has the benefit that it should also work for other situations where a disconnect is requested.

src/Moryx.Drivers.Mqtt/Messages/IRetainAwareMessage.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Interface for messages send via MQTT to mark that they care about being either Retain or not
+/// </summary>
+public interface IRetainAwareMessage
+{
+    /// <summary>
+    /// Marks if the mqtt message is retain. Null means using the default value of the topic
+    /// </summary>
+    public bool? Retain { get; set; }
```
dbeuchler: why nullable?

src/Moryx.Drivers.Mqtt/MqttTopics/MqttTopicJson.cs
```diff
@@ -36,43 +35,47 @@ public override string MessageName
...
             {
-                NullValueHandling = NullValueHandling.Ignore,
+                PropertyNamingPolicy = Format == JsonFormat.camelCase ? JsonNamingPolicy.CamelCase : null,
             };
-            if (Format == JsonFormat.camelCase)
-                settings.ContractResolver = new CamelCasePropertyNamesContractResolver();
+            options.Converters.Clear();
+            if (EnumsAsStrings)
+            {
+                options.Converters.Add(new System.Text.Json.Serialization.JsonStringEnumConverter());
```
dbeuchler: Often in this class: Why not importing namespace?

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -58,21 +61,54 @@ public class MqttDriver : Driver, IMessageDriver
...
-    public string Identifier { get; set; }
+    public string Identifier { get => _identifier; set => ConfigChange(ref _identifier, value); }
 
     /// <summary>
     /// URL or IP-Address of the MQTT Broker
     /// </summary>
     [EntrySerialize, DataMember, DefaultValue("127.0.0.1")]
     [Display(Name = nameof(Strings.MqttDriver_BrokerUrl), Description = nameof(Strings.MqttDriver_BrokerUrl_Description), ResourceType = typeof(Strings))]
-    public string BrokerUrl { get; set; }
+    public string BrokerURL
```
dbeuchler: Rename to BrokerUrl

src/Moryx.Drivers.Mqtt/MqttDriver.cs
```diff
@@ -406,19 +619,66 @@ internal async Task SubscribeTopicAsync(string topic)
...
+            if (!_mqttClient.IsConnected)
+            {
+                return;
+            }
+            var result = _mqttClient.UnsubscribeAsync(Identifier + subscribedTopic).Result;
+        }
+        // Ignore exceptions that can occure during shutdown
+        catch (ObjectDisposedException) { }
+        catch (MqttClientDisconnectedException) { }
+        catch (Exception ex)
```
dbeuchler: Format

dbeuchler (approved):

PR #779, opened 2025-11-07
Added ADR to use System.Text.Json instead of Newtonsoft.Json

### LINKED ISSUE #778: Add ADR for "Use System.Text.Json where ever it is possible"
I used some time and have done some research - there are some reddit articles and other web-sites which say that Newtonsoft.Json should not be used anymore. System.Text.Json is faster, more flexible and the asp.net default. We should not introduce more code which must be migrated when removing the dependency in the future. I don't know how complex the migration will be to do it in MORYX10 or 12 - I will do some tests in the next days.

_Originally posted by @dbeuchler in https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/770#discussion_r2495655870_

dbeuchler:
Closes #778

### DIFF
--- a/docs/articles/adr/0001-immutable-jobs.md
+++ b/docs/articles/adr/0001-immutable-jobs.md
@@ -1,10 +1,10 @@
-# ADR: Immutable Jobs
+# ADR-001: Immutable Jobs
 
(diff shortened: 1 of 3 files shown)


1nf0rmagician:
Love it! 🙏

PR #783, opened 2025-11-07
Added ADR for ResourceProxies

### LINKED ISSUE #756: Add Architecture Decision Record for Resource Proxy
The resource management is unique in the way that it makes use of the proxy pattern to provide its business objects to the rest of the application. Reasons for this decisions are however not generally known.

Add an Architecture Decision Record explaining why this decision was made to prevent reiterating that discussion again and again.

**Additional context**
There is a section in the resource management that already describes some of the motivation

> Since the architecture of Moryx includes an internal DI-Container for each module, it imposes a restriction on using components from inside the container outside of the Facade. Because of that, the resource management applies the proxy pattern to provide access to the resources API while simultaneously hiding the resource instance from the user. For more information on the structure of Moryx look into this article.
The proxy types implement the same interfaces as the resource type they are representing. The proxy forwards all calls to the Target and forwards events of the resource to listeners after replacing the sender object with itself. If methods or properties return a resource or collection of resources the proxy converts those on the ﬂy to proxies as well. When the resource management is shut down it calls Detach on the proxy to release the reference to the Target. To spare the developers the additional effort of creating a matching proxy class for each resource, the proxy classes are created on demand at runtime as classes derived from [ResourceProxy](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Resources.Management/Resources/ResourceProxy.cs). When another module resolves a resource instance over the Facade the [ResourceProxyBuilder](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Resources.Management/Resources/ResourceProxyBuilder.cs) determines all interfaces that the resource implements and creates a proxy type that offers the same interfaces. Next the proxy builder moves up the type tree looking for the least speciﬁc base type that implements the same number of interfaces. That way the resulting proxy type can be used for all derived types that only customize the existing behavior. In the ﬁgure above instances of ResourceA and ResourceB would use a different proxy type, but instances of ResourceC would be represented by ProxyB as well. Once the proxy type is created, it is instantiated for the initially requested target. The type and instance are stored within the [ResourceTypeController](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Resources.Management/Resources/ResourceTypeController.cs). All access to the same resource instance is handled by the same proxy to save memory and enable object reference comparison outside the resource management. The stored types are used to create proxies for instances of the same or a compatible resource type.
```

### PR DESCRIPTION
**Pull request by dbeuchler**
Based on #779 

Closes #756

### DIFF
--- a/docs/articles/adr/0003-resource-proxies.md
+++ b/docs/articles/adr/0003-resource-proxies.md
@@ -1,0 +1,42 @@
+# ADR-003: Export `ResourceProxy` out of Facade
+
+**Date:** 2025-11-07 \
+**Status:** Accepted \
+**Context:** Resources / AbstractionLayer 5 / MORYX 6+ Projects
+
+The `ResourceProxy` will be exported out of the module facade to enable interaction with resource APIs while maintaining the internal lifecycle and encapsulation principles defined by the MORYX architecture.
+
+In the MORYX architecture, each module contains its own internal dependency injection (DI) container. This design prevents direct use of internal module components outside the module’s facade. To provide controlled access to resource instances from outside the module, the **ResourceManagement** module applies the **Proxy Pattern**.
+
+This pattern allows external consumers to interact with resource interfaces without directly referencing or managing the actual resource instances, thereby preserving the module’s lifecycle integrity.
+
+## Motivation
+
+1. **Keep Module Lifecycle**
+   - Ensures that the module’s lifecycle remains self-contained and stable.
+   - When a module restarts or shuts down, all internal components are safely disposed.
+   - The proxy detaches from the target resource during shutdown (`Detach`), preventing stale references.
+
+2. **Automatic Proxy Generator**
+   - Reduces developer overhead by generating proxy classes dynamically at runtime.
+   - When a resource instance is resolved via the module’s facade, the system determines all interfaces implemented by the resource and creates a matching proxy type that exposes the same interfaces.
+   - This approach simplifies extensibility and avoids repetitive boilerplate proxy code.
+
+## Exceptions
+
+The ResourceManagement-facade will always create a proxy type for each resource type with the following exceptions:
+
+- Modification methods like `IResourceManagement.Create`, `IResourceManagement.Read`, `IResourceManagement.Modify`: They provide the original resource instance to have full access to it. The resource should only be used in the provided accessor-delegate and should never be exported from that.
+- Method `IResourceManagement.GetAllResources`: Used for endpoint-controllers to export all properties from the resource e.g. internal properties or methods marked with `EntrySerialize` which are not part of any interface.
+
+## Consequences
+
+- Developers can work with a resource proxy as if it were the original resource.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/articles/adr/0003-resource-proxies.md [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+   - The proxy detaches from the target resource during shutdown (`Detach`), preventing stale references.
+
+2. **Automatic Proxy Generator**
+   - Reduces developer overhead by generating proxy classes dynamically at runtime.
+   - When a resource instance is resolved via the module’s facade, the system determines all interfaces implemented by the resource and creates a matching proxy type that exposes the same interfaces.
+   - This approach simplifies extensibility and avoids repetitive boilerplate proxy code.
+
+## Exceptions
+
+There are **no exceptions** to this rule. The ResourceManagement-facade will always create a proxy type for each resource type.
```
1nf0rmagician: This is actually not true 🙈  
This whole section allows for direct interaction with resources https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/7bb2712ba5182aaa6c31a6e7e40d534c49a335f8/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs#L116  

And especially the `GetAllResources<TResource>(Func<TResource, bool> predicate)` method, [here](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/7bb2712ba5182aaa6c31a6e7e40d534c49a335f8/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs#L168C39-L168C98), was probably not a good idea, as it leads to a lot of confusion when the IDE suggest it next to the `GetResources<TResource>()` method, [here](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/7bb2712ba5182aaa6c31a6e7e40d534c49a335f8/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs#L93). One of which returns proxies the other the real objects.

I remember @Toxantron had to create these options for some reasons/features of the UI after we moved away from the internal WCF endpoints which were working on the real resources as well. We should have documented the decision back then...

Also, as the region suggests, these were part of a separate facade interface in v6 I think. If we had kept it separate, at least there would have been less confusion, as the feature would be mostly hidden.

1nf0rmagician (changes requested):

docs/articles/adr/0003-resource-proxies.md
dbeuchler: True, sorry, i  was not aware of that.

andreniggemann:
Thank you for documenting the rational behind resource proxies.
Do you think it would be a good idea, to add some of this info to the doc comments on IResourceManagement?
I am thinking of:
a. Note that all methods generally return proxies with a link to this ADR
b. explicitly stating on the methods that allow "raw" access, that they give direct access to the resource, combined with a warning that references to them should not be stored for later use?

1nf0rmagician (approved):

PR #794, opened 2025-11-13
Prepare EntryConvert to allow localized possible values

### LINKED ISSUE #400: EntryConvert does not support DisplayAttribute on enum members
Using `DisplayAttribute` on enum members does not have any effect.

````cs
public enum MyFoo
{
    [Display(Name = "One")]
    Foo1,
    [Display(Name = "Two")]
    Foo2
}

public class MyClass
{
    public MyFoo MyFoo { get; set; }
}
````

The possible values property is not using the value of the `DisplayAttribute`.

Support DisplayAttribute on enum members

### PR DESCRIPTION
**Pull request by jsonBackup**
Closes #400
Related https://github.com/PHOENIXCONTACT/ngx-moryx-web/pull/28

### CODE REVIEW

File: src/Moryx/Serialization/EntryConvert/EntryToModel/EntryCollectionBase.cs [hunk 1]
```diff
@@ -52,7 +52,7 @@ public string CurrentType
         /// <summary>
         /// Possible entries that can be used to create instances
         /// </summary>
-        public string[] PossibleTypes => CollectionRoot.Value.Possible;
+        public string[] PossibleTypes => CollectionRoot.Value.Possible?.Select(p => p?.Key).ToArray();
```
1nf0rmagician: Can you also make this an object array instead of string, classes could also use `DisplayName` and `Description 🤔`

src/Moryx/Serialization/EntryConvert/EntryConvert.cs
```diff
@@ -495,6 +495,26 @@ private static Entry ConvertParameter(ParameterInfo parameter, ICustomSerializat
             return parameterModel;
         }
 
+        /// <summary>
+        /// Convert string[] possible values to EntryPossible[].
+        /// Key and DisplayName are set to the string value.
+        /// </summary>
+        private static EntryPossible[] ConvertPossible(string[] possible)
```
1nf0rmagician: I think we should move the convertion one level up already, i.e. the API of the `serialization.PossibleValues(parameterType, parameter)` should already return the `EntryPossible`, don't you think? 🤔

dbeuchler: I also think so, `ICustomSerialization` should return `EntryPossible`

src/Moryx/Serialization/EntryConvert/EntryPossible.cs
```diff
@@ -0,0 +1,29 @@
+using System.Runtime.Serialization;
+
+namespace Moryx.Serialization
+{
+    /// <summary>
+    /// Possible value for an entry.
+    /// </summary>
+    [DataContract]
+    public class EntryPossible
```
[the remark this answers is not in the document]
dbeuchler: I also thought about that, but in this case `EntryPossible` is okay for me to be consistent with the other `Entry` related classes

src/Moryx/Configuration/PossibleValuesSerialization.cs
```diff
@@ -93,7 +93,7 @@ public override string[] PossibleElementValues(Type memberType, ICustomAttribute
 
             // Use attribute
             var values = valuesAttribute.GetValues(Container, ServiceProvider);
```
1nf0rmagician: Do we need to enhance the API of `GetValues` on the attribute or the Attribute itself as well?

dbeuchler (changes requested):
Looks good so far. What about 
- all UI updates, every usage of the entry editor should be broken?
- Migration guide information

jsonBackup:
> Looks good so far. What about
> 
>     * all UI updates, every usage of the entry editor should be broken?
> 
>     * Migration guide information

I've implemented the required changes in Moryx.CommandCenter.Web so the entry editor usages are now aligned with the new `EntryPossible` handling. Since https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/792 no other web projects reference EntryValue directly anymore, so they should not be broken by this but require the new version of ngx-moryx-web with the changes I've implemented here: https://github.com/PHOENIXCONTACT/ngx-moryx-web/pull/28.
Because of the ongoing supply-chain incident, I'm currently holding off on updating the package.json files to reference the new ngx-moryx-web version.

I've also extended the migration guide :)

dbeuchler (approved):

PR #810, opened 2025-11-20
Added ActivityConstraintContext to provide more information during constraint-check

### LINKED ISSUE #737: Constraint needs information about relevant Activity
originally from @dacky179 

Your lite resource facing problems during restart.
The system would redeliver already running activities the cell that causes problems.

A Constrain could check if a activitiy is already started by checking the tracing:

Image

Please change the constrain to get the activity, not the process.

dbeuchler:
@dacky179 fyi

closes #737


src/Moryx.AbstractionLayer/Processes/ProcessWorkplanContext.cs
```diff
@@ -9,12 +9,12 @@ namespace Moryx.AbstractionLayer.Processes
     /// <summary>
     /// Context for workplans that execute a process
     /// </summary>
-    public class ProcessContext : IWorkplanContext
```
1nf0rmagician: Could you mention this rename in the Migration Guide? 🙂

src/Moryx.AbstractionLayer/Activities/ActivityConstraintContext.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Moryx.AbstractionLayer.Constraints;
+using Moryx.AbstractionLayer.Processes;
+
+namespace Moryx.AbstractionLayer.Activities;
+
+/// <summary>
+/// Implementation of <see cref="IConstraintContext"/> used during process and activity handling
+/// </summary>
+public class ActivityConstraintContext : IConstraintContext
```
1nf0rmagician: Do we want to keep the interface, or would you count it as one of the interfaces that do not provide any benefits?

I mean it's empty and hides the one implementation that the ProcessEngine will always use.. 😅

PR #823, opened 2025-11-21
dependabot should not notifiy major for Microsoft-deps and group updates

dbeuchler:
- Groups merge requests for patch and minor changes
-  should not notifiy major for Microsoft-deps

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,19 @@ updates:
     directory: "/" # Location of package manifests
     target-branch: "future"
     open-pull-requests-limit: 10
+    ignore:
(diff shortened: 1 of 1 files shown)


1nf0rmagician (approved):

dbeuchler:
> Considering other dependencies like OPC UA and MQTT for example, I would prefer to keep the major PRs from dependabot. Not to easily merge the PRs but as an notification service that a new Major for a dependency has been released

Okay, good point - I have changed the ignore to limit microsoft packages by `Microsoft.*`

1nf0rmagician (approved):

PR #839, opened 2025-11-27
Removed EntryToModelConverter

dbeuchler:
This component was used in WPF UIs to map an entry to a view model and vice versa. This is not used anymore and brings no benefit to the platform. I removed it completely.


dbeuchler:
> Fascinating what things were forgotten and left to rot in the past 4 years 🙈

Thats what I said! I am really not done with my cleanup.

Issue #830, opened 2025-11-26
Reactivate Tests for future


dbeuchler:
The tests are not activated in future branch. This must be reactivated and all failing tests have to be fixed.

PR #891, opened 2025-12-02
Fix browsing nodes

seveneleven:
It's common, that OPC UA nodes appear in multiple positions of the tree. But
after they appeared once, they've been skipped for all other occurrences.

This is fixed now: Nodes will be listed, if they occur in different branches of
the tree. They will be ignored though, if they appear more than once on the
same branch. Also, if already found, they won't be updated by following
occurrences, assuming that nodes with the same ID will be the same entity.

### DIFF
--- a/src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
+++ b/src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0
 
 using System.Globalization;
-using System.Text.RegularExpressions;
 using Microsoft.Extensions.Logging;
 using Moryx.AbstractionLayer.Drivers;
 using Moryx.AbstractionLayer.Drivers.Message;
@@ -21,6 +20,7 @@ public class OpcUaNode : IMessageChannel
 {
     private IOpcUaDriver _driver => (IOpcUaDriver)Driver;
 
+    private const int DefaultNamespaceLength = 5;
     private readonly IModuleLogger _logger;
 
     /// <inheritdoc/>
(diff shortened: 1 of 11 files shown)


src/Tests/Moryx.Drivers.OpcUa.Tests/HandlingCyclicNodesTests.cs
```diff
@@ -95,6 +89,149 @@ public void DoNotShowCyclicNodesInTheUi()
...
+    /// Sets up a cyclic node hierarchy:
+    ///     ROOT
+    ///     ├─ NODE_1
+    ///        ├─ NODE_1
+    ///           ├─ NODE_1
+    ///           ...
+    /// </summary>
+    private void SetupDirectCyclicNodes()
+    {
+
```
dbeuchler: format

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -314,7 +313,7 @@ private async Task<bool> CreateSession(bool firstTry)
         {
             if (firstTry)
             {
-                Logger.Log(LogLevel.Error, "Failed to connect {Uri} ({Message})", builder?.Uri.ToString() ?? OpcUaServerUrl, e.Message);
+                Logger?.Log(LogLevel.Error, "Failed to connect {Uri} ({Message})", builder?.Uri.ToString() ?? OpcUaServerUrl, e.Message);
```
dbeuchler: OpcUaDriver is a container component and the logger is required. Null-Checks are not necessary!

src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
```diff
@@ -82,12 +81,34 @@ public event EventHandler<object> Received
...
+    /// <summary>
+    /// Converts a NodeId string into an expanded NodeId string that matches the OPC UA specification;
+    /// </summary>
     public static string CreateExpandedNodeId(string nodeId)
     {
-        var result = Regex.Replace(nodeId, ";ns=\\d+", "");
-        result = result.Trim();
+        var result = nodeId.Trim();
+        result = OpcUaNodeHelpers.StringWithUriAndNamespaceIndex().Replace(result, "");
+        result = RemoveNamespaceIndexZero(result);
```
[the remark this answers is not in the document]
seveneleven: I wanted to have the same level of abstraction here and keep low level details outside this function to improve consistency and readability.
This should also help to reason about what should happen here instead of just some conditional string manipulation that I would have needed to clutter with some comments otherwise.

seveneleven:
> @seveneleven Are all OpcUa-tests fixed with this PR? Maybe merge future into this branch. We weit for that in #830.

Yes, they are

dbeuchler:
> > @seveneleven Are all OpcUa-tests fixed with this PR? Maybe merge future into this branch. We weit for that in #830.
> 
> Yes, they are

nice!!

PR #892, opened 2025-12-02
Rework of TypeName usage in data models

### LINKED ISSUE #735: Use Assembly, ClassName and Namespace instead of TypeName in ProductEntity

dbeuchler:
We use the Type-fullname anywhere in MORYX and never the AssemblyFullQualifiedName. This is okay, because we should not get type conflicts. We only get type conflicts in MORYX if type is defined in different assemblies in the same namespace. If this happens, more stuff is broken in MORYX. Therfore: I removed the assembly from StepEntity and replaced by TypeName which is the full name. This is now consistent over all modules.

There was also a TypeName on ProcessEntity. This was not used in any line of code.

closes #735 and #734


src/Moryx.Products.Management/Model/Migrations/Sqlite/20251202123359_StepTypeName.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
1nf0rmagician: We should still add the license header in the auto-generated files, don't you think? 🤔

1nf0rmagician (dismissed):

1nf0rmagician (approved):

Issue #710, opened 2025-10-15
Allow optional integration of configuration providers in MORYX ConfigurationManager


1nf0rmagician:
**Is your feature request related to a problem? Please describe.**
As described in #672 it is an anti-pattern to put secrets into unencrypted config files within the applications working directory. 
> Never store passwords or other sensitive data in source code or configuration files. Production secrets shouldn't be used for development or test. Secrets shouldn't be deployed with the app. Production secrets should be accessed through a controlled means:
_[Source](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-9.0&tabs=windows)_

Currently, the means of getting configurations **and** secrets into MORYX components is through the `IConfigManager`. This component does not allow any special handling of secrets though.

**Describe the solution you'd like**
The .net way of managing secrets is by using [configuration providers](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration-providers). e.g. the .net [secret manager](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-9.0&tabs=windows#secret-manager) tool for development purposes or the [Azure Key Vault configuration provider](https://learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-9.0) for production purposes (among others).  
We should offer means to reference settings from configuration providers via the `IConfigManager` component. 

Acceptance Criteria:
- Properties in config classes are populated with values from config providers during runtime
  - Properties marked with a `PasswordAttribute` should search for these settings by default
  - Further options to indicate a property to be populated from a config provider _can_ be evaluated
- Any in the program.cs added config provider can be used as a data source
- _To be discussed:_ Referencing properties from a config provider is done just like from the code, e.g. `["Movies:ServiceApiKey"]` [(more information)](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-9.0&tabs=windows#read-the-secret-via-the-configuration-api)

**Describe alternatives you've considered**
#672 describes an alternative for securing app secrets based on an encryption mechanism for the whole MORYX configuration with a key stored as an environment variable.  
We propose this alternative solution for the following reasons:
- It is the way .net proposes to handle secrets
- It allows for very simple replacements of the secret providers between dev and prod environments
- It allows for a wide range of secret managers to be included and integrated with MORYX
[...]

PR #894, opened 2025-12-03
Allow optional integration of configuration providers

jsonBackup:
#710

### DIFF
--- a/src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
+++ b/src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
@@ -1,9 +1,13 @@
 // Copyright (c) 2025, Phoenix Contact GmbH & Co. KG
 // Licensed under the Apache License, Version 2.0
 
+using Microsoft.Extensions.Configuration;
 using Moryx.Configuration;
 using Moryx.Serialization;
 using Newtonsoft.Json;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+using System.Reflection;
 
 namespace Moryx.Runtime.Kernel
 {
@@ -14,6 +18,9 @@ namespace Moryx.Runtime.Kernel
     {
         private readonly ConfigLiveUpdater _liveUpdater = new();
         private readonly SharedConfigProvider _sharedProvider;
+
+        // NEW: hook into Microsoft.Extensions.Configuration
+        public IConfiguration? Configuration { get; set; }
 
         /// <summary>
         /// Extension of the config files used in the MORYX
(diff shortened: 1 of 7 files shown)


src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
```diff
@@ -139,6 +199,12 @@ protected virtual ConfigBase TryGetFromDirectory(Type confType, string name)
             else
             {
                 config = CreateConfig(confType, ConfigState.Generated, "Config file not found! Running on default values.");
+
+                // Defaults & shared configs
+                ValueProviderExecutor.Execute(config, new ValueProviderExecutorSettings().AddProviders(ValueProviders));
```
1nf0rmagician: What does this line do?

src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
```diff
@@ -28,6 +35,56 @@ public ConfigManager()
             _sharedProvider = new SharedConfigProvider(this);
         }
 
+        protected virtual void ApplyConfigurationProviders(object config)
```
1nf0rmagician: Can't you already limit this to `ConfigBase`?

src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
```diff
@@ -28,6 +35,56 @@ public ConfigManager()
...
+                }
+                else if (hasPasswordAttribute && !string.IsNullOrWhiteSpace(defaultSectionName))
+                {
+                    candidateKey = $"{defaultSectionName}:{prop.Name}";
+                }
+
+                if (string.IsNullOrWhiteSpace(candidateKey))
+                    continue;
+
+                var providerValue = Configuration[candidateKey!];
```
1nf0rmagician: What happens here if multiple providers are available?

1nf0rmagician (commented):
First draft looks promising to me 👍 The following point you could consider next:
- Add/Modify documentation accordingly
- Add a short tutorial on how to populate config values e.g. in the MQTT or OPC UA Driver config from environment variables 
- Add code or a seperate ticket for nested properties in objects in the config
- Clean up some of the code, test names, comments, etc. 

I think the limitation to string properties could only be overcome with the implicit config population and since you only consider password properties there they will need to be of type string to 😅 But I think you could split this up into a separate issue, if you already have an idea on how to solve this, keeping this change limited to string properties would be alright for me.

andreniggemann:
Feature merged via #1259

PR #897, opened 2025-12-03
Make sure `OpcUaDisplayNode.NodeId` returns a proper node id

seveneleven:
… in the form of `nsu=http://namespace.org/;i=1234`


src/Tests/Moryx.Drivers.OpcUa.Tests/OpcUaDisplayNodeTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    [Test]
+    public void DisplayName()
+    {
+        Assert.That(_displayNode.DisplayName, Is.EqualTo("NodeDisplayName"));
+    }
+
+    [Test]
+    public void ToStringCombinesIdAndDisplayName()
+    {
+        Assert.That(_displayNode.ToString(), Is.EqualTo("i=1234 - NodeDisplayName"));
```
dbeuchler: Split Act and Assert - in this case `ToString` is the Act

PR #833, opened 2025-11-26
Async life-cycle

dbeuchler:
## AsyncStateBase

We now provide a full async implementation of the `StateBase`. To keep consistent naming, the `StateBase` was splitted to `SyncStateBase` and `AsyncStateBase`.
Due to the reduction of unnecessary interfaces (#59), `IState` was removed and `StateBase` will be used in e.g. `IStateContext` now.

The `StateMachine` class was extended by `WithAsync()` and `ForceAsync()`. The `AsyncStateBase` provides async all the way: `NextStateAsync()`, `OnEnterAsync()`, `OnExitAsync`.\
**Upgrade hint:** Replace `StateBase<TContext>` by `SyncStateBase<TContext>`.

The same convention was applied to `DriverState`. It was renamed to `SyncDriverState` and a new implementation `AsyncDriverState` was added with full async support.\
**Upgrade hint:** Replace `DriverState<TContext>` by `SyncDriverState<TContext>`.

## Async ServerModuleBase life-cycle
As described in #24: More and more apis are refactored to async task based methods (e.g. loading entitites from database on startup). We introduced already `IAsyncInitializable` and `IAsyncPlugin` in 827ba8014ce5b9abedcd7404f3eb70b6a7802240. For server modules we have no async context in initialize, start and stop. Async methods must be synchronized to keep the starting and stopping behavior.

The lifecycle methods of ServerModules have been migrated from void to async Task to enable modern asynchronous programming with async/await. Since there was no native async context for lifecycle methods, asynchronous calls must be synchronized to preserve the existing start/stop behavior. The runtime now supports asynchronous initialization, startup, and shutdown processes (e.g., loading entities from a database during startup).

````cs
protected override Task OnInitializeAsync()
{
    return Task.CompletedTask;
}

protected override async Task OnStartAsync()
{
    var asyncFoo = Container.Resolve<IAsyncFoo>();
    await asyncFoo.StartAsync();
}

protected override async Task OnStopAsync()
{
    var asyncFoo = Container.Resolve<IAsyncFoo>();
    await asyncFoo.StopAsync();
}
````

## Async Resource life-cycle

Same as done in ServerModules has been done for Resources. The ResourceManagement now supports asynchronous initialization, startup, and shutdown processes.

````cs
protected override Task OnInitializeAsync()
{
    return base.OnStartAsync();
}

protected override Task OnStartAsync()
{
    return base.OnStartAsync();
}

protected override Task OnStopAsync()
{
    return base.OnStopAsync();
}
````

[...]

Issue #831, opened 2025-11-26
Release MORYX 10


dbeuchler:
- [ ] Test of future packages in Application successful
- [ ] Migration Guide is complete
- [ ] All issues of Milestone 10.0.0 are finalized
- [ ] ...

to be continued

PR #900, opened 2025-12-04
Support async Methods in EntryConvert

dbeuchler:
- Supports async invocation of methods now by `InvokeMethodAsync`. Synchronous methods are executed synchronously.
- The synchronous `InvokeMethod` does now support async methods too. They are executed synchronously.

After merge of #833 we can think about removing synchronous API, it is mostly used in async context (controllers).

### DIFF
--- a/src/Moryx/Serialization/EntryConvert/EntryConvert.cs
+++ b/src/Moryx/Serialization/EntryConvert/EntryConvert.cs
@@ -385,7 +385,7 @@ namespace Moryx.Serialization
         /// </summary>
         public static MethodEntry EncodeMethod(MethodBase method, ICustomSerialization serialization)
         {
-            return new MethodEntry
+            var methodEntry = new MethodEntry
             {
                 Name = method.Name,
                 IsConstructor = method.IsConstructor,
@@ -399,6 +399,14 @@ namespace Moryx.Serialization
                     SubEntries = method.GetParameters().Select(p => ConvertParameter(p, serialization)).ToList()
                 }
             };
+
+            // Constructors have no return type
+            if (method is MethodInfo methodInfo)
+            {
+                methodEntry.IsAsync = IsAsyncMethod(methodInfo);
+            }
+
+            return methodEntry;
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Moryx/Serialization/EntryConvert/EntryConvert.cs
```diff
@@ -749,22 +757,100 @@ public static Entry InvokeMethod(object target, MethodEntry methodEntry)
...
+        {
+            var (method, arguments) = GetMethodInfoAndArguments(target, methodEntry, customSerialization);
+            object invokeResult;
+
+            // Also part of MethodEntry but be safer than sorry
+            var isAsync = IsAsyncMethod(method);
+            if (isAsync)
+            {
+                var task = (Task)method.Invoke(target, arguments)!;
+                await task.ConfigureAwait(false);
```
1nf0rmagician: If I'm not mistaken, `.ConfigureAwait(false)` is not really relevant in .Net anymore, hence we don't use it anywhere.

src/Moryx/Serialization/EntryConvert/EntryConvert.cs
```diff
@@ -749,22 +757,100 @@ public static Entry InvokeMethod(object target, MethodEntry methodEntry)
...
+        }
+
+        private static (MethodInfo, object[]) GetMethodInfoAndArguments(object target, MethodEntry methodEntry, ICustomSerialization customSerialization)
         {
             var method = target.GetType().GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
                 .FirstOrDefault(m => m.Name == methodEntry.Name && (m.IsPublic || m.IsAssembly) && ParametersProvided(m.GetParameters(), methodEntry));
             if (method == null)
             {
-                throw new MissingMethodException();
+                throw new MissingMethodException($"Method {methodEntry.Name} not found on target!");
```
1nf0rmagician: Want to add the target type to the exception for an even better error experienca?

src/Moryx/Serialization/EntryConvert/MethodEntry.cs
```diff
@@ -56,7 +62,8 @@ public MethodEntry Clone(bool deep)
             {
                 Name = Name,
                 DisplayName = DisplayName,
-                Description = Description
+                Description = Description,
+                IsAsync =  IsAsync
```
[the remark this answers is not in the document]
dbeuchler: I know, but I dont mentioned it because it is not required to have this. If it is async or not is evaluated in EntryConvert based on the Method definition. I will add it to #831

1nf0rmagician (approved):

PR #913, opened 2025-12-09
Support of filter by ProductState and added Generated

dbeuchler:
- `ProductState` and `ProductInstanceState` are now Flags-Enum.
- Introduced `ProductState.Generated` to classify a product as generated for internal use
- Added support to query by required product state

@1nf0rmagician fixed line endings to be LF


src/Moryx.AbstractionLayer/Products/ProductInstanceState.cs
```diff
@@ -4,33 +4,35 @@
...
         /// The production on this instance was paused
         /// </summary>
-        Paused,
+        Paused = 1 << 2,
 
         /// <summary>
         /// The production process succeeded.
         /// </summary>
-        Success,
+        Success = 1 << 3,
```
1nf0rmagician: `Success` and `Failure` would require a database migration of some sort, wouldn't they?

dbeuchler: oh sh**, thanks! Added migration (the database column was wrong type too)

PR #920, opened 2025-12-11
Support of CancellationToken and async ConfigBasedComponentSelector

dbeuchler:
### CancellationToken support

- All public APIs which are Task base provide a cancellation token parameter to support cancellation of long-running operations.
  - Plugins of modules use an none optional `CancellationToken` parameter.
  - Facade methods use an optional `CancellationToken` parameter with default value.
  - Module internal plugins which are exposed to plugins like `IProductStorage`or `IResourceGraph` use an optional `CancellationToken` parameter with default value.
- If cancellation is not supported by the component, no `CancellationToken` parameter is provided.

regex to find occurrences: `\b(?:public|private|protected|internal|static|partial|\s)*\s*(?:async\s+)?(?:Task|ValueTask)(?:<[^<>]*(?:<(?:[^<>]|<(?:[^<>])*>)*>)*>)?\s+\w+\s*\(`

Also important for this PR: This is not implementing cancellation in each component but prepares the API for minor updates.

### ConfigBasedComponentSelector async support

The `ConfigBasedComponentSelector` was extended to support async initialization of the selected component. If the component implements `IAsyncConfiguredInitializable`, its `InitializeAsync` method will be called during selection. It is not required that the `Create` method returns a task, the `InitializeAsync` will run synchronously after creation. If it returns a Task, it will await its completion.

The following sample shows the new possibilities of the `ConfigBasedComponentSelector`:

````cs
public interface ISample : IAsyncConfiguredInitializable<SampleConfig>;

[PluginFactory(typeof(IConfigBasedComponentSelector))]
internal interface ISampleFactory
{
    // InitializeAsync will be called synchronously after creation using CancellationToken.None
    ISample Create(SampleConfig config);

    // InitializeAsync will be called synchronously after creation with passing cancellationToken
    ISample Create(SampleConfig config, CancellationToken cancellationToken);

    // InitializeAsync will be called asynchronously after creation using CancellationToken.None
    Task<ISample> Create(SampleConfig config);

    // InitializeAsync will be called asynchronously after creation with passing cancellationToken
    Task<ISample> Create(SampleConfig config, CancellationToken cancellationToken);
}
````

### DIFF
--- a/docs/migrations/v8_to_v10.md
+++ b/docs/migrations/v8_to_v10.md
@@ -417,6 +417,10 @@ The API of `IResourceInitializer` was adjusted
 
 - API `LoadType` is using `IIdentity` instead of `ProductIdentity` but in MORYX 10.0, only `ProductIdentity` is supported.
 
+## Modules-Orders
+
+- Removed **Amount Reached Notification** from OperationData. **Upgrade hint:** Replace by custom module, to add notifications for certain modules.
+
 ### Product importer
 
 - Introduced `ProductImporterAttribute` for harmonized registration of importers.
--- a/src/Moryx.CommandCenter.Web/Moryx.CommandCenter.Web.csproj
+++ b/src/Moryx.CommandCenter.Web/Moryx.CommandCenter.Web.csproj
@@ -2,7 +2,7 @@
 
   <Target Name="BuildUI" Condition="('$(Configuration)'!='Debug' Or !Exists('wwwroot\bundle.js')) And ('$(NoBuild)' != 'true') ">
    <Exec WorkingDirectory="./" Command="npm config set strict-ssl false" />
-   <Exec WorkingDirectory="./" Command="npm install" />
+   <Exec WorkingDirectory="./" Command="npm ci" />
    <Exec WorkingDirectory="./" Command="npm run build --force" />
    <Exec WorkingDirectory="./" Command="npm config set strict-ssl true" />
   </Target>
--- a/src/Moryx.CommandCenter.Web/package-lock.json
+++ b/src/Moryx.CommandCenter.Web/package-lock.json
@@ -1,0 +1,7642 @@
+{
+    "name": "commandcenter",
+    "version": "1.0.0",
+    "lockfileVersion": 3,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "commandcenter",
+            "version": "1.0.0",
+            "license": "Apache-2.0",
+            "dependencies": {
+                "@emotion/react": "^11.14.0",
+                "@emotion/styled": "^11.14.1",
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: src/Moryx.Products.Management/Components/IProductManager.cs [hunk 1]
```diff
@@ -100,7 +100,7 @@ internal interface IProductManager : IAsyncPlugin
         /// </summary>
         /// <param name="ids">The IDs of instances that should be loaded</param>
         /// <returns>The instance with the id when it exists.</returns>
-        Task<IReadOnlyList<ProductInstance>> GetInstances(params long[] ids);
+        Task<IReadOnlyList<ProductInstance>> GetInstances(long[] ids);
```
1nf0rmagician: Why did you change this? 🤔

src/Moryx.Shifts.Management/Model/API/IShiftAssignementRepository.cs
```diff
@@ -8,8 +8,6 @@ namespace Moryx.Shifts.Management.Model
     public interface IShiftAssignementRepository : IRepository<ShiftAssignementEntity>
     {
         ShiftAssignementEntity CreateFromContext(ShiftAssignementCreationContext context);
-
-        Task<ShiftAssignementEntity> CreateFromContextAsync(ShiftAssignementCreationContext context);
```
1nf0rmagician: Why did you remove this?

1nf0rmagician (approved):
Looks good, there are just some issues in the docs which could be fixed after the API Freeze and some questions for my understanding 👍

src/Moryx.Products.Management/Components/IProductManager.cs
dbeuchler: Because the public api was also with params and passing cancellationtoken with default in combination was irritating in the api usage. The api user can pass an array too

src/Moryx.Container/ComponentSelectors/ConfigBasedComponentSelector.cs
```diff
@@ -80,5 +94,73 @@ protected override Func<IKernelInternal, IReleasePolicy, object> BuildFactoryCom
...
+
+        private class TaskHelperClass
+        {
+            public static Task<T> CastTaskResult<T>(Task<object> task)
+            {
+                return Inner();
+
+                async Task<T> Inner()
+                {
+                    var result = await task.ConfigureAwait(false);
```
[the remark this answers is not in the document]
dbeuchler: I want to keep it here. Its library code which can run in different scenarios (in our case only asp where it is not relevant, but we dont know. Good library code should be aware of that. If there is no synchronization context, it has no relevance.

PR #921, opened 2025-12-11
Add option to pre-allocate a Job in the Process Engine

1nf0rmagician:
WIP...


dbeuchler:
Changes are looking really good. Tests and Docs should be extended, but from my POV after api-freeze. We should think about making the new Dispatcher as the new default and removing the SingleJobOperationDispatcher.

Please let me know when it is out of draft state.

PR #950, opened 2026-01-05
WIP: Fix `StartAllModules` returns before all modules have startet

1nf0rmagician:
### Summary
- Runs branches of the module tree in parallel
- Awaits calls to module start and initialize methods
- Does not put work for module starts on seperate thread
- Removes unreliable timing issues in tests

TODO:
- Even though the code compiles and all tests run, executing the StartProject runs into a Deadlock
- The code and even the tests expect the initialize method to be called twice; which does not make sense...


dbeuchler:
I will close this PR. Reasons described in #1483. Iw ill create a new ticket for reworking StartModulesAsync.

PR #951, opened 2026-01-05
Fixed `NullReferenceException` if no Launcher config was supplied

1nf0rmagician:
### Summary
Issue was caused by typo in `ShellNavigator.GetConfiguration`. Also adds null checks for conbfiguration to provide a better error experience.

### DIFF
--- a/src/Moryx/Configuration/ConfigManagerExtensions.cs
+++ b/src/Moryx/Configuration/ConfigManagerExtensions.cs
@@ -81,6 +81,7 @@ namespace Moryx.Configuration
         /// <param name="configuration">Object to save</param>
         public static void SaveConfiguration(this IConfigManager configManager, ConfigBase configuration)
         {
+            ArgumentNullException.ThrowIfNull(configuration);
             configManager.SaveConfiguration(configuration, configuration.GetType().FullName, false);
         }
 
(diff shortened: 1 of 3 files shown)


src/Moryx/Configuration/ConfigManagerExtensions.cs
```diff
@@ -81,6 +81,7 @@ public static ConfigBase GetConfiguration(this IConfigManager configManager, Typ
         /// <param name="configuration">Object to save</param>
         public static void SaveConfiguration(this IConfigManager configManager, ConfigBase configuration)
         {
+            ArgumentNullException.ThrowIfNull(configuration);
```
dbeuchler: Do we need this here? You have added the exception to the ConfigManager, for me this is duplicate.

1nf0rmagician: Yes, because `configuration.GetType().FullName` still throws a `NullReferenceException` before we reach the config manager itself. In the specific case we had here, the Exception would still have been nullref instead of parameter null for example.

In https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/951/changes/BASE..e40d4632042ba5d836686c5b6d21e9749195d6dd#diff-17c7f2e4509c99b39fb84f8683f5ae65315fb43315af36c408045682206fa52bL106, I did not add it, as there it would actually be a duplicate

PR #953, opened 2026-01-06
Fixed missleading migration toast notification

dbeuchler:


1nf0rmagician (changes requested):

src/Moryx.Model/Configuration/ModelConfiguratorBase.cs
```diff
@@ -200,11 +208,11 @@ protected async Task<IReadOnlyList<string>> AppliedMigrationsAsync(DbContext con
             try
             {
                 var appliedMigrations = await context.Database.GetAppliedMigrationsAsync(cancellationToken);
-
                 return appliedMigrations.ToArray();
             }
-            catch (Exception)
+            catch (Exception e)
             {
+                Logger.LogError(e, "Could not retrieve applied database migrations for context '{0}' failed!", context.GetType().Name);
```
seveneleven:
```suggestion
                Logger.LogError(e, "Could not retrieve applied database migrations for context '{0}'!", context.GetType().Name);
```

PR #956, opened 2026-01-06
Added support for publishing notifications async

dbeuchler:
NotificationAdapter supports async publishing now. `PublishAsync` will return `true` if the notification was published and processed, and `false` if it was not processed before acknowledgement. It adds the ability to await the processing of the notification to get information (like texts) from the processed notification.

The implementation is a POC. For the 10.0.0 release, we could just add the methods to the INotificationAdapter interface as preparation and add feature implementation in 10.x. @1nf0rmagician does this make sense for you?


dbeuchler:
The pull request wasn't based on a specific requirement. It was an idea, but waiting via a callback within a state machine is probably the way to go.

PR #996, opened 2026-01-13
Marked Initialized and InitializedAsync obsolete in ResourceInitializerBase

dbeuchler:
Renamed to `IntializedWith(resources)` and `InitializedAsyncWith(resources)`. 

@1nf0rmagician this ist my best idea, do you have another one?


1nf0rmagician:
While I agree with the change in itself, could we keep the old methods as well for MORYX 10? 

While the API fix in #993 should not have caused any issues as the exceptions were not usable before, and the changes in itself were not breaking, this API fix would introduce a breaking change.

I propose we leave the old methods and mark them obsolete hinting towards the new methods. We can fix that on future right away, but I think it is to easy to not make this a "minor conform" API fix.

1nf0rmagician (approved):

PR #1021, opened 2026-01-18
Allow hosting with an URL Path Prefix

andreniggemann:
### Summary

This feature tries to establish support for hosting the MORYX Web Applications under a prefix.
This can be very useful if MORYX is run behind reverse proxy together with other services.

For instance when hosting Grafana and Moryx on the same host via the default https port via a reverse proxy
one might like to use `https://moryx-host/grafana/...` to reach grafana and `https://moryx-host/moryx/...` to access the moryx application. The alternative would be host base routing like `https://grafana.moryx-host/` but that introduces new challenges for  DNS and TLS certificate management.

I tried a very low tech approach that works very well. I did not find any problems yet (see the test setup if you want to try it yourself) . The only thing I am not very happy with is the duplication of code on the type script side.  Maybe there is an opportunity to put that into some shared library.

### Checklist for Submitter

### Review

**Typical tasks**

**Clean code**

### DIFF
--- a/docs/articles/reverse-proxies/run-behind-reverse-proxy.md
+++ b/docs/articles/reverse-proxies/run-behind-reverse-proxy.md
@@ -12,6 +12,7 @@ This can be done in one of two ways. Either by hostname based routing or by path
 ## Setup
 If you want to follow along with the examples you can use this docker-compose file together with the nginx configuration examples to setup a reverse proxy and modify the StartProject.Asp project to see if everything works as expected.
 The nginx configuration needs to be placed in a file named `nginx.conf` next to your `docker-compose.yml` when executing `docker compose up`.
+
 ```yaml
 services:
   nginx-proxy:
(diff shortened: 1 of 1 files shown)


src/Moryx.Launcher/Pages/Shared/_Layout.cshtml
```diff
@@ -28,13 +28,15 @@
 <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <!-- could compute from base tag, but I think this is clearer -->
+    <meta name="moryx-pathbase" content="@Context.Request.PathBase" />
```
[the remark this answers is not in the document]
andreniggemann: We could take it from the <base href=.. /> tag. That contains the prefix as well as the module path.
This means instead of just reading the value you would have a path splitting operation. Definitely doable. Especially if we hide it in our framework library.
Because I was working with duplicated code, I wanted to keep it as simple as possible on the Typescript side.

Regarding the exchangeable launcher: I guess our UIs at least make the assumption to run in a programmable ASP.NET based launcher. This means it would be trivial to add this metadata as needed in other launchers.

docs/articles/reverse-proxies/run-behind-reverse-proxy.md
```diff
@@ -0,0 +1,137 @@
+# Run behind reverse proxy
+
+Running behind a reverse proxy is a common scenario for web applications.
+That's why there is proper support for this in ASP.NET Core and hence in MORYX.
+There is ample documentation for it by [Microsoft](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balance).
```
dbeuchler:
```suggestion
There is sample documentation for it by [Microsoft](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balance).
```

docs/articles/reverse-proxies/run-behind-reverse-proxy.md
```diff
@@ -0,0 +1,137 @@
...
+
+```html
+<!--absolute-->
+<script src="/_content/<YourC#ProjectName>/main.js" type="module"></script>
+
+<!-- base path aware -->
+<script src="~/_content/<YourC#ProjectName>/main.js" type="module"></script>
+
+<!-- Example eventstream annotation from VisualInstructions.cshtml -- >
+[ModuleEventStream("~/api/moryx/instructions/stream")]
```
dbeuchler:
```suggestion
[ModuleEventStream("~/api/moryx/<your-api>/stream")]
```

PR #1035, opened 2026-01-21
Casting `(Func<Resource, Resource>)` to `(Func<Resource, IEnumerable<…

seveneleven:
…Resource>>)` threw exception


src/Tests/Moryx.AbstractionLayer.Tests/ResourceExtensionsTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using Moryx.Resources.Samples;
+using NUnit.Framework;
+
+namespace Moryx.AbstractionLayer.Tests;
+
+[TestFixture]
+public class ResourceExtensionsGetFirstRelatedResourceTests
+{
+    [Test]
+    public void Returns_First_Related_Resource()
```
1nf0rmagician: Don't want to be picky but prevent missunderstandings, the .net convention for test naming is not about underscrores but three distinct parts of the name seperated by underscores 🙈
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#follow-test-naming-standards

PR #1055, opened 2026-01-27
Upgrade: Moryx.CommandCenter.Web dependencies

dbeuchler:
Not directly angular but also part of the launcher and upgraded all deps

### DIFF
--- a/src/Moryx.CommandCenter.Web/app/tsconfig.json
+++ b/src/Moryx.CommandCenter.Web/app/tsconfig.json
@@ -4,7 +4,7 @@
         "sourceMap": true,
         "noImplicitAny": true,
         "module": "commonjs",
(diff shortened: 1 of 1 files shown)


src/Moryx.CommandCenter.Web/app/tsconfig.json
```diff
@@ -5,7 +5,8 @@
         "noImplicitAny": true,
         "module": "commonjs",
         "target": "es6",
```
1nf0rmagician: Wanna update this to es2022 as well? Does not provide any benefit to keep it on es6 if all other modules are build for 2022, does it?

1nf0rmagician (approved):

1nf0rmagician (approved):

PR #1072, opened 2026-02-05
Fix duplicate DisplayName ArgumentException

robertsmns:
This PR adds safe handling for duplicate DisplayName values within ToCellPropertySettings to prevent ArgumentException caused by duplicate dictionary keys.


src/Moryx.FactoryMonitor.Endpoints/Converter/Converter.cs
```diff
@@ -69,34 +80,64 @@ public static CellStateChangedModel ToCellStateChangedModel(Resource current)
...
             var property = new CellPropertySettings
             {
                 IsDisplayed = false,
                 CurrentValue = cellEntry.Value.Current,
                 IconName = entryVisualizer?.Icon,
                 ValueUnit = entryVisualizer?.Unit,
             };
-            cellProperties.Add(cellEntry.DisplayName, property);
+
+            if (cellProperties.ContainsKey(cellEntry.DisplayName))
```
1nf0rmagician: I think adding the display name of the property to the `CellPropertySettings` and using a really unique key like the property name would be more preferable here, then just replacing it and logging a warning 🤔 The display name is never a good dictionary key 🙈
Mind, however, you will have to adjust that on the UI side as well then.

1nf0rmagician (changes requested):

1nf0rmagician:
@MathoMathiasCamara could you rebase the branch on the latest dev? there are conflicts and i would make the review when they are resolved 🙈

src/Moryx.FactoryMonitor.Endpoints/Converter/Converter.cs
```diff
@@ -143,4 +169,6 @@ public static TransportRouteModel ToTransportRouteModel(TransportPathModel pathM
             Paths = pathModel.WayPoints
         };
     }
+
+    public static FactoryStateModel ToFactoryStateModel(IManufacturingFactory factory) => new() { Id = factory.Id };
```
1nf0rmagician: Sure this is still used? Might be an artifact of the merge 🙈

src/Tests/Moryx.FactoryMonitor.Endpoints.Tests/BaseTest.cs
```diff
@@ -39,15 +40,16 @@ public abstract class BaseTest
     protected Position _assemblyCellposition;
     protected Position _solderingCellPosition;
     protected ResourceGraphMock _graph;
-
+    protected Mock<ILogger<FactoryMonitorController>> _logger;
```
1nf0rmagician: While it's fine for me, there is also a .net logger instance for tests `NullLogger<T>.Instance` 😊

src/Moryx.FactoryMonitor.Endpoints/Converter/Converter.cs
MathoMathiasCamara: yeah i think so.

src/Tests/Moryx.FactoryMonitor.Endpoints.Tests/BaseTest.cs
MathoMathiasCamara: Nice !

1nf0rmagician (approved):

Issue #1081, opened 2026-02-11
Operations are not sorted by their index


dbeuchler:
**Describe the bug**

Operations are not sorted by their sort index in Web UI

dbeuchler:
completed in #1093

Issue #1079, opened 2026-02-11
Missing Success-count and broken progress in Orders.Web


dbeuchler:
**Describe the bug**

<img width="651" height="256" alt="Image" src="https://github.com/user-attachments/assets/fb43d61a-392c-43a0-ad2c-3b49023de842" />

Was removed in https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/23f5d6e9402bbff75b8ca808020f906af43e1818#diff-69cbfe47a6609865e7ca65ec5cd2a0f11f14d754c299d2a5fd34edd5a8ff0951L140-R157

Progress tooltip is x100. E.g. 1000% for 10 articles.

PR #1093, opened 2026-02-23
Improvement of Orders.Web

### LINKED ISSUE #1080: Operation state indicator is always black
**Describe the bug**
The operation state indicator is always black so it is not possible to see the current state of an operation in the list

**Wish**

Restore color code from WPF UI:

- Initial = White or Gray
- Running = PxC Light Green
- Interrupting = DarkOrange
- Ready / Interrupted = PxC Turquoise
- Failed = Red

Image

Image

dbeuchler:
- restores color codes from older UIs #1080
- added SuccessCount back, fixed tooltips #1079
- added hide-completed selection (known from wpf UIs) #1095
- Operations are sorted by their index and then by PlannedStart #1081
- Added side-nav for internal module tools
- reused progress-bar in begin-dialog and operations

close #1080 #1079 #1095 #1081

Screenshot 2026-02-26 at 15 58 18

Screenshot 2026-02-26 at 15 58 32

Screenshot 2026-02-26 at 15 58 52

Screenshot 2026-02-26 at 15 59 52

### DIFF
--- a/src/Moryx.Orders.Web/app/src/app/components/operations/operations.html
+++ b/src/Moryx.Orders.Web/app/src/app/components/operations/operations.html
@@ -147,28 +147,13 @@
                           </div>
                         </div>
                         <div class="operation-header-row operation-progress">
-                          <div class="multi-progress-container">
-                            <div class="bar success"
-                                 [style.flex-basis.%]="operation.progressSuccessPercent"
-                                 matTooltip="{{ TranslationConstants.OPERATIONS.SUCCESS | translate }}: {{ operation.model.successCount }}">
-                            </div>
-                            <div class="bar scrap"
-                                 [style.flex-basis.%]="operation.progressScrapPercent"
-                                 matTooltip="{{ TranslationConstants.OPERATIONS.SCRAP | translate }}: {{ operation.model.scrapCount }}">
-                            </div>
-                            <div class="bar running"
-                                 [style.flex-basis.%]="operation.progressRunningPercent"
-                                 matTooltip="{{ TranslationConstants.OPERATIONS.RUNNING | translate }}: {{ operation.model.runningCount }}">
-                            </div>
-                            <div class="bar pending"
-                                 [style.flex-basis.%]="operation.progressPendingPercent"
-                                 matTooltip="{{ TranslationConstants.OPERATIONS.PENDING | translate }}: {{ operation.model.pendingCount }}">
-                            </div>
-                            <div class="bar residual"
-                                 [style.flex-basis.%]="operation.progressResidualPercent"
-                                 matTooltip="{{ TranslationConstants.OPERATIONS.RESIDUAL | translate }}: {{ operation.progressResidualCount }}">
-                            </div>
-                          </div>
+                          <app-multi-progress-bar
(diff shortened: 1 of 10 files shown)


src/Moryx.Orders.Web/app/src/app/components/operations/operations.html
```diff
@@ -61,220 +61,143 @@ <h2>
...
-                          operation.model.classification === 'Running' &&
-                          operation.isProducing
-                          ) {
-                          <mat-icon
-                            color="accent"
-                            class="operation-state-indicator operation-indicator-animation"
-                            >settings
-                          </mat-icon>
+                      @case (OperationStateClassification.Running) {
+                        <mat-icon color="accent"
```
1nf0rmagician: You define the color twice here, through the color and from the running style, don't you?

1nf0rmagician (changes requested):
Could we split this PR in the fixes and the addtion of the hide/diplay completed feature? The fixes seam fine to me, even though the html part is really hard to review the way github shows the diff unfortunately 🙈

The design of hide-completed feature directly violates the material design guideline, however. As only one FAB button should be displayed per screen, [source](https://m2.material.io/components/buttons-floating-action-button#usage). Also adding the feature should already take into account that there is another pending feature request to filter orders e.g. by product, recipe, operation number which serve a similar purpose and hence belong into a coherent "how do we display filter options" concept for the UI.  
@MathoMathiasCamara would be interested in your thoughts on that matter. Maybe a top-bar (like shifts UI), maybe a bottom-bar? The products UI has filters in the left component, which is unfortunately not applicable here. The Visual Instructions UI basically filters instructions by client, which is configured through a settings button in the top right...

Also, considering that there are more filter options coming, I think separating the local storage interaction into a service would be nice

MathoMathiasCamara:
> Could we split this PR in the fixes and the addtion of the hide/diplay completed feature? The fixes seam fine to me, even though the html part is really hard to review the way github shows the diff unfortunately 🙈
> 
> The design of hide-completed feature directly violates the material design guideline, however. As only one FAB button should be displayed per screen, [source](https://m2.material.io/components/buttons-floating-action-button#usage). Also adding the feature should already take into account that there is another pending feature request to filter orders e.g. by product, recipe, operation number which serve a similar purpose and hence belong into a coherent "how do we display filter options" concept for the UI. @MathoMathiasCamara would be interested in your thoughts on that matter. Maybe a top-bar (like shifts UI), maybe a bottom-bar? The products UI has filters in the left component, which is unfortunately not applicable here. The Visual Instructions UI basically filters instructions by client, which is configured through a settings button in the top right...
> 
> Also, considering that there are more filter options coming, I think separating the local storage interaction into a service would be nice

Yes i think having the top bar here is a good idea.

src/Moryx.Orders.Web/app/src/app/components/operations/operations-filter/operations-filter.html
```diff
@@ -0,0 +1,5 @@
+<p>
```
1nf0rmagician: Any reason for this additional dom element?

PR #1096, opened 2026-02-24
Fixed execution of resource methods

dbeuchler:
- reduced complexity with separate service
- made ResourceMethod signal-based

### DIFF
--- a/src/Moryx.Resources.Web/app/src/app/components/details-view/resource-methods/resource-methods.ts
+++ b/src/Moryx.Resources.Web/app/src/app/components/details-view/resource-methods/resource-methods.ts
@@ -3,7 +3,8 @@
  * Licensed under the Apache License, Version 2.0
 */
 
-import { Component, inject, signal } from '@angular/core';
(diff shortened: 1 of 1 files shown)


src/Moryx.Resources.Web/app/src/app/components/details-view/resource-methods/resource-methods.ts
```diff
@@ -23,14 +36,123 @@ import { MatButtonModule } from '@angular/material/button';
...
+
+  selectedMethod = signal<MethodEntry | undefined>(undefined);
+  methodResult = signal<Entry | undefined>(undefined);
+  resultView = signal(false);
+
   TranslationConstants = TranslationConstants;
 
-  ngOnInit(): void {
+  constructor() {
+    this.editResourceService.activeResource$.subscribe((resource) => {
```
1nf0rmagician: In the service it was not necessary to unsubscribe as the service was a singleton. Within the component, we need to unsubscribe from this observable, however. Probably best to place it in oninit and ondestroy here🤔

PR #1132, opened 2026-02-27
Harmonized openapi-gen.json, added to package.json and added fixup script

dbeuchler:
- Every web project uses the same file-name for api-definition
 - Added ng-openapi-gen as dev-dependency to use defined version on each generation
 - Node.js does not trust self-signed certificates by default (which localhost typically uses for HTTPS)
   - Workaround by NODE_TLS_REJECT_UNAUTHORIZED=0 
- Added script to replace entry files with ngx-web-framework exports and adding copyright header to generated files

I regenerated all apis with the new script. The location of the script is currently in ./src folder because I have no better location for it. Maybe we add it to the web-framework later if it is interesting in other projects too. 

There are tons of changes to the last empty line in the generated files. This comes from new ng-openapi-gen version. For review, do it commit by commit.


src/Moryx.Resources.Web/app/openapi-gen.json
```diff
@@ -0,0 +1,13 @@
+{
```
1nf0rmagician: Don't you need the include tags here?

PR #1142, opened 2026-03-04
Port #1140: Limit module notifications 

seveneleven:
To not exceed memory with too many notifications over time, that are usually
created through logging errors and warnings, the notifications collection now
acts as a ringbuffer, limited to 5000 entries.

This change is breaking bevaviour in the way, that it changes the expected
output. But as discussed, that should be considered less an issue than having
an application to crash.


dbeuchler (changes requested):
Could you rebase the branch to dev do not merge all merge commits from release?

seveneleven:
> Could you rebase the branch to dev do not merge all merge commits from release?

Of course, I didn't like that either.

PR #1194, opened 2026-04-13
Refactor: Rework ResourceModificationController to conform to HTTP and OAS standards

mschnuell:
## Context

The `ResourceModificationController` is the entry point for all resource management operations. It works, but it accumulated a number of design decisions that quietly violated HTTP semantics and produced misleading OpenAPI output.
This PR serves as a proof-of-concept. The controller is the first to be brought fully in line with the new standard.

### HTTP method corrections

The query endpoint was previously a `POST`, even though it only reads data. This made the intent of the endpoint unclear.
It has been changed to `GET` with `[FromQuery]` binding, which is a better approach for read-only, parameterized queries.

### Response type cleanup

Several endpoints used `Status417ExpectationFailed` as a catch-all error code. 417 is a very specific HTTP status meant for the expect request header. These have been replaced with `Status500InternalServerError` for unexpected server-site failures and `Statuis400BadRequest` where invalid input is the likely cause.
Additionally, all `ProducesResponseType` attributes now carry the actual response model type (e.g. `typeof(ResourceModel)`), so the generated OpenAPI schema reflects what the end point actually returns instead of being untyped.

### Route attribute consolidation

Routes were previously declared with a separate `[Route]` attribute alongside `[HttpGet]`/`[HttpPost]`. Summarizing the `[Route]` attribute directly into the HTTP method attribute (e.g. `[HttpGet("{id}")]`) is the proper ASP.NET Core style and reduces visual noise.

### XML documentation

All endpoints now have XML doc comments. This feeds directly into Swagger UI and the generated OpenAPI spec, giving consumers meaningful descriptions for each endpoint, its parameters, and its return value - rather than just an auto-generated stub.


dbeuchler:
Can you describe what is wrong / check the changed line endings in the change that we can see the diff of the controller? If there is any change in the API, we must move this to future.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -100,44 +122,78 @@ public ActionResult<ResourceModel> GetDetails(long id)
...
+    /// or an empty body when the action returns void (204 No Content).
+    /// </returns>
+    [HttpPost("{id}/actions/{action}")]
+    [ProducesResponseType(typeof(Entry), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(typeof(MoryxExceptionResponse), StatusCodes.Status404NotFound)]
+    [ProducesResponseType(typeof(MoryxExceptionResponse), StatusCodes.Status422UnprocessableEntity)]
     [Authorize(Policy = ResourcePermissions.CanInvokeMethod)]
-    public async Task<ActionResult<Entry>> InvokeMethod(long id, string method, Entry parameters)
+    public async Task<ActionResult<Entry>> InvokeAction(long id, string action, Entry parameters)
```
1nf0rmagician: Why did you change here to Action instead of method? The term method is used in the UI as well and is more meaningful 🙈

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -100,44 +122,78 @@ public ActionResult<ResourceModel> GetDetails(long id)
...
             {
-                entry = EntryConvert.InvokeMethod(r.Descriptor, new MethodEntry { Name = method, Parameters = parameters }, _serialization);
+                entry = EntryConvert.InvokeMethod(r.Descriptor, new MethodEntry { Name = action, Parameters = parameters }, _serialization);
                 return Task.FromResult(true);
             });
         }
         catch (MissingMethodException)
         {
-            return BadRequest("Method could not be invoked. Please check spelling and access modifier (has to be `public` or `internal`).");
+            return NotFound(new MoryxExceptionResponse { Title = $"Action '{action}' does not exist on resource {id}." });
```
1nf0rmagician: The information that the method need to be internal or publis on the resource is helpful, I wouldn't drop it for no reason 🙈

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -100,44 +122,78 @@ public ActionResult<ResourceModel> GetDetails(long id)
...
     }
 
-    [HttpPost]
-    [ProducesResponseType(StatusCodes.Status200OK)]
-    [ProducesResponseType(StatusCodes.Status404NotFound)]
-    [ProducesResponseType(StatusCodes.Status417ExpectationFailed)]
-    [Route("types/{type}")]
+    /// <summary>
+    /// Constructs a new resource instance of the specified type without persisting it.
+    /// Optionally invokes a constructor method with the supplied arguments, which does persist it.
```
1nf0rmagician:
```suggestion
    /// Optionally invokes a constructor method with the supplied arguments, which does persist the resource in the process.
```

1nf0rmagician: Also you could reference the `RecourseConstructor` specifically in this constructor to prevent missunderstandings 🙈

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -100,44 +122,78 @@ public ActionResult<ResourceModel> GetDetails(long id)
...
+    /// </summary>
+    /// <param name="type">
+    /// The resource type name to construct (e.g. MyNamespace.MyResource).
+    /// Available types are returned by GET /types.
+    /// </param>
+    /// <param name="method">
+    /// The name of an optional constructor method to invoke.
```
1nf0rmagician, line 180: specifically the full name of the type

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -69,12 +82,17 @@ public ActionResult<ResourceModel[]> GetDetailsBatch([FromQuery] long[] ids)
...
+    /// </summary>
+    /// <param name="query">Query filter passed in the URL.</param>
+    /// <returns>An array of resource models.</returns>
+    [HttpGet("query")]
+    [ProducesResponseType(typeof(ResourceModel[]), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(MoryxExceptionResponse), StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [Authorize(Policy = ResourcePermissions.CanViewTree)]
-    public ActionResult<ResourceModel[]> GetResources(ResourceQuery query)
+    public ActionResult<ResourceModel[]> GetResources([FromQuery] ResourceQuery query)
```
1nf0rmagician, line 95: Do you have any stronger arguments then this post for replacing this with POST+FromQuery? 
https://stackoverflow.com/a/37145044/14014749

Especially, have you verified, that .Net can properly deserialize the complex type form query in contrast to the statement in the post?

src/StartProject.Asp/Startup.cs
```diff
@@ -57,11 +57,17 @@ public void ConfigureServices(IServiceCollection services)
...
-            .AddJsonOptions(jo => jo.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter()));
-
+            .AddJsonOptions(jo => jo.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter()));
+
         services.AddSwaggerGen(c =>
         {
             c.CustomOperationIds(api => ((ControllerActionDescriptor)api.ActionDescriptor).MethodInfo.Name);
+
+            //Include XML comments
+            var xmlFile = $"{typeof(ResourceModificationController).Assembly.GetName().Name}.xml";
```
1nf0rmagician, line 67: Could you check whether you can instead include the xml comments from all subclasses of `ControllerBase`? 🤔

1nf0rmagician (changes requested):

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: As far as I know this is simply a common practice. I think it stems mainly from "method" being a keyword for POST, GET, PUT, etc. in REST/HTTP context.
However, you've got a good point with the term being used in the UI as well. If you'd prefer, I can change it back to method.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: Is this better, or did I misunderstand what you meant?

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: You're right, I've adjusted the message.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: Done

src/StartProject.Asp/Startup.cs
mschnuell: I think that would be a good idea, but I think that solution would be quite a lot more complex. I think it would be better to adjust that in the future.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
dbeuchler: @1nf0rmagician is right, Method is the term from our Entry-Format and this should be consistent

src/StartProject.Asp/Startup.cs
dbeuchler: Is there a way in swagger to provide the docs by attributes instead of loading the xml file? In release builds the xml is not part of the application. So every application must be configured to load the xml docs and also xml docs must be enabled in every release build explicitly.

As far as I know DataAnnotations Attributes are automatically picked. `Description` e.g, can be added by filters. Using DataAnnotations attributes with a custom Swashbuckle filter gives you a more maintainable API documentation without the deployment and application complexity.

If the XML documentation is to be retained, then I see an extension here that would make the XMLs available for all ControllerBase. That would mean yet another package referencing Swashbuckle. I did like to avoid that somehow. Perhaps using DataAnnotations would be okay?

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: Alright, I've changed it back to InvokeMethod.

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
mschnuell: I did some additional research on this topic and found that there is no strict best practice for this kind of endpoint.
The general opinions on this topic go to both sides:
1. Either use POST with the query in the body (This is technically not forbidden in HTTP semantics)
2. Or use GET with the query in the URL string (The HTTP semantics could be interpreted so that this is the intended way)

I've also ran several manual tests on the updated endpoint, including:
- Empty queries
- Simple and complex queries
- URL length limits

All tests passed successfully, and the objects were correctly initialized in every case.

I've also found two alternatives approaches to this implementation:
1. Either flattening complex objects
2. Or keeping the current implementation with POST + Query in Body

Given this, I believe using GET is the best practice for this specific use case because:
- It is more true to the intended HTTP semantics
- The implementation is tested and works as intended
- The query objects are not complex enough to reach the URL length limit
- The resulting URLs still remain reasonably readable

Although I've noticed one limitation with swagger:
The UI does not correctly format some parameters (e.g. the IncludedReference JSON). Therefore, when testing in Swagger, the URL may needs to be adjusted manually.

PR #1202, opened 2026-04-24
Include startup hooks

andreniggemann:
### Summary

This feature adds a standardized and reusable way to add logic to the startup of a MORYX application

### Breaking Changes

No breaking changes

### Checklist for Submitter

### Review

**Typical tasks**

**Clean code**


docs/articles/startup/startuphooks.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Definition
+
+Startup hooks implement the lightweight IStartupHook interface.
+That just gives them a priority to make sure they run in the correct order and a RunAsync function to execute the functionality.
+
+## Usage
+
+After including the package and namespace "Moryx.Startup.Hooks" you can use the
+extension methods AddStartupHooks() and RunHooksAsync() in your program.cs to register and then execute all StartupHooks.
```
1nf0rmagician: Open Question: Do we want to keep this optional  or extend the central extensions on the service collection, e.g. `AddMoryxService` to automatically include this feature?

dbeuchler:
After discussion with @andreniggemann, @1nf0rmagician  and me, here a short summary

- Move module specifics to the module responsibility
  - (Optional for this PR): Add possibility to execute ResourceInitializer on startup
  - (Optional for this PR): Add possibility to execute ProductImporter on startup
- Move IStartUpHooks (and additional apis) to Moryx.Runtime (implementation to Moryx.Runtime.Kernel)
- (Optional for this PR): Create extension for IDbContextManager to create databases, remove all, clear all....
- Use Moryx.Configuration instead of ASP.NET IConfiguration (to be consitent in MORYX 10)

andreniggemann:
@dbeuchler and @1nf0rmagician after the long delay I cleaned up this PR and modified it according to our discussion and building on top of the changes in the config PR [1259](https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1259)

I only included the optional item to create the  IDbContextManager.
I did not include the Resource and Products yet. After some discussion with @1nf0rmagician  this was not so easy as it first appeared, especially for the ResourceManagement.

docs/articles/framework/lifecycle-hooks.md
```diff
@@ -0,0 +1,176 @@
...
+
+To make use of this hook, you must configure at least one operation via `OrdersHookConfig`.
+
+Each operation configuration requires:
+
+- Order number
+- Operation number
+- Product identifier
+- Product revision
+- Quantity to produce
```
1nf0rmagician, line 77: Which recipe is selected, the default one?

docs/articles/framework/startup-hooks.md
```diff
@@ -0,0 +1,176 @@
...
+## Summary
+
+Startup hooks provide a clean and extensible way to prepare your application before it starts:
+
+- Execute pre-start logic in a structured way
+- Keep setup code modular and reusable
+- Integrate with DI and configuration systems
+- Control execution order via priorities
+
+By leveraging startup hooks, you can improve reliability, simplify testing, and ensure your application always starts in the expected state.
```
1nf0rmagician: Could be put as a summary at the beginning, e.g. in a note panel

> [!NOTE]  
> Startup hooks provide a clean and extensible way to prepare your application before it starts:
> 
> - Execute pre-start logic in a structured way
> - Keep setup code modular and reusable
> - Integrate with DI and configuration systems
> - Control execution order via priorities
> 
> By leveraging startup hooks, you can improve reliability, simplify testing, and ensure your application always starts in the expected state.

docs/articles/framework/startup-hooks.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+### What happens under the hood?
+
+- `AddStartupHooks()` scans for all implementations of `IStartupHook` and registers them in the service collection.
+- `RunHooks()` resolves all registered hooks and executes them in order of their configured priority.
+
+## Predefined Hooks
+
+The MORYX framework already provides ready-to-use hooks in the `Moryx.Startup.Hooks` project.
```
1nf0rmagician: Alternative package name suggestion: `Moryx.Hooks.Samples` similar to the resource samples?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+            await dbContextManager.DeleteDatabaseInternalAsync(context, logger, token);
+        }
+
+        /// <summary>
+        /// Deletes the database for the context with the given name
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="logger"></param>
+        /// <param name="token"></param>
+        /// <returns></returns>
```
1nf0rmagician: Fill or remove for all the methods in here 🙂

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+        {
+
+            var dbContextManager = serviceProvider.GetRequiredService<IDbContextManager>();
+            logger ??= serviceProvider.GetService<ILogger>();
+            await dbContextManager.DeleteDatabaseByNameAsync(name, logger, token);
+        }
+
+        
+    }
+    extension (IDbContextManager dbContextManager)
```
1nf0rmagician: lines

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+                var configurator = dbContextManager.GetConfigurator(context);
+                var testResult = await configurator.TestConnectionAsync(configurator.Config, token);
+
+                if (testResult is TestConnectionResult.ConnectionOkDbDoesNotExist)
+                {
+                    logger?.LogInformation("Db for {db} does not exist. Trying to create", context.Name);
+                    var createResult = await configurator.CreateDatabaseAsync(configurator.Config, token);
+                    if (createResult)
+                    {
+                        logger.LogInformation("Successfully created database for context {context}", context.Name);
```
1nf0rmagician: missing null check on logger, maybe check with ctrl+f if i missed it somewhere as well

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+                {
+                    logger?.LogDebug("Skipping context {context} because db already exists", context.Name);
+                }
+                else if (testResult is TestConnectionResult.PendingMigrations)
+                {
+                    var summary = await configurator.MigrateDatabaseAsync(configurator.Config, token);
+                    if (summary.Result is Configuration.MigrationResult.Error)
+                    {
+                        logger?.LogWarning("Failed to apply missing migration to context {context}", context.Name);
+                    }
```
1nf0rmagician, line 117: dont't you need to create the db after applying migrations?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+                else if (testResult is TestConnectionResult.Success)
+                {
+                    logger?.LogDebug("Skipping context {context} because db already exists", context.Name);
+                }
+                else if (testResult is TestConnectionResult.PendingMigrations)
+                {
+                    var summary = await configurator.MigrateDatabaseAsync(configurator.Config, token);
+                    if (summary.Result is Configuration.MigrationResult.Error)
+                    {
+                        logger?.LogWarning("Failed to apply missing migration to context {context}", context.Name);
```
1nf0rmagician: error log`?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+                    }
+                }
+                else if (testResult is TestConnectionResult.ConfigurationError or TestConnectionResult.ConnectionError)
+                {
+                    logger?.LogWarning("Skipping context {context}, because the configuration is in state {result}", context.Name, testResult);
```
1nf0rmagician: error log?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+    extension (IDbContextManager dbContextManager)
+    {
+        /// <summary>
+        /// Tries to create the database for each context.
+        /// Applies missing migrations as well.
+        /// </summary>
+        /// <param name="logger"></param>
+        /// <param name="token"></param>
+        /// <returns></returns>
+        public async Task CreateAllConfiguredDatabasesAsync(ILogger logger, CancellationToken token = default)
```
1nf0rmagician: Not sure but matching your comment you could go for the `Try...` method structure and return a boolean to indicate success?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+        {
+            foreach (var context in dbContextManager.Contexts)
+            {
+                if (context.Name == name || context.FullName == name)
+                {
+                    await DeleteDatabaseInternalAsync(dbContextManager, context, logger, token);
+                    return;
+                }
+            }
+            logger?.LogWarning("No context with the name {name} found", name);
```
1nf0rmagician: Wild not to see this as a linq statement. I mean it is correct but weird xD

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+            {
+                await configurator.DeleteDatabaseAsync(configurator.Config, token);
+                var result = await configurator.TestConnectionAsync(configurator.Config, token);
+                if (result is TestConnectionResult.ConnectionOkDbDoesNotExist)
+                {
+                    logger.LogInformation("Successfully deleted database for context {context}", context.Name);
+                }
+                else
+                {
+                    logger.LogWarning("Failed to delete db for context {context}", context.Name);
```
1nf0rmagician: error log?

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,206 @@ namespace Moryx.Model;
...
+                {
+                    await DeleteDatabaseInternalAsync(dbContextManager, context, logger, token);
+                    return;
+                }
+            }
+            logger?.LogWarning("No context with the name {name} found", name);
+        }
+
+
+        private async Task DeleteDatabaseInternalAsync(Type context, ILogger logger, CancellationToken token = default)
```
1nf0rmagician: Same as for the create you coul consider using Try... with boolean return

src/Moryx.Runtime.Kernel/HookExtension.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Moryx.Runtime.Kernel;
+
+public static class HookExtension
+{
+    extension(IServiceCollection services)
+    {
+        public IServiceCollection AddStartupHooks()
+        {
+            var hooks = ReflectionTool.GetPublicClasses<IStartupHook>(h => h.IsClass && !h.IsAbstract);
```
1nf0rmagician: The filter is unnecessary 

https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/f31098b9cc80d21ad8b13c377aa2db987c28802c/src/Moryx/Tools/ReflectionTool.cs?plain=1#L61C1-L61C70

src/Moryx.Runtime.Kernel/HookExtension.cs
```diff
@@ -0,0 +1,37 @@
...
+            return services;
+        }
+    }
+    extension(IServiceProvider provider)
+    {
+        public async Task RunHooksAsync()
+        {
+            var hooks = provider.GetServices<IStartupHook>().ToArray();
+
+            foreach (var hook in hooks.OrderBy(h => h.Priority))
```
1nf0rmagician: want to add `ThenBy` on the name to have a defined behaviour on duplicated priority values?

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Moryx.Runtime.Modules.Hooks;
+
+
+/// <summary>
+/// Base class for StarupHooks that react to state changes of modules.
+/// </summary>
+/// <typeparam name="TFacade">The facade of the module. Is used to select the correct module</typeparam>
+/// <typeparam name="TConfig">The type of config this hook requires</typeparam>
+public abstract class ModuleHook<TFacade, TConfig> : IStartupHook where TConfig : ConfigBase, new()
```
1nf0rmagician: Some comments missing and usually we put fields above ctor I think 🙈

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// Logger to record actions and problems
+    /// </summary>
+    protected readonly ILogger _logger;
+
+    /// <summary>
+    /// <see cref="IModuleManager"/>
+    /// </summary>
+    protected readonly IModuleManager _moduleManager;
+
+    protected FunctionResult InitResult { get; }
```
1nf0rmagician: uncommon abbreviation in property name

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public ModuleHook(IModuleManager moduleManager, IConfigManager configManager, ILogger logger)
+    {
+        _logger = logger;
+        _moduleManager = moduleManager;
+
+        _config = configManager.GetConfiguration<TConfig>();
+        if (_config is null)
+        {
+            InitResult = FunctionResult.WithError("Not configured");
```
1nf0rmagician:
```suggestion
            InitResult = FunctionResult.WithError($"{this.GetType()} is not configured, skipping {naemof(Initialize)} and {nameof(RunAsync)}");
```

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            Task.Run(async () =>
+            {
+                if (sender is not IServerModule module || module is not IFacadeContainer<TFacade> facadeContainer)
+                {
+                    return;
+                }
+                await OnStateChanged(module, facadeContainer.Facade, eventArgs);
+            }).GetAwaiter().GetResult();
+        };
```
1nf0rmagician: This is asyn void isn't it?

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
```diff
@@ -0,0 +1,78 @@
...
+    public int Priority { get; set; }
+
+    protected virtual FunctionResult Initialize(TConfig config)
+    {
+        return FunctionResult.Ok();
+    }
+
+    protected abstract Task OnStateChanged(IServerModule module, TFacade facade, ModuleStateChangedEventArgs eventArgs);
+
+    public virtual async Task RunAsync()
```
1nf0rmagician: Do you want to keep this virtual? Is there a case to override it? 🤔

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+        catch (Exception ex)
+        {
+            base._logger.LogError(ex, "OnModuleStarted handler failed");
+        }
+        finally
+        {
+            _firstStart = false;
+            _semaphore.Release();
+        }
```
1nf0rmagician: You could use the [ExecuteAsync](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/f31098b9cc80d21ad8b13c377aa2db987c28802c/src/Moryx/Tools/Extensions/SemaphoreSlimExtensions.cs?plain=1#L35C1-L35C111) extension if you don't need the general exception catch and loggin.

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
```diff
@@ -0,0 +1,62 @@
...
+    protected bool _firstStart = true;
+
+    /// <summary>
+    /// Will be false until OnModuleStarted ran to completion without throwing an exception.
+    /// </summary>
+    protected bool _succeededAtLeastOnce;
+
+    private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
+
+    protected override async Task OnStateChanged(IServerModule module, TFacade facade, ModuleStateChangedEventArgs eventArgs)
```
1nf0rmagician: (Base) Method does not take a cancellation token and is not handing it over to the semaphore

src/Moryx.Runtime/IStartupHook.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+namespace Moryx.Runtime;
+
+public interface IStartupHook
```
1nf0rmagician: comment

src/Moryx.Runtime.Hooks/DatabaseHook.cs
```diff
@@ -0,0 +1,40 @@
...
+using Moryx.Model;
+using Moryx.Runtime;
+
+namespace Moryx.Startup.Hooks;
+
+public class DatabaseHook(IConfigManager configuration, IDbContextManager dbContextManager, ILogger<DatabaseHook> logger) : IStartupHook
+{
+    public int Priority => 0;
+
+    public async Task RunAsync()
```
1nf0rmagician: Should the base method take an optional cancellation token?

src/Moryx.Startup.Hooks/DatabaseHook.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Moryx.Configuration;
+using Moryx.Model;
+using Moryx.Runtime;
+
+namespace Moryx.Startup.Hooks;
+
+public class DatabaseHook(IConfigManager configuration, IDbContextManager dbContextManager, ILogger<DatabaseHook> logger) : IStartupHook
```
1nf0rmagician: comments

src/Moryx.Startup.Hooks/Moryx.Startup.Hooks.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+  <!-- 
+    <PackageReference Include="Moryx" />
+    <PackageReference Include="Moryx.Runtime" />
+    <PackageReference Include="Moryx.Orders" />
+    <PackageReference Include="Moryx.Products.Management" /> -->
```
1nf0rmagician: comments

src/Moryx.Startup.Hooks/Moryx.Startup.Hooks.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <PackageReference Include="Microsoft.Extensions.Configuration" />
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Moryx\Moryx.csproj" />
+    <ProjectReference Include="..\Moryx.Runtime\Moryx.Runtime.csproj" />
+    <ProjectReference Include="..\Moryx.Orders\Moryx.Orders.csproj" />
+    <ProjectReference Include="..\Moryx.Products.Management\Moryx.Products.Management.csproj" />
```
1nf0rmagician: clean up references, should not need all of them i gues

src/Moryx.Startup.Hooks/OrdersHook.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.Extensions.Logging;
+using Moryx.Configuration;
+using Moryx.Orders;
+using Moryx.Runtime.Modules;
+using Moryx.Runtime.Modules.Hooks;
+using Moryx.Tools;
+
+namespace Moryx.Startup.Hooks;
+
+public class OrdersHook(IModuleManager moduleManager, ILogger<OrdersHook> logger, IConfigManager configuration)
```
1nf0rmagician: comments

1nf0rmagician: sealed by default? also for the other sample hooks?

src/Moryx.Startup.Hooks/OrdersHook.cs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    private static OperationCreationContext CreateOperationsContext(OrdersHookConfig.ImporterConfig operation)
+    {
+        return new()
+        {
+            Number = operation.Number,
+            Order = new OrderCreationContext()
+            {
+                Number = operation.OrderNumber ?? $"30{operation.ProductIdentifier}",
```
1nf0rmagician: The 30 as a prefix seems random xD

src/Moryx.Startup.Hooks/OrdersHookConfig.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        /// <summary>
+        /// Allows disabling this config entry
+        /// </summary>
+        public bool Disabled { get; set; }
+
+        /// <summary>
+        /// Only run this hook, when the database contains no operations 
+        /// </summary>
+        public bool OnlyOnFreshDb { get; set; }
```
1nf0rmagician:
```suggestion
        public bool OnlyOnEmptyDb { get; set; }
```

src/Moryx.Orders.Hooks/OrdersHookConfig.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// Maps to OrderCreationContext.OrderType
+        /// </summary>
+        [DefaultValue("default")]
+        public string? OrderType { get; set; }
+
+        /// <summary>
+        /// Identifier of the product to produce
+        /// </summary>
+        public string? ProductIdentifier { get; set; }
+        /// <summary>
```
1nf0rmagician: lines

1nf0rmagician (changes requested):

docs/articles/framework/startup-hooks.md
andreniggemann: I believe the Resource.Samples aren't actually used anywhere, right? They are just for demonstration.
The hooks in Startup.Hooks are actually supposed to be used beyond a Demo app.

I am open to a discussion about naming. On thing I considered was Runtime.Hooks, because the Core Definitions have been moved to the Runtime Project as well.

docs/articles/framework/lifecycle-hooks.md
andreniggemann: I did not mention it here, because it's optional and kind of hard to explain, but there is a Recipe preselection Parameter that maps to the recipe preselection of the OperationCreationContext.

With that the RecipeAssignment is free to choose the Recipe, but in most cases it would be the default recipe.

docs/articles/framework/startup-hooks.md
andreniggemann: I think the core idea already explained at the top so most of it would be duplicated right there.
I think I would rather remove it completely than move it to the top.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: I was not sure how to approach this. What would you do if someone called this and single Database was not able to be created?
With a single boolean, the only thing we can realistically do is either ignore it or to stop application startup completely, but aborting basically never feels like the way to go in MORYX app, because you need to have a running app to correct the config in the CommandcCenter.

With my comment I wanted to express that this is a "best effort" approach.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: My understanding is that a migration is applied to an existing database to migrate the existing data to the new format

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: I had another approach in mind when initially setting this up.
You are right that this could be easily refactored to a LINQ statement in it's current form.

src/Moryx.Runtime.Kernel/HookExtension.cs
andreniggemann: Thanks for the heads up. I think I have some other places in my code where I can remove a filter now.

I mean given name it's relatively obvious, that it filters out things that are not classes, but it was not at all obvious to me that it filters out abstract classes by default

src/Moryx.Runtime.Kernel/HookExtension.cs
andreniggemann: To make it repeatable/predictable is a good idea, but I didn't really want people to rely on the existence of this rule, if you know what I mean.
I am going to add it anyway.

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
andreniggemann: Would you prefer InitializationResult?

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
andreniggemann: Task.Run has an overload that takes Func<Task> so no async void here.

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
andreniggemann: There is one case and that's when you want to do something before the event registration.
For example you might want to register a hook that documents each module state change in an external service. You could create the connection to that service and than run base.RunAsync.

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
andreniggemann: None of the hooks currently support cancellation in any form.
For general hooks that run on startup it did not make sense to me to create a general cancellation mechanism, because in my mind they should all run to completion to put the app in it's expected environment. Waiting for external services  to launch, creating databases, etc. which could take an unpredictable amount of time.

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
andreniggemann: Interesting, I did not know about that extension. Maybe you know, but we have an internal library that contains a similar abstractions and also one based on the RAII that would handle this very neatly using a using statement.
Maybe we could upstream that functionality as well.

In this case I think I need it like this to keep track of the _firstStart, that should be set to false in all cases.

src/Moryx.Startup.Hooks/OrdersHook.cs
andreniggemann: I guess we could seal them. There is no real reason to inherit from them

src/Moryx.Startup.Hooks/OrdersHook.cs
andreniggemann: That's because it is xD. I did not want to just use the productIdentifier and it just so happened that I saw a bunch of real orders starting with 30 something and concatenating 30 made might fake orders feel more realistic, LMAO.

It felt to insignificant to make it configurable separately as everyone who cares can just configure the ordernumber directly. The correct thing to would probably to be to remove this hack entirely, right?

docs/articles/framework/startup-hooks.md
1nf0rmagician: `Moryx.Runtime.Hooks` sounds good to me to 🙂

src/Moryx.Model/DbContextContainerExtension.cs
1nf0rmagician: I definitely saw other once and did not comment on every one

src/Moryx.Runtime/Modules/Hooks/ModuleHook.cs
1nf0rmagician: I was aware of that, I as wondering what the compiler makes of the function but apparently it is smart and creates `Func<Task`

https://stackoverflow.com/a/20395969/14014749

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
1nf0rmagician: > Maybe you know, but we have an internal library that contains a similar abstractions and also one based on the RAII that would handle this very neatly using a using statement.

Didn't know that, we should mark that internal as obsolete, do you know where it is?

>In this case I think I need it like this to keep track of the _firstStart, that should be set to false in all cases.

Right I missed that the variable is probabyl for use by base classes and needs to be set within the semaphore after the method call 🙈

src/Moryx.Runtime/Modules/Hooks/ModuleStartHook.cs
1nf0rmagician: > Waiting for external services to launch, creating databases, etc. which could take an unpredictable amount of time

That is basically why I would add it to the base method 😅 You don't necessarily need it in your samples I didn't want to argue about that, but I feel like the base class should offer it if only for some API hygene. Users could use the following code if they expect their start-up hooks to be meaningfully cancelable 🙂

```csharp
var builder = Host.CreateApplicationBuilder(args);
var app = builder.Build();

var lifetime = app.Services.GetRequiredService<IHostApplicationLifetime>();

await MyAsyncMethod(lifetime.ApplicationStopping);
```

src/Moryx.Startup.Hooks/OrdersHook.cs
1nf0rmagician: Well alternatively you could replace it with "Order for {ProductType}" that would make it more obvious and understandable that this was generated automatically😅

Also talking from experience, in application development people might miss that they still have the mock orders configured when they deploy to production and having these orders being obviously fake might help with that 😁

docs/articles/framework/startup-hooks.md
andreniggemann: I renamed it to Moryx.Runtime.Hooks in the new commit

docs/articles/framework/startup-hooks.md
andreniggemann: Removed it

src/Moryx/Tools/ILifecycleHook.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0
+
+namespace Moryx.Runtime;
+
+public interface IStartupHook
+{
+    /// <summary>
+    /// Lower runs earlier
+    /// </summary>
+    int Priority { get; }
```
seveneleven, line 14: Should that be `Order`? As with priorities usually highest comes first which could be the expectation here without further explanation.

andreniggemann: For me Priority is clearer as Order and it matches the priority logic the c# [Priority Queue](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-10.0#remarks) uses as well as task scheduling in operating systems

seveneleven: Ok, sounds reasonable. Maybe it was only the comment, that got me here, as I read as *lower (priority) runs earlier*

src/Moryx.Runtime/IStartupHook.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Lower runs earlier
```
seveneleven:
```suggestion
    /// Lowest value gets executed first
```

src/Moryx.Runtime.Kernel/HookExtension.cs
```diff
@@ -0,0 +1,48 @@
...
+            return services;
+        }
+    }
+
+    extension(IServiceProvider provider)
+    {
+        /// <summary>
+        /// Runs all registered startup hooks
+        /// </summary>
+        public async Task RunHooksAsync(CancellationToken cancellationToken)
```
1nf0rmagician: Sorry, I meant to make this optional 🙈 i.e. `CancellationToken cancellationToken = default` here and in the interface and implementations

I didn't mean to force everyone to supply a token 😬

1nf0rmagician (changes requested):

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,185 @@ namespace Moryx.Model;
...
-}
\ No newline at end of file
+
+    extension(IServiceProvider serviceProvider)
+    {
+        /// <summary>
+        /// Tries to create the database for each context.
+        /// Applies missing migrations as well.
+        /// </summary>
+        public async Task CreateAllConfiguredDatabasesAsync(ILogger logger = null, CancellationToken token = default)
```
dbeuchler: naming convention in the whole framework is CancellationToken (and if overriden in a delegate, then cancelToken). I would like to keep that.

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,185 @@ namespace Moryx.Model;
...
+        /// Deletes the database for the given context type
+        /// </summary>
+        /// <param name="context">Type of the context. Used to find the db to delete</param>
+        /// <param name="logger">optional logger</param>
+        /// <param name="token">Token to cancel the operation</param>
+        /// <returns></returns>
+        public async Task DeleteDatabaseAsync(Type context, ILogger logger, CancellationToken token = default)
+        {
+
+            await DeleteDatabaseInternalAsync(dbContextManager, context, logger, token);
```
dbeuchler: remove async/await and return task diretly, format

src/Moryx.Model/DbContextContainerExtension.cs
```diff
@@ -11,16 +13,185 @@ namespace Moryx.Model;
...
+            }
+        }
+
+        /// <summary>
+        /// Deletes the database for the given context type
+        /// </summary>
+        /// <param name="context">Type of the context. Used to find the db to delete</param>
+        /// <param name="logger">optional logger</param>
+        /// <param name="token">Token to cancel the operation</param>
+        /// <returns></returns>
```
dbeuchler: For all methods: fill or remove

src/Moryx.Runtime.Hooks/Moryx.Runtime.Hooks.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration" />
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Moryx.Model\Moryx.Model.csproj" />
+    <ProjectReference Include="..\Moryx\Moryx.csproj" />
+    <ProjectReference Include="..\Moryx.Runtime\Moryx.Runtime.csproj" />
+    <ProjectReference Include="..\Moryx.Orders\Moryx.Orders.csproj" />
```
dbeuchler: I still do not like the reference to orders. Do not mix core framework and higher dependencies. So Moryx.Runtime should never have a dependency to orders. My vote here: Remove this OrdersHook, I think there is not a real benefit for it, every application is creating its order somehow different. You could move that to a sample in the docs.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: Fixed it. I just checked there are a couple of other places that don't follow that convention so we should keep that in mind for MORYX 12 to rename those.
Also is that rule stated anywhere? Otherwise we might want to write these down.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: Removed. Should I create another PR to remove those everywhere with a simple search and replace?
According to my IDE there 140 other instances of this in the repo

src/Moryx.Model/DbContextContainerExtension.cs
dbeuchler: couple of other places? Fixed them all before release :-/

src/Moryx.Runtime.Hooks/Moryx.Runtime.Hooks.csproj
andreniggemann: I think it is quite useful, because all the differentiating logic i am aware of, on how orders are created are in the different assignment steps. This extension allows you to pass basically all relevant parameters to create an Order that MORYX allows you to specify.

I would constantly copy paste the documentation implementation and the version in the documentation runs the risk of becoming out of date, because it is not compiled.

Regarding the dependency: I guess this just comes down to naming, right?
The base class for the hooks are in the Runtime, but these Implementations are completely optional addons to that so in my mind they should not be treated the same way as the runtime project itself.

I could see adding it to a separate project, but I have the feeling that we have enough miniscule projects as it is.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: Image

src/Moryx.Runtime.Hooks/Moryx.Runtime.Hooks.csproj
dbeuchler: You can rotate the dependency and add the Moryx.Runtime.Hooks project to Moryx.Orders.Management and move the hook to the module.

andreniggemann: Will do that, because I don't see a better option. I just want to say that this feels unclean, because it couples the functionality that only requires the Facade right now tightly to the concrete module, making it harder to swap out the implementation.

src/Moryx.Model/DbContextContainerExtension.cs
dbeuchler: The local functions can be fixed to be named cancelToken, cancellationToken is not possible because it would override the method definition (endpoints). Regarding Mqtt.. I dont know since when they are only called token...

src/Moryx.Runtime.Hooks/Moryx.Runtime.Hooks.csproj
dbeuchler: Thats correct and regarding separation of concerns we must introduce a new project containing this hook. Moryx.Orders.StartupHooks? This would also support not referencing it in applications which are not using it.

andreniggemann: Would you consider Orders to be part of the controlsystem?
If so, we might call it Moryx.ControlSystem.[Startup]Hooks. That would allow us to at least integrate hooks we might need in the future, without having to create a project for each one.
Alternatively we might use Moryx.Factory.[Startup]Hooks given that the old Moryx-Factory umbrella used to contain the ControlSystem and Order definition, if I am not mistaken.

dbeuchler: No Orders is not Part of the ControlSystem. ControlSystem is everything regarding process-engine. Factory was Factory and not Orders (it was only part of the same repo). Make it clean - it is separate, it is optional -> new project. I dont want to introduce transitive references for stuff which is not used in applications.

src/Moryx.Model/DbContextContainerExtension.cs
andreniggemann: A few of them might be my fault. I am not sure. But I was unaware of the rule and it was not caught in review.
At least on my side the IDE always suggests just token and I just used that suggestion

dbeuchler: donet is using cancellationToken is the most APIs and therefore I harmonized that for MORYX 10 release. Some are introduced again... around the endpoints is my fault too. PR is comming.

dbeuchler: cleaned here: https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1304

1nf0rmagician (dismissed):

src/Moryx.Runtime.Hooks/Moryx.Runtime.Hooks.csproj
andreniggemann: Done

dbeuchler (changes requested):
There are some code issues which I have already solved (easier than commenting each line) in #1306

dbeuchler (dismissed):

PR #1206, opened 2026-05-04
Fix: Collection changed is not called when Resource References are changed through the UI

### LINKED ISSUE #524: Resource references are not raising CollectionChanged event
**Describe the bug**
Resource references are not raising `CollectionChanged` event

**To Reproduce**
````cs
public IReferences<IMyResource> MyResources { get; set; }

protected override void OnStart()
{
    var refCol = (IReferenceCollection)MyResources;
    refCol.CollectionChanged += RefColOnCollectionChanged;
}

private void RefColOnCollectionChanged(object sender, ReferenceCollectionChangedEventArgs e)
{
    // not raised
}
````
Add resource to the references and save. Nothing happens.

**Expected behavior**
CollectionChanged event should be raised.

### PR DESCRIPTION
**Pull request by andreniggemann**
### Summary

Collection changed events on collections of resources are not called when manipulating the resource graph.
See the linked issue

NOTE: To see only the logic changes, view the first commit. The second one applies linting rules to the changed files which leads to a lot of noise in the changes.

### Linked Issues

Fixes #524 

### Checklist for Submitter

- [ ] I have added or updated tests as appropriate

### Review

**Typical tasks**

- [ ] *Tests* are extended

**Clean code**

### DIFF
--- a/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ModelToResourceConverter.cs
+++ b/src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ModelToResourceConverter.cs
@@ -8,6 +8,7 @@ namespace Moryx.AbstractionLayer.Resources.Endpoints;
 
 /// <summary>
 /// Converts ResourceModel to Resource
+/// TODO: Can this be removed? It does not appear to be used
 /// </summary>
 internal class ModelToResourceConverter
 {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ModelToResourceConverter.cs [hunk 1]
```diff
@@ -8,6 +8,7 @@ namespace Moryx.AbstractionLayer.Resources.Endpoints;
 
 /// <summary>
 /// Converts ResourceModel to Resource
+/// TODO: Can this be removed? It does not appear to be used
```
1nf0rmagician: Sure that it is not used in the controller? If it isn't please go ahead and remove it. I'm wondering where the conversion is happening though 🙈

dbeuchler (approved):

src/Moryx.Resources.Management/Resources/ResourceLinker.cs
```diff
@@ -429,20 +492,36 @@ public void RemoveLinking(IResource deletedInstance, IResource reference)
...
+                             select property).FirstOrDefault();
 
         // If the referenced resource does not define a back reference we don't have to do anything
         if (backReference == null)
+        {
             return;
+        }
 
         // Remove the reference from the property
+        // TODO: Should we really test for IEnumerable and then cast to other types?
```
1nf0rmagician, line 507: @andreniggemann, @dbeuchler could you clarify this question?

andreniggemann: Well we check if  backReference.PropertyType is assignable to IEnumerable and if it is, we cast it to IReferenceCollection on the next line.

I think this approach is a little strange and would prefer checking directly if  backReference.PropertyType is IReferenceCollection, so that the cast does not fail unexpectedly.

In practice it does not fail, because Resources are always kept in a ReferenceCollection

PR #1210, opened 2026-05-05
Improve content of the projects documentation

1nf0rmagician:
### Summary

- Improve architecture diagrams with more information
- Replace png with mermaid diagrams
- Move reverse proxy instructions to tutorial section
- Extens tutorials toc

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,9 @@ To use MORYX to _build your own application_, we recommend to
 
 To quickly get a running MORYX application to _check out the different components_, we offer the [MORYX Demo](https://github.com/PHOENIXCONTACT/MORYX-Demo), a fully functional, simulated production system based on MORYX.
 
+And lastly, to _participate in the development_ of MORYX you can either submit a general feature request as described in our introduction to [Feature Specifications](./docs/requirements/index.md), or contribute your changes directly.
+Please refer to our [guidelines section](https://github.com/PHOENIXCONTACT/MORYX-Home#guidelines) to allow a smooth integration of your proposals.
+
 ## The Ecosystem
 
 Here we list all available packages in the MORYX ecosystem separated into components linking to their documentation.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -125,15 +128,123 @@ Here we list all available packages in the MORYX ecosystem separated into compon
...
+      ModCfgC["Module Configuration"]
+      ModCtrlC["Module Controller<br>(Lifecycle + DI)"]
+      FacC["Facade C<br>(public API boundary)"]
+      CompC["Component(s) C<br>(internal implementation)"]
+
+      ModCtrlC --> ModCfgC
+      ModCtrlC --> FacC
+      ModCtrlC --> CompC
+      FacC --> CompC
+    end
```
andreniggemann: I think the layout may get nicer, when adding another subgraph for all modules, although it's likely impossible to get the perfect layout with mermaid

README.md
```diff
@@ -125,15 +128,123 @@ Here we list all available packages in the MORYX ecosystem separated into compon
...
+
+    %% Framework Core
+    subgraph Core["MORYX Core"]
+      direction RL
+      Runtime(["Runtime / Host Infrastructure"])
+      GDI(["Global DI Container<br>(app-wide composition)"])
```
andreniggemann: That's not part of MORYX anymore, right? We just use the Asp.Net Core provided global container

README.md
```diff
@@ -125,15 +128,123 @@ Here we list all available packages in the MORYX ecosystem separated into compon
...
+      ModCtrlC --> ModCfgC
+      ModCtrlC --> FacC
+      ModCtrlC --> CompC
+      FacC --> CompC
+    end
+
+    %% Framework Core
+    subgraph Core["MORYX Core"]
+      direction RL
+      Runtime(["Runtime / Host Infrastructure"])
```
andreniggemann: What is part of this runtime & infra? On the spot I can only think of the things listed separately and the commandcenter.
If it is just the commandcenter, I would replace this point, but perhaps you had something else in mind.

andreniggemann (commented):
Updating our documentation is definitely necessary, so thanks for approaching that topic.

I am a fan of replacing pictures with diagrams as code, but I am not sure mermaid is really suited to these large more complicated diagrams.
We would have to see if that works out.

dbeuchler (commented):
> I am a fan of replacing pictures with diagrams as code, but I am not sure mermaid is really suited to these large more complicated diagrams.

The textually described diagrams should not lose comprehensibility. We have put a lot of effort into drawing state charts that only became readable and understandable through manual drawing. The now changed diagrams are a good example of how readability is lost. Mermaid is really cool, however readability and customizability are often lacking. We can also continue to draw them manually and place a draw.io XML file alongside them so they can be edited.

Is there a possibility to have a better format for the mermaid diagrams? Maybe split then into separate ones? Even on a big screen this is not readable and the layout is irritating. 

image

Our images before were much better undastandable

image

README.md
1nf0rmagician: Right, that is what I wanted to refer to, tried to make it clearer.

README.md
1nf0rmagician: Na you are right, the runtime is actually the content below. I replaced it with Communication and Launcher, which I think should rather be considered there

README.md
1nf0rmagician: Damn, after quite some time experimenting I found this thread/answer 

https://github.com/mermaid-js/mermaid/issues/2509#issuecomment-3419365163

Apparently I followed many other people and encountered the most upvoted open issue of all time in mermaid... A fix wa done in march but not released yet. Maybe I'll just have to wait a bit longer 😅

1nf0rmagician:
I'm quite happy with the new version, aside from the fact that the mermaid-bug mentioned above still makes the detailed diagram hard to read. 

@dbeuchler, @andreniggemann as seen in the architecture guideline document we had moved the components and endpoints guidelines into the home repository. I'm unsure what exactly would be the best place for these documents but having them separated in two repositories seems not the best approach to me.

PR #1214, opened 2026-05-08
Fix/resource axis movement UI

kumarnitin17:
Update AxisMovement constructor comment to be framework.


1nf0rmagician:
Implemented in #1219

PR #1217, opened 2026-05-12
Load parent ProductInstance information from entities and provide information

1nf0rmagician:
### Summary

Allow user to get information about the parent product instance.

### Linked Issues

Ported from https://github.com/PHOENIXCONTACT/MORYX-AbstractionLayer/pull/233

### Checklist for Submitter

### Review

**Typical tasks**

- [ ] *Tests* are extended

**Clean code**


src/Moryx.AbstractionLayer/Products/IProductInstance.cs
```diff
@@ -27,4 +27,15 @@ public interface IProductInstance
...
+/// <summary>
+/// Extended product instance
+/// </summary>
+public interface IProductInstanceExtended : IProductInstance
+{
+    /// <summary>
+    /// The parent product instance, null if it has no parent.
+    /// </summary>
+    ProductInstance Parent { get; }
+}
```
dbeuchler, line 41: Do you think this is really necessary? I would say that in this case it is totally okay to just add it to the ProductInstance-type and add it to the API if required (do you know cases?). We have done it in the passt too (e.g. WorkplanStep.Name).

Normally you work on the implementation itself. We keept the interfaces for extensions and to add a typed hierachy on custom interfaces.

If you really want to keep it: Add TODO to merge it in MORYX 12

dbeuchler (commented):
Before we merge that: We should first think about the default behavior. Do we *always* want to fetch the whole product-instance hierachy if you request a child? Since it was never implemented, I think it is more an edge case.

We should maybe add facade methods with a `ProductInstanceQuery` to enable the parent?

dacky179:
What about this PR? @1nf0rmagician

dbeuchler:
Before we make this the standard behavior and drastically slow down the requests to a single instance, we should introduce a minimal ProductInstanceQuery here.

dacky179:
I think it will only slow down a bit vs getting a full instance tree.
You already got all children before, but you were not able to navigate to the parents. 
Finding the root will only cost time if you request the identity of a sub instance (and specially in this case its very nice to get the parent).

1nf0rmagician:
> What about this PR? @1nf0rmagician

It is still planned, not forgotten but unfortunately on the top of my big stack 🙈

PR #1224, opened 2026-05-18
fix/EntryConvert_ParseWithFallback

H4llJ0rdan:
We found the bug, that floating point properties of resources could not be edited, if the culture was changed to de-DE. Together with [PR](https://github.com/PHOENIXCONTACT/ngx-moryx-web/pull/42), this problem should be fixed. May be we should wait for the PR to adujust the version of the ngx-moryx-web package?

- Introduced ParseWithFallback for numeric parsing
  - Attempts parsing using the provided culture first
  - Falls back to InvariantCulture if parsing fails
 
- Replaced direct Parse calls for floating point types
  - Applied to float, double, and decimal

- Improved robustness of type conversion
  - Reduces failures caused by culture-specific number formats (e.g. , vs .)
  - Ensures better compatibility for internationalized inputs

### DIFF
--- a/src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
+++ b/src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0
 
 using System.Collections;
+using System.Globalization;
 
 namespace Moryx.Serialization;
 
@@ -17,6 +18,7 @@ public static partial class EntryConvert
     public static EntryValueType TransformType(Type propertyType)
     {
         var valueType = EntryValueType.String;
+
         if (propertyType == typeof(Byte))
         {
             valueType = EntryValueType.Byte;
@@ -73,6 +75,7 @@ public static partial class EntryConvert
         {
             valueType = EntryValueType.Class;
         }
+
         return valueType;
     }
 
@@ -90,7 +93,7 @@ public static partial class EntryConvert
         if (type == typeof(string))
         {
             result = value;
-        }
+        } 
         else if (type == typeof(Byte))
         {
             result = Byte.Parse(value, formatProvider);
@@ -98,7 +101,7 @@ public static partial class EntryConvert
         else if (type == typeof(Boolean))
         {
             result = Boolean.Parse(value);
-        }
+            }
         else if (type == typeof(Int16))
         {
             result = Int16.Parse(value, formatProvider);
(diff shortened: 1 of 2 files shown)


1nf0rmagician (changes requested):
I like the change! We don't have to wait on the referenced PR, even though the bug won't be fixed just by this. But as it is a pure enhancement without breaking existing behaviour that is fine. 

Only thing, I need to ask you to please check formatting once. With the editorconfig the IDE should tell you several spots to adjust 😬

src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
```diff
@@ -143,10 +159,25 @@ public static object ToObject(Type type, string value, IFormatProvider formatPro
...
+    IFormatProvider provider,
+    NumberStyles styles,
+    TryParseDelegate<T> parser,
+    string typeName)
+    {
+        if (parser(s, styles, provider, out var v)) return v;
+
+        if (parser(s, styles, CultureInfo.InvariantCulture, out v)) return v;
+
+        throw new FormatException($"Cannot parse '{s}' to {typeName}.");
```
dbeuchler: you use typeof(T).Name - for exceptions it is totally okay.

src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
```diff
@@ -143,10 +159,25 @@ public static object ToObject(Type type, string value, IFormatProvider formatPro
...
+    string s,
+    IFormatProvider provider,
+    NumberStyles styles,
+    TryParseDelegate<T> parser,
+    string typeName)
+    {
+        if (parser(s, styles, provider, out var v)) return v;
+
+        if (parser(s, styles, CultureInfo.InvariantCulture, out v)) return v;
+
```
dbeuchler: Could you check the format here? 
- Keep attributes in method definition (until line gets to long)
- Add intendation if in separate line
- Do not use singe-line ifs
- Maybe give `v` a better name-  something like `parsedValue`

src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
```diff
@@ -90,15 +93,15 @@ public static object ToObject(Type type, string value, IFormatProvider formatPro
...
+        } 
         else if (type == typeof(Byte))
         {
             result = Byte.Parse(value, formatProvider);
         }
         else if (type == typeof(Boolean))
         {
             result = Boolean.Parse(value);
-        }
+            }
```
dbeuchler: Check format here (spaces)

dbeuchler (commented):
- I think it i worth to add 1 or 2 tests for this case
- Maybe a better commit message (you write something about package-lock?)

src/Moryx/Serialization/EntryConvert/EntryConvertGenerator.cs
H4llJ0rdan: Now the method signature is in one line. (I would prefer it in multiple lines, if there are 4 parameters or more)
Maybe we should add 
max_line_length = 120
in the editorconfig?

dbeuchler: There is no generally supported option in .editorconfig as far as I know. But yes we can add this for editors which support that (separate pr, limit file type)

https://developercommunity.visualstudio.com/idea/567214/support-max-line-length-in-editorconfig-for-visual.html.

dbeuchler (approved):

1nf0rmagician (approved):
Love it!

PR #1226, opened 2026-05-21
Fix collection properties becoming null after instantiation

kumarnitin17:
Fixed an issue where collection properties were becoming null during ResourceGraph instantiation.
Added a post-instantiation initialization step to ensure all collections are properly initialized before usage.

### DIFF
--- a/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs
+++ b/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs
@@ -121,10 +121,38 @@ internal class ResourceManagementFacade : FacadeBase, IResourceManagement
         ValidateHealthState();
 
         var resource = ResourceGraph.Instantiate(resourceType.ResourceType());
-        await initializer(resource);
+        CollectionsInitialized(resource);
+        await initializer(resource); 
         await ResourceGraph.SaveAsync(resource, cancellationToken);
         return resource.Id;
     }
+
+
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs
1nf0rmagician, line 1: Please revert this file, as discussed these changes are not necessary

src/Moryx.Resources.Management/Resources/ResourceEntityAccessor.cs
```diff
@@ -67,11 +67,12 @@ public Resource Instantiate(IResourceTypeController typeController, IResourceGra
...
         Instance.Id = Id;
         Instance.Name = Name;
-        Instance.Description = Description;
 
         // Copy extended data from json, or simply use JSON to provide defaults
-        JsonConvert.PopulateObject(ExtensionData ?? "{}", Instance, JsonSettings.Minimal);
-
+        if (!string.IsNullOrEmpty(ExtensionData))
+        {
+            JsonConvert.PopulateObject(ExtensionData ?? "{}", Instance, JsonSettings.Minimal);
```
1nf0rmagician, line 74: You can remove the null-coalescing operator within the if

1nf0rmagician (changes requested):
As discuessed the solution in the `ResourceEntityAccessor` should be correct, but please extend the respective unit tests to also persist the behaviour for the future.

PR #1228, opened 2026-05-26
Fix the resource initializer ignores field initilizer.

kumarnitin17:
Fixed issue where collection properties became null during JSON populate. So added conditional check to handle initialization correctly and prevent null values and do the unit tests added to verify the behavior.

### DIFF
--- a/src/Tests/Moryx.Resources.Management.Tests/ResourceEntityAccessorTests.cs
+++ b/src/Tests/Moryx.Resources.Management.Tests/ResourceEntityAccessorTests.cs
@@ -11,6 +11,7 @@ using NUnit.Framework;
 using System.ComponentModel;
 using System.Threading.Tasks;
 using Moryx.Resources.Management.Model;
+using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


src/Tests/Moryx.Resources.Management.Tests/ResourceEntityAccessorTests.cs
```diff
@@ -34,7 +35,7 @@ public void Setup()
     }
 
     [Test(Description = "Calling Instantiate without an entity sets default value")]
-    public void InstantiateWithoutEntitySetsDefaults()
+    public void InstantiateWithoutExtensionDataDoesNotOverrideInitializeProperty()
```
1nf0rmagician: You have replaced instead of added a unit test here, please correct this. Afterwards we can merge this PR 🙂

1nf0rmagician (changes requested):

1nf0rmagician (approved):

PR #1238, opened 2026-06-03
Fix several issues in the factory monitor module

1nf0rmagician:
### Summary

This PR combines several changes that are split into separate coherent commits. I recommend reviewing them one by one.

[Fix state stream event types](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/369224224961670b59d941ca33cc56e6f3e038e8)  
Multiple events where combined under the same event type which did not match the actual objects anyway.
Also adds proper event listeners on UI side, events were not processed before as onMessage is not fired for typed messages.

[Fix image paths in dev environment](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/1d0586209a972dc5392af40676202e2842dd929f)  
Environment root url wasn't used for relative urls before.

[Simplify observables/signals usage in FactoryMonitor](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/9e971369f4d6eb72b8e73479161304b34ec43f39)

[Fix mapped order state enum](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/6805744f45df2c11c07e36bac12920d8d29de744)  
After the update to MORYX 10 we forgot about the mapping happening in the endpoint. As a result the endpoint send meaningless integers instead of the string-enum values

[Simplify backgorund handling in factory monitor](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/56390b810946a298ccae5ee8ee050457b417c72e)  
Before there was an unnecessary call to the initial state api, which never populates the background url, so it was always reset to a blank background. Depending on the timing this caused a bug in the UI.
We removed the API call and simplified the backround url handling.

[Fix missing assignements on duplicated properties](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/4c02acd4449bd55305e5414faad3437b683af033)  
While the duplicated properties in themselves cannot be explained the can also not be removed in this major.
Hence, we added the missing assignements and todos to verify the properties.

[Some clean-up in Factory Monitor UI components](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/1e193dca68d38233d0af8a75671b3ea2b104fa40)

[Resolve UI update issues in Factory Monitor](https://github.com/PHOENIXCONTACT/MORYX-Framework/commit/877092a34e8aa2278aaf2f6cefd7186f6f734311)  
The issues leading to stream updates not being processed were manifold.
- Remove unnecessary API call to get all cells
- Properly processed initial change models from factory state
- Finished incomplete application of signals in factory and cell components
- Added missing object recreations on updates

Also started to clean-up code, remove duplicates, unused imports, splitting of responsibilities, removal of unnecessary service references, etc.

[Fix unsynchronized machine locations](https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1238/commits/4e6fe8aaf492da1ff5f6a3ee6eaab83a97437b3f)  
The behavior before had mixed up several aspects. For cells the location is taken from the cached cell in the cell store service, which was never correctly updated and also not set on the observable with a new reference, so the update was not applied and not triggered by the signal infrastructure. For factory items the behaviour is different, as the location is retrieved on each navigation to a new factory board and not cached locally. Hence no update in the cache is necessary. Because of this, we however need to keep the drag event translation (instead of resetting it as done for the cell).

[Fix inconsistent usage of interfaces and base types](https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1238/commits/fc8b9df57d90139b9e03d360df7dd0d7fe8f66a5) 
In the factory monitor controller we should make use of the defining interfaces underneath the web module, however some functions only worked with the utility base classes which are not necessary to be used.

[Reorganize factory monitor initialization](https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1238/commits/cfe33b6801af74e6c7c2737f6fc253ee86331653)  
Before the initialization of the factory monitor ui app happened hidden in the cell-store service and from there spread into other services. We replaced this with central app initialization via an app initializer. This reduced coupling beteween the apps and fixed issues were reloading in sub factory boards lead to timing issues.

Also adds an error page when the initialization failed instead of the current blank page and removed a lot of unused code especially from the order store service

[Register new operations in factory monitor](https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1238/commits/a6c97ad9da42cadc99d11f534add31316b3e90cf)  
The order store service caches the known orders in the system, however it did not add new orders thus far.
Also the endpoint did send incomplete order models on the order started event missing the order color.

### Relevant logs and/or screenshots

**Recovered functionality:**
https://github.com/user-attachments/assets/9f26985a-62ab-429e-894c-4390157cfa8c

**Failed loading Screen:**
Snag_c14b37

### Breaking Changes

While technically the changes to the event types on the factory monitor stream break existing applications, we consider them to be bugs. They mixed different types on the same typed stream and used a misleading/wrong type name for the stream.

### Checklist for Submitter

- [ ] I have added or updated tests as appropriate

### Review

**Typical tasks**

- [ ] *Tests* are extended

**Clean code**

- [x] *All* unused references are removed
- [x] Clean code rules are respected with passion (naming, ...)
- [x] Avoid *copy and pasted* code snippets

### DIFF
--- a/src/Moryx.FactoryMonitor.Endpoints/Models/ResourceChangedModel.cs
+++ b/src/Moryx.FactoryMonitor.Endpoints/Models/ResourceChangedModel.cs
@@ -14,6 +14,7 @@ public class ResourceChangedModel : VisualizableItemModel
     [DataMember]
     public virtual string CellName { get; set; }
 
+    // ToDo: Verify why this is duplicated and not using the location of the visual item
     [DataMember]
     public virtual string CellIconName { get; set; }
 
@@ -23,6 +24,7 @@ public class ResourceChangedModel : VisualizableItemModel
     [DataMember]
     public long Id { get; set; }
 
+    // ToDo: Verify why this is duplicated and not using the location of the visual item
     [DataMember]
     public CellLocationModel CellLocation { get; set; }
 
--- a/src/Moryx.FactoryMonitor.Endpoints/Extensions/CellExtensions.cs
+++ b/src/Moryx.FactoryMonitor.Endpoints/Extensions/CellExtensions.cs
@@ -29,8 +29,10 @@ internal static class CellExtensions
                 .GetResources(cellFilter)
                 .FirstOrDefault(x => x.Machine?.Id == cell.Id);
             resourceChangedCellModel.Location = Converter.Converter.ToCellLocationModel(machineLocation);
+            resourceChangedCellModel.CellLocation = resourceChangedCellModel.Location;
             resourceChangedCellModel.CellImageURL = machineLocation.Image;
             resourceChangedCellModel.IconName = machineLocation.SpecificIcon;
+            resourceChangedCellModel.CellIconName = resourceChangedCellModel.IconName;
             resourceChangedCellModel.FactoryId = GetFactoryId(cell, resourceManager);
 
             return resourceChangedCellModel;
--- a/src/Moryx.FactoryMonitor.Endpoints/Extensions/FactoryMonitorHelper.cs
+++ b/src/Moryx.FactoryMonitor.Endpoints/Extensions/FactoryMonitorHelper.cs
@@ -18,11 +18,10 @@ internal static class FactoryMonitorHelper
     private const string Activity_Event_Type_Key = "activityChangedModel";
     private const string Recource_Event_Type_Key = "resourceChangedModel";
 
-    public static void OrderStarted(OperationStartedEventArgs orderEventArg, Action<string, object> broadcast)
+    public static void OrderStarted(OperationStartedEventArgs orderEventArg, List<OrderModel> models, Action<string, object> broadcast)
     {
-        var orderModel = Converter.Converter.ToOrderModel(orderEventArg.Operation);
+        var orderModel = models.Single(o => o.Order == orderEventArg.Operation.Order.Number && o.Operation == orderEventArg.Operation.Number);
         broadcast(Order_Event_Type_Key, orderModel);
-
     }
 
     public static void OrderUpdated(OperationChangedEventArgs orderEventArg, Action<string, object> broadcast)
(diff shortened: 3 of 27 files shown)


src/Moryx.FactoryMonitor.Endpoints/Models/ResourceChangedModel.cs
```diff
@@ -14,6 +14,7 @@ public class ResourceChangedModel : VisualizableItemModel
     [DataMember]
     public virtual string CellName { get; set; }
 
+    // ToDo: Verify why this is duplicated and not using the location of the visual item
```
MathoMathiasCamara, line 17: If my memory serves me right this has something to do with the fact that we had 2 types  `Location` and `MachineLocation` i don't remember anymore what exactly in MORYX 6-8. But it was supposed to be simplified in MORYX 10 i don't know if that was done.

MathoMathiasCamara (commented):
@1nf0rmagician  I like it so far. If i remember Paul made some changes in MORYX 8 that were supposed to be cleaned. Do you know if those clean up were done ?

1nf0rmagician:
> @1nf0rmagician I like it so far. If i remember Paul made some changes in MORYX 8 that were supposed to be cleaned. Do you know if those clean up were done ?

If clean-up was done it was incomplete at best 🙈 I did some now, but after one week I want to lock in what I have achieved the next clean-up/refactoring is already prepared with the ToDo comments in several places.

src/Moryx.FactoryMonitor.Endpoints/Models/ResourceChangedModel.cs
1nf0rmagician: Anyway, they are both populated now, which also makes the UI work again 🙈

MathoMathiasCamara (approved):
Looks nice!

PR #1239, opened 2026-06-03
Appreciate that a process holders parent may be null

seveneleven:
A missing parent should resolve to an empty list instead of null reference exception leading to items not being updated.

### DIFF
--- a/src/Moryx.ControlSystem.Processes.Endpoints/EventHandlers/ProcessHolderEventHandlers.cs
+++ b/src/Moryx.ControlSystem.Processes.Endpoints/EventHandlers/ProcessHolderEventHandlers.cs
@@ -62,9 +62,10 @@ internal class ProcessHolderEventHandlers
         else
         {
             var group = ProcessHolderMappers.ModelFromParent(resource);
-            var positions = resource.Parent.Children
+            var positions = resource.Parent?.Children
                 .OfType<IProcessHolderPosition>()
-                .Select((p, index) => p.ToDto(index, group.Id));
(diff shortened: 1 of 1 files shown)


src/Moryx.ControlSystem.Processes.Endpoints/EventHandlers/ProcessHolderEventHandlers.cs
```diff
@@ -62,9 +62,10 @@ public static void SendUpdate(Action<ProcessHolderGroupModel> broadcast, IProces
...
         {
             var group = ProcessHolderMappers.ModelFromParent(resource);
-            var positions = resource.Parent.Children
+            var positions = resource.Parent?.Children
                 .OfType<IProcessHolderPosition>()
-                .Select((p, index) => p.ToDto(index, group.Id));
+                .Select((p, index) => p.ToDto(index, group.Id))
+                ?? [];
             group.Positions.AddRange(positions);
             broadcast(group);
```
1nf0rmagician, line 70: If I'm not mistaken there is another bug in here, don't we need to send the update for the process holder position itself, even if it has no parent (process holder group)? Or is this update handled somewhere else where i didnt see it and this is really only for the parents 🤔

seveneleven: What I fixed here is, that the UI didn't show a single thing when positions without parents existed. Apart from that, I couldn't observe any malfunction except the UI didn't update immediately sometimes which I would tackle separately.

1nf0rmagician (approved):

PR #1253, opened 2026-06-08
Update the driver after real world experience

seveneleven:
The changes in this commit is the result of observations made within real world applications. It substantially addresses bugs and performance issues. Unfortunately the change set got quite big because many changes have been made in parallel or caused each other.

The two biggest changes are:

* The async API has already been introduced with MORYX 10 and got implemented properly
* NodeBrowsing got abandoned: It took too much time to read all nodes on the OPC UA server (depending on servers ~40s to serveral minutes). While most of the time, only a small subset is used in an application, there was no reason to browse them all (and keep them in memory) compared to lazy load them when needed.

> Sorry for this being quite big. Please take into account, that this is not even ready yet and I also left some 'TODOs' here and there, where I did not have the time to further investigate. 
> I'd like this PR to get a review because these changes have been around for some time now and I also want to gather opinions on some of the decisions being made so far (there is a breaking change for example which I considered fine as the driver should be beta).


src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
```diff
@@ -135,19 +128,14 @@ public OpcUaNode(IOpcUaDriver driver, IModuleLogger logger, ExpandedNodeId nodeI
     }
 
     /// <inheritdoc />
-    public OpcUaNode(IOpcUaDriver driver, IModuleLogger logger, string identifier)
+    public OpcUaNode(IOpcUaDriver driver, IModuleLogger logger, string identifier, NodeClass nodeClass)
```
dbeuchler: breaking change

src/Moryx.Drivers.OpcUa/IOpcUaDriver.cs
```diff
@@ -16,9 +17,9 @@ public interface IOpcUaDriver : IMessageDriver, IInOutDriver
     /// <summary>
     /// Subscribes to a variable node. Nothing happens, if the node is not a variable
     /// </summary>
-    /// <param name="node">OpcUaNode to be subscribed</param>
+    /// <param name="nodeId">NodeId to subscribe to</param>
     /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is None.</param>
-    Task AddSubscriptionAsync(OpcUaNode node, CancellationToken cancellationToken = default);
+    Task AddSubscriptionAsync(string nodeId, CancellationToken cancellationToken = default);
```
dbeuchler: breaking change

src/Moryx.Drivers.OpcUa/IOpcUaNodeReader.cs
```diff
@@ -0,0 +1,21 @@
...
+internal interface IOpcUaNodeReader
+{
+    /// <summary>
+    /// Reads node attributes from an OPC UA server
+    /// </summary>
+    /// <param name="nodeId"></param>
+    /// <param name="namespaceTable"></param>
+    /// <param name="session"></param>
+    /// <param name="cancellationToken"></param>
+    /// <returns></returns>
```
dbeuchler: fill or remove

dbeuchler (commented):
I know it's draft... but I have scrolled through the changes and found some breaking changes for `dev`. Just keep in mind.

src/Moryx.ControlSystem.Processes.Endpoints/Extensions/ResourceExtensions.cs
```diff
@@ -9,6 +9,10 @@ internal static class ResourceExtensions
     public static Category ParentCategory(this Resource resource)
     {
         var parentType = resource.Parent?.GetType();
+        if (parentType == null)
```
1nf0rmagician: Looks like a fix related to #1239, did it slip into this PR?

src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
```diff
@@ -62,7 +62,7 @@ public class OpcUaNode : IMessageChannel
     /// <summary>
     /// List of all Subnodes. This property is null, when the node is no object node
```
1nf0rmagician: Comment is a missmatch?

src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
1nf0rmagician: Should be easy to make minor conform, just add a commit after this PR is ready which can then be easily reverted on future 🙂

src/Moryx.Drivers.OpcUa/States/BrowsingNodesOnServerState.cs
```diff
@@ -1,21 +0,0 @@
-// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
-// Licensed under the Apache License, Version 2.0
-
-using Moryx.AbstractionLayer.Drivers;
-
-namespace Moryx.Drivers.OpcUa.States;
-
-internal class BrowsingNodesOnServerState(OpcUaDriver context, StateMachines.StateBase.StateMap stateMap) : DriverOpcUaState(context, stateMap, StateClassification.Initializing)
```
1nf0rmagician, line 8: While I agrre 40s waiting time during a normal state machine is way to long, what do you think of keeping the feature as an option requiring an external trigger, e.g. an entry method? That would mean keeping the state as well, but it would not be part of automatic state transitions. 
Since I don't know what the node browsing is actually doing, I can not determine the value of such a feature

src/Moryx.Drivers.OpcUa/States/InitializingSubscriptionsState.cs
```diff
@@ -2,33 +2,34 @@
...
       : DriverOpcUaState(context, stateMap, StateClassification.Initializing)
 {
-    internal override Task OnSubscriptionsInitializedAsync()
+    public override async Task OnEnterAsync(CancellationToken cancellationToken)
     {
-        NextState(StateRunning);
-        Context.ReadDeviceSet();
-        return Task.CompletedTask;
+        await base.OnEnterAsync(cancellationToken);
+        Context.RemoveSubscription();
```
1nf0rmagician, line 15: Haven't we done this on the disconnect?

src/Moryx.Drivers.OpcUa/States/InitializingSubscriptionsState.cs
```diff
@@ -2,33 +2,34 @@
...
-    internal override OpcUaNode GetNode(string identifier)
+    internal override async Task OnSubscriptionsInitializedAsync(CancellationToken cancellationToken)
     {
-        return Context.GetNodeAsync(identifier).GetAwaiter().GetResult();
+        await NextStateAsync(StateRunning, cancellationToken);
+        await Context.ReadDeviceSetAsync(cancellationToken);
     }
 
-    internal override void AddSubscription(OpcUaNode node)
+    internal override OpcUaNode GetNode(string identifier)
```
1nf0rmagician, line 25: Any reason you didn't make this async as well?

src/Moryx.Drivers.OpcUa/States/RunningState.cs
```diff
@@ -2,16 +2,16 @@
...
 
 internal class RunningState(OpcUaDriver context, StateMachines.StateBase.StateMap stateMap) : DriverOpcUaState(context, stateMap, StateClassification.Running)
 {
-    internal override async Task RebrowseNodesAsync()
+    public override async Task OnEnterAsync(CancellationToken cancellationToken)
     {
-        Context.RemoveSubscription();
-        NextState(StateBrowsingNodes);
-        await Context.BrowseNodesAsync();
+        Context.PublishRunningState();
```
1nf0rmagician, line 13: Not sure if relevant at any point, but we publish the running state before we read the device set. Could this be an issue for subscribers of the running state publishing?

src/Moryx.Drivers.OpcUa/OpcUaNodeReader.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (node == null)
+            {
+                return null;
+            }
+            var result = ConvertToNode(node, namespaceTable);
+            return result;
+        }
+        catch
+        {
+            return null;
```
1nf0rmagician, line 62: Since you have a logger in the class anyway why not log the broadly catched exceptions here?

Also I assume you only need to out the sesion call into the try catch or am I missing any possible exceptions?

src/Tests/Moryx.Drivers.OpcUa.Tests/Moryx.Drivers.OpcUa.Tests.csproj
```diff
@@ -16,4 +16,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Moryx.Drivers.OpcUa\Moryx.Drivers.OpcUa.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
1nf0rmagician: ncessary?

1nf0rmagician (commented):
All test I saw removed concerned the browse feature, or have I overlooked anything that I should have paid closer attention to?

src/Moryx.Drivers.OpcUa/IOpcUaDriver.cs
seveneleven: > [...] and I also want to gather opinions on some of the decisions being made so far (there is a breaking change for example which I considered fine as the driver should be beta).

seveneleven:
> All test I saw removed concerned the browse feature, or have I overlooked anything that I should have paid closer attention to?

Yes, they were mainly related to that.

src/Moryx.Drivers.OpcUa/States/RunningState.cs
seveneleven: The device set is currently not read in the current state of the PR and is not required for any function except 'showing' it to users.
I only left it because I wanted to start the discussion, if we want to keep it at all.

src/Moryx.Drivers.OpcUa/States/InitializingSubscriptionsState.cs
seveneleven: Now that I browsed through some of the changes, I think, back then I decided to not update it because the name didn't fit the `*Async` convention (which is the same for `AddSubscription`). I'm open to change this, though.

src/Moryx.Drivers.OpcUa/IOpcUaDriver.cs
dbeuchler: > there is a breaking change for example which I considered fine as the driver should be beta

What is beta? The driver was released several times and we should not break the API

src/Moryx.Drivers.OpcUa/States/BrowsingNodesOnServerState.cs
seveneleven: Node browsing meant browsing all nodes that a server provides. Which resulted in

1. Long initialization time
2. Possibly keeping a huge graph in memory for no practical reason

So that was basically 'only' required for two reasons: 
1. Nodes had to 'exist' in order to subscribe to them
2. Bringing the tree to the UI (EntrySerialize) required the tree to be complete.

My thoughts on removing that were: 

* Subscribing to a single node is (and should be) possible without reading all other nodes in advance. This is a lot resources to be saved as in an application you probably only need a relatively small subset.
* Bringing the `Nodes` property to the UI with EntrySerialize was a neat idea, but not practical. As EntrySerialize is not aware of lazy loading trees, it required the whole tree. That made a limit in depth necessary, as the UI couldn't handle it otherwise, which made the 'feature' usesless for deeper nested nodes. Cyclic references were also something to be handled and navigating through nodes was rather complicated than helpful (this might bear some ideas to improve EntrySerialize).
If someone wants to investigate the node structure on an OPC UA server, a dedicated UI would be more helpful. Even though I would suggest to use an existing tool instead of reinventing the wheel here.

TLDR

Browsing (all) nodes was technically not necessary for what it was used for. In case of navigating nodes via UI, that feature wasn't practical which rather lead you to other tools that were built for that purpose.

src/Moryx.Drivers.OpcUa/States/InitializingSubscriptionsState.cs
seveneleven: Will investigate and test that

1nf0rmagician: Was the test successful? 😬

src/Moryx.Drivers.OpcUa/States/InitializingSubscriptionsState.cs
1nf0rmagician: For all internal methods this is not an issue. We have several of these cases and there is an ADR which states that only for public APIs it is required to have the Async Suffix 😊

src/Moryx.Drivers.OpcUa/IOpcUaNodeReader.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.Drivers.OpcUa.Nodes;
+using Opc.Ua;
+using Opc.Ua.Client;
+
+namespace Moryx.Drivers.OpcUa;
+
+internal interface IOpcUaNodeReader
```
1nf0rmagician, line 10: Maybe I forgot I suggested this myself, but why having an internal interface here? could just be removed if it is just internal stuff right? 😅

1nf0rmagician (commented):
And mind the merge conflict 😊

src/Moryx.Drivers.OpcUa/IOpcUaNodeReader.cs
seveneleven: Was introduced with improving the tests to loosen the driver (and test code) from the OPCUA Foundation implementation.

src/Tests/Moryx.Drivers.OpcUa.Tests/OpcUaTestBase.cs
```diff
@@ -18,11 +21,13 @@ public class OpcUaTestBase
...
     private const ushort IndexNamespace2 = 2;
 
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
     protected Mock<ISession> _sessionMock;
     protected Dictionary<NodeId, ReferenceDescription> _rootNodes;
     protected ReferenceDescription _root;
-    protected NamespaceTable _namespaceTable = CreateNamespaceTable();
     protected OpcUaDriver _driver;
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
+    protected NamespaceTable _namespaceTable = CreateNamespaceTable();
```
dbeuchler: Fields should not be protected; that exposes mutable state to subclasses without any encapsulation

src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
```diff
@@ -9,7 +9,7 @@
 using Opc.Ua;
 using Opc.Ua.Client;
 
-namespace Moryx.Drivers.OpcUa;
+namespace Moryx.Drivers.OpcUa.Nodes;
```
dbeuchler: Breaking change

src/Moryx.Drivers.OpcUa/Nodes/OpcUaNode.cs
```diff
@@ -120,33 +120,22 @@ private OpcUaNode(IOpcUaDriver driver, IModuleLogger logger)
 
     }
 
-    /// <inheritdoc />
-    public OpcUaNode(IOpcUaDriver driver, IModuleLogger logger, string namespaceUri, string nodeIdValue)
-        : this(driver, logger)
-    {
-        NodeId = new ExpandedNodeId(nodeIdValue, namespaceUri);
-    }
-
```
dbeuchler, line 129: Breaking change

dbeuchler (commented):
I cannot really review that because I have no realistic option to test that. Only static code analysis from me is possible. I still found some breaking changes. "the driver should be beta" is not the case. It's released under several MORYX release versions. If the driver is still beta, we should move it out of the Repo and publish beta releases, or do not publish releases of the package to nuget e.g. by manual releases published to myget. 

Btw
From my point of view explicit driver implementations should not be part of the repo and moved to another, but this discussion should not be part of this PR.

PR #1259, opened 2026-06-09
Allow integrating data into IConfigManager

### LINKED ISSUE #710: Allow optional integration of configuration providers in MORYX ConfigurationManager

andreniggemann:
### Summary

Allow populating ModuleConfigs from IConfiguration entries.
This is our way to allow injecting secrets and configuration from ConfigurationProviders based on Microsoft.Extensions.Configuration into the MORYX IConfigManager.

To properly handle nested configs (an important part of the IConfiguration concept the ValueProviderExecutor has been extended to allow access to context information through the new IContextAwareValueProvider interface.

### Linked Issues

CLoses #710 
It does not depend on the Password Attribute as mentionend in the Issue, because the Password Attribute still means that the value will be serialized to disk and over the API.

### Checklist for Submitter

### Review

**Typical tasks**

- [x] Merge request is well described
- [x] Critical sections are *documented in code*
- [x] *Tests* available or extended
- [x] *Documentation* is created / updated
- [x] Running in test environment

**Clean code**

- [x] *All* unused references are removed
- [x] Clean code rules are respected with passion (naming, ...)
- [x] Avoid *copy and pasted* code snippets


src/Moryx/Configuration/ValueProvider/IContextAwareValueProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the Apache License, Version 2.0
+
+using System.Reflection;
+
+namespace Moryx.Configuration;
+
+/// <summary>
+/// Stackframe to keep track of a recursive calls to the value provider to provide context.
+/// </summary>
+/// <param name="parent"></param>
```
1nf0rmagician: REmonde to fully populate

src/Moryx/Configuration/InjectedConfigAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+/// <code>
+/// [InjectedConfig(".:ApiKey")]
+/// </code>
+/// Load the actual config key from the `ConfigKey` property
+/// <code>
+/// [InjectedConfig(nameof(ConfigKey), fromProperty: true)]
+/// </code>
+/// </example>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Class)]
+public class InjectedConfigAttribute(string configKey, bool fromProperty = false) : Attribute
```
1nf0rmagician: Suggestion to change to `ProvidedConfigAttribute` similar to the IConfiguration**Provider** which is used to provide the config?

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/ConfigManagerTests.cs
```diff
@@ -139,9 +151,170 @@ public void ByName()
...
+            name: "SecretConfig",
+            getCopy: true);
+
+        // Assert
+
+        Assert.That(config.NormalSetting, Is.EqualTo("from-json"));
+        Assert.That(config.ServiceApiKey, Is.EqualTo("from-provider"));
+    }
+
+    [Test(Description = "Resolves an explicit configuration key placeholder from IConfiguration.")]
```
1nf0rmagician: copy paste

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/ConfigManagerTests.cs
```diff
@@ -139,9 +151,170 @@ public void ByName()
...
+        var writtenConfig = File.ReadAllText(configPath);
+
+        using var _ = Assert.EnterMultipleScope();
+        Assert.That(config.ServiceApiKey, Is.EqualTo("from-provider"));
+        Assert.That(writtenConfig, Does.Not.Contain("from-provider"));
+    }
+
+
+    [Test(Description = "Applies provider values to generated configs when no JSON file exists.")]
+    public void GeneratedConfig_ReceivesValuesFromConfiguration()
```
1nf0rmagician: MIxed naming conventions, either keep it with the existing tests or apply the .net guideline consistently for new tests

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/MsExtensionConfigProviderTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        // Act
+        var result = provider.Handle(stack);
+
+        // Assert
+        using var _ = Assert.EnterMultipleScope();
+        Assert.That(result, Is.EqualTo(ValueProviderResult.Handled));
+        Assert.That(testConfig.ServiceApiKey, Is.EqualTo(Expected));
+    }
+
+    private static IConfigurationRoot BuildConfig(Dictionary<string, string> configDict)
```
1nf0rmagician: could be used in other test class as well

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/MsExtensionConfigProviderTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        // Assert
+        using var _ = Assert.EnterMultipleScope();
+        Assert.That(result, Is.EqualTo(ValueProviderResult.Handled));
+        Assert.That(testConfig.ServiceApiKey, Is.EqualTo(Expected));
+    }
+
+    [TestCase(0)]
+    [TestCase(1)]
+    [TestCase(2)]
+    [TestCase(3)]
```
1nf0rmagician: I think 2 test cases are enough here 🤣

src/Moryx.Runtime.Kernel/Configuration/ConfigManager.cs
```diff
@@ -34,8 +38,9 @@ public ConfigManager()
     protected IValueProvider[] ValueProviders =>
     [
         _sharedProvider,
+        new ActivatorValueProvider(),
```
1nf0rmagician: I assume this change in the order of the collection is what you are refering to in the description, which would need a proper testing for its side effects?

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+    public ValueProviderResult Handle(Stack<ExecutorLevel> levels)
+    {
+        var (config, prop, dict) = levels.Peek();
+        if (_configuration == null || config == null || !prop.CanRead || !prop.CanWrite)
+        {
+            return ValueProviderResult.Skipped;
+        }
+        var currentValue = prop.GetValue(config);
+        var defaultValue = GetDefaultValue(prop.PropertyType);
+        if (!Equals(currentValue, defaultValue)) // value is already assigned
```
1nf0rmagician: For Meaning, the intention behind the usage of configuration providers with MORYX 10 is to initialize or to provide not externally visible config values. We do not store values back to any configuration provider and we skip any application changed properties.

Should we have a test case to document this as intended behaviour?

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            return ValueProviderResult.Skipped;
+        }
+
+
+        var configKey = GetConfigKeyNameCached(levels);
+
+        if (_configuration.GetSection(configKey) is ConfigurationSection section && section.Exists())
+        {
+            prop.SetValue(config, section.Get(prop.PropertyType));
```
1nf0rmagician: This would throw if the config contains a wrong datatype, correct? Should be catched with an understandable error message then (and tested)

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+            var baseKey = GetRootConfigKey(levels);
+            return $"{baseKey}:{key}";
+        }
+        
+    }
+
+    private static string KeyFromAttribute(object config, InjectedConfigAttribute attribute)
+    {
+        var configKey = attribute.ConfigKey;
+        if (attribute.FromProperty)
```
1nf0rmagician: You could throw nice exceptions if someone missconfigured the property name or the property is not a string property. I think currently the behaviour is rather undefined

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (!(property is null || property.PropertyType != typeof(string) || !property.CanRead))
+            {
+                configKey = property.GetValue(config) as string;
+            }
+        }
+
+        return configKey;
+    }
+
+    private object GetRootConfigKey(Stack<ExecutorLevel> levels)
```
1nf0rmagician: This should return string shouldn'T it?

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+        if (_configuration.GetSection(configKey) is ConfigurationSection section && section.Exists())
+        {
+            prop.SetValue(config, section.Get(prop.PropertyType));
+            return ValueProviderResult.Handled;
+        }
+
+        return ValueProviderResult.Skipped;
+    }
+
+    private string GetConfigKeyNameCached(Stack<ExecutorLevel> levels)
```
1nf0rmagician: I must admit I do not quite get the caching part in this, maybe a little explanation comment/example comment would help me🙈

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/MsExtensionConfigProviderTests.cs
1nf0rmagician, line 1: While you have implemented collection handling (as far as i can see) there is no test for this

1nf0rmagician (commented):
I'm happy with the concept and think it is a nicely limited change!

I already went quite into the details with the review, but please take this not as a being overly picky with the PR. I needed to read it closely for full understanding and when I spotted something I didn't want to forget about it later when you had brought it out of the draft state 😬🤗

dbeuchler (commented):
I like that implementation! Thanks for that. (just small things from my end: code format, braces, no one liners...).

One thing to think about: Why do you have touched the default `ConfigManager`? I think this can easily archived by derive from it and provide an additional implementation which the application developer can select if required. Keeps the default implementation as it is and for extended usage, it is possible to use it in application.

Before we merge it to dev and make an release instantly, could we apply this into an application (I don't mean the demo) to test if it behaves as wanted?

andreniggemann:
> I like that implementation! Thanks for that. (just small things from my end: code format, braces, no one liners...).

Thank you very much :)

> One thing to think about: Why do you have touched the default `ConfigManager`? I think this can easily archived by derive from it and provide an additional implementation which the application developer can select if required. Keeps the default implementation as it is and for extended usage, it is possible to use it in application.

That's good idea. I would add a switch to the AddMoryxKernel Extension method that allows opting in to the new behavior, if that's ok with you

> Before we merge it to dev and make an release instantly, could we apply this into an application (I don't mean the demo) to test if it behaves as wanted?

That's my plan. That's why it's still in draft. Additionally I intend to add a regression test that takes the ModuleConfigs of the core Modules and checks that, if the IConfiguration is empty, you get the exact same configs.

dbeuchler:
> That's good idea. I would add a switch to the AddMoryxKernel Extension method that allows opting in to the new behavior, if that's ok with you

Instead of adding options to the extensions, another idea: You can replace registered services with `services.Replace`? Maybe a simple hint in the docs and this behavior is fully optional.

andreniggemann:
Isn't it completely optional if I define the Method as `AddMoryxKernel(bool configurationIntegration = false)`?

Nobody would have to change their code, it's much more discoverable as a hint in the docs and the "new ConfigManager" (not sure on naming yet, or if even needs a new type. Maybe it could just be a different constructor that takes a list of IValueProviders) can be internal. 

By the way: Is there a reason for the existing ConfigManager being public? In an application we always use it through the IConfigManager interface

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MsExtensionsConfigProvider.cs
andreniggemann: It only stores the config key for an element it has already seen in the dict attached to the stackframe.
This means that if we have nested configuration, you don't have to build the entire path from scratch each time.

docs/articles/framework/configuration.md
```diff
@@ -64,11 +64,137 @@ In the configuration editor each entry is represented by its property name and v
...
+### Example
+The ModuleConfig of the ProductManagement module contains a property called MaxImporterWait.
+To configure this value via IConfiguration, define the following key:
+```
+Moryx.Products.Management.ModuleConfig:MaxImporterWait
+```
+This can be set in appsettings.json like this:
+
+
+You can archive this by adding the following to your appsettings.json
```
1nf0rmagician: little duplicate

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/ConfigManagerWithConfigProviderTests.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        var provider = serviceCollection.BuildServiceProvider();
+
+        var manager = provider.GetRequiredService<ConfigManager>();
+        manager.ConfigDirectory = _tempDir;
+
+        return manager;
+    }
+
+    // TODO: Consider removing -> Tested by MicrsoftExtensionConfigProviderTests
```
1nf0rmagician: Open todos in this file, if you ask me, I like keeping some integration tests (if they actually test just a littel bit more than the `MicrsoftExtensionConfigProviderTests`.

1nf0rmagician (dismissed):

src/Tests/Moryx.Runtime.Kernel.Tests/Configuration/ConfigManagerWithConfigProviderTests.cs
andreniggemann: Ok, than I'll remove the TODOs. They test the integrate a little bit

src/Moryx.Runtime.Kernel/KernelServiceCollectionExtensions.cs
```diff
@@ -20,10 +21,27 @@ public static class KernelServiceCollectionExtensions
...
+                {
+                    var manager = ActivatorUtilities.CreateInstance<ConfigManager>(serviceProvider);
+                    return manager.ConfigureValueProviders(original => [
+                        manager._sharedProvider,
+                        new ActivatorValueProvider(),
+                        ActivatorUtilities.CreateInstance<MicrosoftExtensionsConfigProvider>(serviceProvider),
+                        new DefaultValueAttributeProvider(),
+                    ]);
+                });
+            }
```
dbeuchler, line 40: Instead of define that again here, couldnt it be also an option of the config manager which you set here? The extension knows to much about the internals of the config manager just for the flag.

andreniggemann: Note for everyone else: @dbeuchler and I discussed this on the phone.

Because of the cyclic dependency between the SharedConfigProvider and the ConfigManager and their difference in visibility, it's not easily possible to add a subclass of the ConfigManager that does all this configuration, so we are going to keep it as is.

On the Future Branch this can be resolved by just removing the SharedConfigProvider as it is basically unused.

dbeuchler (approved):

PR #1261, opened 2026-06-10
Fix generation of the `SimpleGraph` in the factory monitor

1nf0rmagician:
Currently the `SimpleGraph` and the factory state generation have a duplicate implementation of the logic to scan the resource graph.
After the logic was fixed for the state models in #1238 we now follow up on the generation of the `SimpleGraph`.

With this PR we prioratize the machine property of the locations over scanning the resource children. Also we add robustness to custom implementations of the interfaces in which a resource can implement both `IMachineLocation` and `ICell`, in these cases the graph now adds the location first and then either the target cell or a cell underneath the location.

Also this commit cleans up a lot of code, e.g. duplicate generation of the simple graph, unnecessary calls to the unsafe resource facade apis, unused code, unnecessary restrictive extensions, unnecessary resource fetching of already retrieved resources...

### DIFF
--- a/src/Moryx.FactoryMonitor.Endpoints/FactoryMonitorController.cs
+++ b/src/Moryx.FactoryMonitor.Endpoints/FactoryMonitorController.cs
@@ -49,7 +49,7 @@ public class FactoryMonitorController : ControllerBase
         Converters = { new JsonStringEnumConverter() }
     };
 
-    private List<ICell> _cells;
+    private Dictionary<IMachineLocation, ICell> _locationToCellMappings;
     private Timer _resourceChangedTimer;
     private readonly ILogger<FactoryMonitorController> _logger;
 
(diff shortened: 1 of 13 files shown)


src/Moryx.FactoryMonitor.Endpoints/Models/SimpleGraph.cs
```diff
@@ -47,60 +47,54 @@ public static SimpleGraph Create(Resource resource)
             Id = factory.Id,
             Type = nameof(IManufacturingFactory)
         };
-        resource.Children.ForEach(graph.Append);
+        resource.Children.ForEach(graph.AppendLocation);
         return graph;
     }
 
+    // ToDo: Use existing graph model
```
MathoMathiasCamara, line 54: Do you mean to replace VisualizableItemModel  with SimpleGraph ?

src/Moryx.FactoryMonitor.Endpoints/Models/SimpleGraph.cs
```diff
@@ -121,35 +115,89 @@ public SimpleGraph GetSubGraphById(long id)
...
+        // Add the node itself
+        var subGraph = AddSubGraphRoot(addition);
+
+        // Proceed through the resource tree
+        switch (addition)
         {
-            addition.Children.ForEach(subGraph.Append);
+            // Prefer machine property as target of the location before using children
+            // This unifies behaviour with the gathering of resource changed models in the controller
+            // ToDo: With MORYX 12 locations shhould hold a list of targets which should be used exclusively instead of children here
```
MathoMathiasCamara, line 164: Mmm we really need to reconsider the assumption that a ` Location can only have 1 Machine/Cell`. If you now add a Targets field what is that for ? is that replacing the `Machine` property?

src/Moryx.FactoryMonitor.Endpoints/FactoryMonitorController.cs
```diff
@@ -114,6 +112,42 @@ public ActionResult<FactoryStateModel> InitialFactoryState()
...
+        }
+        catch (Exception e)
+        {
+            _logger.LogError(e, "Could not retrieve order information in {name}", nameof(FactoryMonitorController));
+            // We want to fail gracefully, if e.g. the facade is not yet available
+            return [];
+        }
+    }
+
+    // ToDo: This is sub-optimal as IMachineLocation does not require a cell to be referenced (and shouldn't) but we
```
MathoMathiasCamara, line 145: That's something we really need to discuss. Because the MachineLocation has a Machine property which means "At a location there should be a Machine" but given that we need Layered feature, there was a need for a workaround.  But now (MORYX 12) we need to totally revisit the initial assumption.

src/Moryx.FactoryMonitor.Endpoints/FactoryMonitorController.cs
```diff
@@ -122,11 +156,12 @@ public ActionResult<FactoryStateModel> InitialFactoryState()
...
+        var locations = _resourceManager.GetResources<IMachineLocation>();
+        var locationToCellMappings = MapCellsTo(locations);
 
         var converter = new Converter.Converter(_serialization);
-        return cells.Select(x => new SimpleGraph { Id = x.Id }.ToVisualItemModel(_resourceManager, _logger, converter, CellFilterBaseOnLocation)).ToList();
+        return locationToCellMappings.Select(l2cMapping => new SimpleGraph { Id = l2cMapping.Key.Id }
+            .ToVisualItemModel(_resourceManager, _logger, converter)).ToList();
     }
 
     // ToDo: Endpoints usually return arrays and pagination would be good.
```
MathoMathiasCamara, line 167: This depends on what you are trying to do in the UI. The better question is `Should the UI dictates what the Endpoint returns ?`

MathoMathiasCamara (changes requested):
Good so far

MathoMathiasCamara (approved):

PR #1275, opened 2026-06-15
Add `IMaterialContainer`  resource interface with base classes.

1nf0rmagician:
Adds the resource interface defining the minimal API of a material container which is not linked to any other domain objects/concepts. Adds `MaterialContainer` as an abstract base class for material flow handling of a material container matching the feature specifications, i.e.
- adds material flow state machine to the container resource
- adds persistance and reload functionality to a material container
- adds material registration behaviour from the resource UI, resource initializers and other resources
- adds material request behaviour from the resource UI, resource initializers and other resources
- adds user friendly display attributes for relevant members
- adds container host resource relation for resource linking of containers
- add integration tests for these functionalities
- adds documentation for setup in MORYX applications


src/Moryx.Material/Events/FillingLevelChangedEventArgs.cs
```diff
@@ -1,35 +1,24 @@
...
+namespace Moryx.Material.Events;
 
 /// <summary>
 /// Event args for <see cref="IMaterialContainer.FillingLevelChanged"/>.
 /// </summary>
-public class FillingLevelChangedEventArgs : EventArgs
+/// <remarks>
+/// Creates a new instance of <see cref="FillingLevelChangedEventArgs"/>.
+/// </remarks>
+public class FillingLevelChangedEventArgs(IMaterialContainer container, decimal oldQuantity, decimal newQuantity) :
```
andreniggemann: I think, we should stick to doubles like in the rest of the framework. I think this still allows to address all ints up to 2^53.

src/Moryx.Material/Events/MaterialChangedEventArgs.cs
```diff
@@ -1,35 +1,24 @@
...
+namespace Moryx.Material.Events;
 
 /// <summary>
 /// Event args for <see cref="IMaterialContainer.MaterialChanged"/>.
 /// </summary>
-public class MaterialChangedEventArgs : EventArgs
+/// <remarks>
+/// Creates a new instance of <see cref="MaterialChangedEventArgs"/>.
+/// </remarks>
+public class MaterialChangedEventArgs(IMaterialContainer container, string? oldMaterial, string? newMaterial) :
```
andreniggemann: In my mind a MaterialChange always implies a filling level change, so it should include at least the new quantity or even inherit from FillingLevelChanged

src/Moryx.Material/States/InboundStateInformation.cs
```diff
@@ -0,0 +1,53 @@
...
+/// </summary>
+[DataContract]
+public class InboundStateInformation : StateInformation
+{
+    /// <summary>
+    /// Optional identifier of the announcement.
+    /// </summary>
+    [DataMember]
+    [Display(Name = "Announcement ID", Description = "Unique Identifier of the material announcement (optional).")]
+    public string? AnnouncementId { get; set; }
```
andreniggemann, line 20: Should these Ids be Identity objects?

src/Moryx.Material/States/InboundStateInformation.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// Optional expected arrival of the announced material.
+    /// </summary>
+    [DataMember]
+    [Display(Name = "Expected Arrival", Description = "Expected arrival date and time of the inbound material (optional).")]
+    public DateTime? ExpectedArrival { get; set; }
```
andreniggemann: According to ADR-005  DataTimeOffset is prefered

src/Moryx.Material/States/MaterialContainerState.cs
```diff
@@ -0,0 +1,32 @@
...
+using Moryx.StateMachines;
+
+namespace Moryx.Material.States;
+
+internal abstract class MaterialContainerState(MaterialContainer context, StateBase.StateMap stateMap) :
+    SyncStateBase<MaterialContainer>(context, stateMap)
+{
+    public abstract StateClassification Classification { get; }
+
+    public abstract void Advance(StateInformation info);
```
andreniggemann, line 13: Is it your plan to only have the Advance method as the only method on the state machine?
Currently it's only called via the OnEnter of the uninitialized state to jump to restore the state, right?

I am against having only a single method on the state machine, because it makes the expected events implicit.
Additionally I don't think there is generally a 1:1 mapping between calls that need to be routed through the state machine and the extra information stored for a state. So if you want to keep the 1:1 mapping I would take  object or a separate base class as a param.

src/Moryx.Material/States/InboundState.cs
```diff
@@ -1,40 +1,26 @@
...
-    /// <summary>
-    /// Optional cross-reference to a preceding material request.
-    /// </summary>
-    [DataMember]
-    public Guid? RequestReference { get; set; }
-}
\ No newline at end of file
+    public override void Advance(StateInformation info)
+    {
+        switch (Context.StateInformation)
```
andreniggemann: Why is the info parameter ignored? The same happens on most other implementations of Advance

andreniggemann (changes requested):

src/Moryx.Material/States/InboundStateInformation.cs
1nf0rmagician: In my mind at least, `IIdentity` is connected to the `IIdentifiableObject`, meaning it is used for actual cyber-physical things and I did not expect complex identifiers to be used for such processes by outside systems. If it is alright for you, I would leave it a string for now. Maybe use cases during the pre-release phase will show that an object is necessary here, then I would change that 🙂

src/Moryx.Material/States/MaterialContainerState.cs
1nf0rmagician: I'm with you here, I expect this to change in a following PR when I introduce the necessary features to change states through the module/facade. This is only the minimal implementation for now 🙂

src/Moryx.Material/States/InboundState.cs
1nf0rmagician: That is actually an error 🙈

src/Moryx.Material/Events/MaterialChangedEventArgs.cs
1nf0rmagician: I'm mostly with you here, the use case for which I kept this separate goes as follows: I have a packaging machine which processes multiple product types and the machine should automatically select the correct material container, when a new piece arrives. Now I put a new, empty material container on the station, hence I set the material of the container to the expected one, but the filling level did not change. Hence I only fire one event for exactly this.

In any other case I would fire both events and could also process both on the consumer side. 

Typing this, I can also see, that there might be a case made for such changes being processed atomically, i.e. consumers can be sure which event to expect if both change. Otherwise it is implementation specifci knowledge if I first fire material changed or quantity change 🤔

src/Moryx.Material/States/InboundStateInformation.cs
andreniggemann: It's ok for me. I just thought it might be possible that different Warehouse Management Systems use different formats or structures for this, but you are probably correct to keep it simple unless we really need it.

src/Moryx.Material/States/UninitializedState.cs
```diff
@@ -0,0 +1,41 @@
...
+            case AvailableStateInformation:
+                NextState(StateAvailable);
+                return;
+            case OutboundStateInformation:
+                NextState(StateOutbound);
+                return;
+            case DeregisteredStateInformation:
+                NextState(StateDeregistered);
+                return;
+            default:
```
andreniggemann, line 38: Should the default case call InvalidState as well?

src/Moryx.Material/BasicMaterialContainer.cs
```diff
@@ -1,20 +1,15 @@
...
+using System.ComponentModel.DataAnnotations;
 
 namespace Moryx.Material;
 
 /// <summary>
 /// Ready-to-use, non-abstract <see cref="MaterialContainer"/> for the most common cases.
 /// Application engineers may register this resource type directly without subclassing.
 /// </summary>
-[DataContract]
-[ResourceRegistration]
```
andreniggemann, line 15: I think the ResourceRegistration here or better on the MaterialContainer would be nice to have so that you can assume a logger exists in MaterialContainer state transitions

src/Moryx.Material/MaterialContainer.cs
```diff
@@ -1,100 +1,197 @@
...
 [DataContract]
-public abstract class MaterialContainer : Resource, IMaterialContainer
+public abstract class MaterialContainer : Resource, IMaterialContainer, IStateContext
 {
-    private MaterialContainerStateBase _state = new RequestedState();
-    private string? _material;
-    private decimal _quantity;
+    #region IMaterialContainer
 
+    // ToDo: If ResourceRelationType.Aggregation should not be used alone, a new type must be introduced
```
andreniggemann, line 27: Could the container be considered a CurrentExchangeablePart? But I would also not be opposed to creating a new ResourceRelationType for MaterialContainers

src/Moryx.Material/MaterialContainer.cs
```diff
@@ -1,100 +1,197 @@
...
-        }
-    }
+    [EntrySerialize, ReadOnly(true)]
+    [Display(Name = "Quantity", Description = "Current amount of material held by this container.")]
+    public double Quantity { get; private set; }
 
     /// <inheritdoc />
     [DataMember]
+    [EntrySerialize]
+    [Display(Name = "Unit", Description = "Unit of measure for the quantity (e.g., kg, pcs).")]
```
andreniggemann, line 58: Should we provide either an enum or a couple of string constants for predefined unit types?

src/Moryx.Material/MaterialContainer.cs
```diff
@@ -1,100 +1,197 @@
...
+        }
+
+        var eventArgs = new MaterialUpdatedEventArgs(this) { Kind = update.Kind };
+        if (update.Kind.HasFlag(UpdateKind.MaterialType))
+        {
+            eventArgs.OldMaterial = Material;
+            eventArgs.NewMaterial = Material = update.Material;
+        }
+        // ToDo: Should a negative Quantity throw an invalid operation exception?
+        // Should update.Quantity be prevented beforehand or interpreted here as a relative update?
```
andreniggemann: Could we maybe define a separte UpdateKind for relative updates?
I could see a need for both versions. If I added 200 pieces it makes sense to model that as relative, but if I empty the container it makes sense to model that as a change of the absolute amount to zero.

src/Tests/Moryx.Material.IntegrationTests/MaterialContainerTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        });
+    }
+
+    [Test]
+    public async Task RegisterMaterial_FromHostResource_CreatesContainerInAvailableState()
+    {
+        // Arrange
+        var hostId = await CreateContainerHostAsync();
+        const string material = "Some Material";
+        const decimal quantity = 42m;
```
andreniggemann: quantities are doubles now, right?

src/Tests/Moryx.Material.IntegrationTests/MaterialContainerTests.cs
```diff
@@ -0,0 +1,135 @@
...
+[TestFixture]
+internal sealed class MaterialContainerTests : TestBase
+{
+    [Test]
+    public async Task CreateUnsafe_OfBasicMaterialContainer_CreatesContainerResourceInUnitializedState()
+    {
+        // Arrange
+        // Act
+        var containerId = await _resourceManagement.CreateUnsafeAsync(typeof(BasicMaterialContainer), _ => Task.CompletedTask);
+        var container = _resourceManagement.GetResource<IMaterialContainer>(containerId);
```
andreniggemann: In the other tests, this is already part of the Assert block

andreniggemann (changes requested):

src/Moryx.Material/States/UninitializedState.cs
1nf0rmagician: I added an error log for it 😊  
Since the exception could be thrown on a restart of the application this would block the whole resource management from starting and I don't want that or think that the error is significant enough for that.

src/Moryx.Material/BasicMaterialContainer.cs
1nf0rmagician: Since I just added logging, I think I will need it yes. Unfortunately there is an issue with resource reference properties of type `IResource` being automatically populated by the DI on creation with a new resource of a random resource type. I'm not sure how to fix that currently 🙈

src/Moryx.Material/MaterialContainer.cs
1nf0rmagician: I would not consider it a `CurrentExchangeablePart`, when compared to tools I feel like material containers have a very different role 🤔

src/Moryx.Material/MaterialContainer.cs
1nf0rmagician: I thought about that and decided against it for now for the following two reasons:
1. currently the unit does not serve a purpose with regard to the feature specifications, instead it is rather supplementary information
2. the general material container could use units for length, volume, weight or amount with several orders of magnitude being viable and required to be considered (leaving aside that if we want to continue to use it to enrich information displayed to the user we would probably also need to include imperial units...), which I though to be to many at this level.

We could think of making the property virtual allowing custom container types to override them and apply dedicated possible value attributes 🤔

src/Moryx.Material/MaterialContainer.cs
1nf0rmagician: >Could we maybe define a separte UpdateKind for relative updates

I like that idea for now, will do that 👌

src/Tests/Moryx.Material.IntegrationTests/MaterialContainerTests.cs
1nf0rmagician: right, missed that

PR #1282, opened 2026-06-16
Load module items once lazily in Launcher

1nf0rmagician:
Increases the performance when switching between modules in the launcher by caching module items and regions in the `IShellNavigator` service. This also removes recurring error logs caused by repeated assembly scans.

**Note:** I feel bad making the launcher config static, but it provides a significant benefit in the static Lazy context. Also the service is usually registered as a singleton and the config is one retrieved once from the config manager. Still, I'm open for suggestions to improve

### DIFF
--- a/src/Moryx.Launcher/ShellNavigator.cs
+++ b/src/Moryx.Launcher/ShellNavigator.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
 // Licensed under the Apache License, Version 2.0
 
+using System.Linq;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.RazorPages;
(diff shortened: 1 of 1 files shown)


src/Moryx.Launcher/ShellNavigator.cs
```diff
@@ -76,74 +101,66 @@ public async Task<IReadOnlyList<ModuleItem>> GetModuleItemsAsync(HttpContext con
...
     }
 
-    private async Task<CompiledPageActionDescriptor[]> CompiledPageActionDescriptors(HttpContext context)
+    public ShellNavigator(
+        EndpointDataSource endpointsDataSource,
+        PageLoader pageLoader,
+        IConfigManager configManager,
+        IOptionsMonitor<MoryxIdentityOptions> options,
+        IMemoryCache memoryCache,
+        ILoggerFactory logger)
```
dbeuchler, line 116: Could you move the ctor back to the top? under the dependencies/fields/properties? private methods to the end

src/Moryx.Launcher/ShellNavigator.cs
```diff
@@ -21,53 +22,77 @@ public class ShellNavigator : IShellNavigator, ILauncher
...
+
+        return regions;
     }
 
-    /// <inheritdoc />
-    public async Task<IReadOnlyList<ModuleItem>> GetModuleItemsAsync(HttpContext context)
+    private static EndpointDataSource _endpointsDataSource;
+    private static PageLoader _pageLoader;
+    private static readonly Lazy<PageActionDescriptorAndModuleItem[]> _descriptorsAndModules =
+        new(LoadCompiledActionDescriptors, LazyThreadSafetyMode.ExecutionAndPublication);
```
dbeuchler, line 66: move to the top of the class

andreniggemann:
Could the static stuff be put into a simple singleton service?
That way it would still be cached and we would get rid of the static stuff.

dbeuchler:
> Could the static stuff be put into a simple singleton service? That way it would still be cached and we would get rid of the static stuff.

I think @1nf0rmagician dont want to refactor the whole ShellNavigator in this PR, but yes this sounds good in another PR

seveneleven:
> Could the static stuff be put into a simple singleton service? That way it would still be cached and we would get rid of the static stuff.

+1 

That was also my take on _"I feel bad making the launcher config static, but it provides a significant benefit in the static Lazy context"_ which I missed to submit.

src/Moryx.Launcher/ShellNavigator.cs
1nf0rmagician: I arranged this just like in the [ReflectionTool](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/b7f873f4f1f761437958ab62f12c6c8cfa4f7334/src/Moryx/Tools/ReflectionTool.cs?plain=1#L26C1-L26C47) and as described e.g. in the clean code book, meaning private method right below it's first usage. I find this convention equally good to follow as the suggested sorting you proposed. 

We had a discussion about such topics just recently in the team and since to our knowlegde there is no supported auto-sorting tooling, which we could broadly apply, I would like to keep it this way. 

I understand of course that now most of the class is the initializer method for the lazy object. This I would solve, however, as suggested by another service, rather than in here.

src/Moryx.Launcher/ShellNavigator.cs
1nf0rmagician: As mentioned above, it is the top, ignoring private methods which are first and only used for the lazy initialization 🙂

1nf0rmagician:
> Could the static stuff be put into a simple singleton service?

Yes, just as you suggest @andreniggemann and yes to what @dbeuchler responded. Also, separating it into a cache service would make this an opt-in feature at best, as it would require modifying the application build section. This is however, is not what I want to got for, as I consider the error spaming in the log from the repeated reflection attempts very "sub-optimal" at best.

src/Moryx.Launcher/ShellNavigator.cs
dbeuchler: we have (again unwritten) rules, that we structure our classes always as follows: 

1. Dependencies (outdated for ctor-injection)
3. private Fields
4. Public Methods
5. Private Methods 
6. Events

Resharper / Rider provides you support and MORYX was always driven by their refactoring rules.

I dont want to complicate things here; I will approve that, but just keep that in mind. We can also fix that in the following steps.

dbeuchler (approved):

Issue #1061, opened 2026-01-29
Rework of Web-Projects


dbeuchler:
The web projects were migrated from one angular version to another. New conventions were not applied.

Here are some conventions we should apply:

- [x] All components, directives, and pipes are standalone. Only use `standalone: false` for legacy compatibility.
- [x] Prefer signals over RxJS for component state management.
- [x] Use `@if`, `@for`, `@switch`, `@defer` instead of structural directives.
- [x] Write code that works without zone.js (Angular 21 default).
- [x] Prefer inject() over Constructor Injection
- [ ] Check usage of deprecated provideAnimations
- [ ] Replace ngOnInit by appropriate [signal handling](https://medium.com/@lielshani/your-component-doesnt-need-ngoninit-d0b0be1acc1d) 
- [ ] Sometimes there are extensions for promises to convert them to observables. I already refactored some usages to `lastValueFrom` to get rid of the extensions. It is still mixed and should be refactored. ng-openapi-gen supports generation of observables. Maybe this is a good point to start and harmonize.

dbeuchler:
@1nf0rmagician That was long-winded, but I think we contributed to raising the web projects to a higher level.

PR #1285, opened 2026-06-18
Redesign of Moryx.Launcher to full Angular-App

dbeuchler:
## Summary
Complete redesign of the launcher from a mixed Razor/Angular implementation to a fully Angular-based web component architecture. This change improves module isolation, eliminates CSS conflicts, simplifies maintenance, and provides a foundation for future extensibility.

Screen Recording 2026-06-19 at 11 52 42

## Background
The previous launcher implementation mixed Razor pages with Angular and web components, which caused:
- CSS isolation problems where launcher styles could override module styles
- Inconsistent module appearances
- Difficulty maintaining separation of concerns

### Architecture
- Entire launcher is now a single Angular web component with complete style isolation
- Slot-Based Architecture Uses slots for regions (top, right, app) allowing modules to be plugged in without code modifications
- Simplified Codebase: Consolidated services for module and culture initialization, removed redundant code

### Code Structure
- Refactored layout management through Angular components
- Support for multiple display modes (full, operator, full screen)
- Dynamic slot-based content rendering
- Removed hard-coded logic (assembly info helper, module headers)
- Consolidated information retrieval through runtime endpoints

### Styling and Theming
- Default Angular Material styling with MORYX theme
- Removed custom/mixed styles from previous implementation
- Resolved flex container and layout issues
- Improved text truncation with ellipsis for long module names

### UI/UX Improvements
- YouTube-style scrollable and collapsible navigation menu
- Persistent UI state (operator mode, navigation collapse) across reloads
- Improved responsive design foundation
- Consistent button styling using Angular defaults

## Compatibility
- Maintains major and minor compatibility with existing modules
- All launcher features re-implemented
- Minor styling differences (intentional improvements, default theme)

## Assets and Localization
- Assets (logos) and localization JSON files loaded from server
- Supports translation while maintaining web component independence
- Module and language inputs passed from Razor layer to Angular component

## Benefits
1. Eliminates CSS conflicts between launcher and modules
2. Cleaner, more organized codebase
3. Slot-based architecture enables custom layout injection
4. Persistent state, improved navigation, responsive design
5. Fully leverages Angular Material 22 capabilities

## Spotlight Search
The search bar was refactored to be a spotlight search, inspired by macOS. Can be opened by search button or by CMD+K / CTRL+K on desktops.

## Dark mode

After replacing hardcoded colors and rely on fully angular material it is easy to implement a theme switch between light, dark and system and makes MORYX more modern web application. 

Screen Recording 2026-06-22 at 13 52 30

## Migration Path
No breaking changes for modules. The slot-based architecture is designed to work with existing module implementations without modification. If custom modules are using hardcoded colors, they may be wrong in dark-mode and can be replaced by global material variables.

## Follow-Up Tasks/Ideas

**Responsivenes**

Make MORYX-launcher fully responsive (desktop/+touch, tablets, mobile phones


src/Moryx.Launcher/app/src/environments/environment.prod.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+ * Licensed under the Apache License, Version 2.0
+*/
+
+import { getPathBase } from '@moryx/ngx-web-framework/environments';
+
+let path_base = getPathBase('/Notifications');
```
dbeuchler: Hint for me to check what should be the correct value if path_base is used

src/Moryx.Resources.Benchmarking/BenchmarkResource.cs
```diff
@@ -190,7 +190,20 @@ public void PublishMarkdownNotification()
...
+            Title = "Markdown Instruction",
+            Instructions =
+            [
+                new VisualInstruction
+                {
+                    Content = message,
+                    Type = InstructionContentType.Text
+                }
+            ]
+        });
```
dbeuchler: Hint for me to remove before merge / add another method to push instruction

src/Moryx.Resources.Web/app/src/app/app.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .resources-toolbar {
-  color: var(--mat-sys-surface-container-low);
```
1nf0rmagician, line 19: Why did you move away from the material variables here?

src/Moryx.Resources.Web/app/src/app/components/resource-tree/resource-tree.scss
```diff
@@ -27,8 +27,8 @@
     background-color: var(--mat-sys-surface-container-highest);
   }
 
-  &.moryx-selected {
-    background: $primary-light;
+  &.selected {
```
1nf0rmagician, line 30: Reminder to decide on the common "selected" color and background color

src/Moryx.Products.Web/app/src/environments/environment.ts
```diff
@@ -10,5 +10,5 @@
 export const environment = {
   production: false,
   assets: "/",
-  rootUrl: 'https://localhost:5000',
+  rootUrl: 'https://localhost:5002',
```
1nf0rmagician: Intended?

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-recipes/product-recipes.scss
```diff
@@ -1,75 +1,68 @@
...
-  margin-left: 10px;
-  overflow: hidden;
-}
-
-.truncate-name {
-  white-space: nowrap;
+  margin-left: 8px;
+  margin-right: 8px;
   overflow: hidden;
-  text-overflow: ellipsis;
```
1nf0rmagician, line 64: Did you remove the truncation of the names intentionally?

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-recipes/product-recipes.html
```diff
@@ -1,13 +1,13 @@
-<div class="recipes-overview">
```
1nf0rmagician, line 1: If you are optimizing styling the outermost div could propably be removed and replaced by styling the :host element in the component

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-recipes/product-recipes-details/product-recipes-details.scss
```diff
@@ -1,10 +1,7 @@
-@use "@moryx/ngx-web-framework/styles/variables" as *; 
-
 :host {
   display: flex;
   flex-flow: column nowrap;
   justify-content: flex-start;
-  width: 100%;
```
1nf0rmagician, line 7: Is the width set somewhere else?

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-parts/product-parts.scss
```diff
@@ -58,5 +54,5 @@
 }
 
 .moryx-selected {
-  background-color: $primary-light;
+  background-color: $primary;
```
1nf0rmagician, line 57: Reminder to check whether primary background has enough contrast with default text color

src/Moryx.Media.Web/app/src/app/components/variant-overview/variant-overview.scss
```diff
@@ -34,75 +34,66 @@
...
-    height: 40px;
-    width: 40px;
-  }
+.variant-item {
+  cursor: pointer;
+  width: 100%;
+  user-select: none;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -webkit-touch-callout: none;
```
1nf0rmagician: Further down you removed the prefix settings and as far as I read up on the topic you were doing it because they are obsolete, if I'm not mistaken you can also ismplify this i guess 🙃

src/Moryx.Launcher/Moryx.Launcher.csproj
```diff
@@ -33,12 +33,9 @@
   </ItemGroup>
 
   <Import Project="..\..\.build\WebProjects.props" />
-  <Target Name="BuildUI" Condition="('$(Configuration)'!='Debug' Or !Exists('wwwroot\js\dist\main.js')) And ('$(NoBuild)' != 'true')">
-    <Exec Command="dotnet tool restore" />
-    <Exec Command="dotnet libman restore" />
```
1nf0rmagician, line 38: Not sure whether you mentioned it in previous discussions, but could you tell me again what the new method of providing offline fonts for modules through the launcher is? At least that was the intention of delivering the fonts with the launcher initially 😅

src/Moryx.Launcher/Pages/Index.cshtml
```diff
@@ -1,36 +1,27 @@
...
+        @{
+            var jsonOptions = new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
+            jsonOptions.Converters.Add(new JsonStringEnumConverter());
         }
-    }
-</div>
+        const moduleOverview = document.getElementById('moduleOverview');
+        moduleOverview.webModuleItems = @Html.Raw(JsonSerializer.Serialize(webModuleItems, jsonOptions));
+    </script>
+}
```
1nf0rmagician, line 27: Is this preferable over an API endpoint from which the module items are loaded? 🤔

src/Moryx.Launcher/Models/CultureModel.cs
1nf0rmagician, line 1: Did the `ModuleItem`, `RegionItem` and `WebModuleItem` become serialized models as well and could be moved in this new folder?

src/Moryx.Launcher/app/package.json
```diff
@@ -1,53 +1,42 @@
...
-    "postcss": "^8.5.15",
-    "sass": "^1.101.0",
-    "sass-loader": "^17.0.0",
-    "typescript": "~6.0.3",
-    "vitest": "^4.1.9",
-    "webpack": "^5.107.2",
-    "webpack-cli": "^7.0.3"
+    "ng-openapi-gen": "^1.0.5",
+    "typescript": "~6.0.2",
+    "vitest": "^4.0.8"
```
1nf0rmagician: you downgraded some dependencies probably by accident

src/Moryx.Launcher/app/package.json
```diff
@@ -1,53 +1,42 @@
...
-    "@angular/platform-browser-dynamic": "^22.0.0",
     "@angular/router": "^22.0.0",
     "@moryx/ngx-web-framework": "^22.0.0",
-    "css-loader": "^7.1.0",
-    "lit": "^3.3.3",
-    "postcss": "^8.5.15",
-    "postcss-cli": "^11.0.0",
+    "@ngx-translate/core": "^18.0.0",
+    "@ngx-translate/http-loader": "^18.0.0",
+    "material-symbols": "^0.45.1",
```
1nf0rmagician, line 28: Related to the font discussion: I agree that we should not continue to support the legacy material icons, but i think we provided it before and would need to kepe them till MORYX 12

src/Moryx.Launcher/Pages/Shared/_Layout.cshtml
```diff
@@ -1,323 +1,87 @@
...
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/css/site.css" />
-    <link rel="icon" type="image/x-icon" href="~/_content/Moryx.Launcher/favicon.ico">
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/lib/material-icons/iconfont/material-icons.css" />
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/lib/material-symbols/outlined.css" />
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/lib/material-symbols/rounded.css" />
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/lib/material-symbols/sharp.css" />
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/lib/fontsource-roboto/400.css" />
-    <link rel="stylesheet" href="~/_content/Moryx.Launcher/css/app.min.css" />
-    @RenderSection("Styles", false)
+    <link rel="stylesheet" href="~/_content/Moryx.Launcher/styles.css"/>
```
1nf0rmagician, line 43: You are trying to include the styles.css from the packages static assets here from what I'm seeing, but that one was removed wasn't it?

src/Moryx.Launcher/.gitignore
1nf0rmagician, line 1: does this still work if there are subfolders within wwwroot left like the page icon?

src/Moryx.Launcher/app/src/styles.scss
```diff
@@ -0,0 +1,15 @@
+@use 'roboto-fontface/css/roboto/roboto-fontface' as *;
+@use 'material-symbols/outlined' as *;
+@use '@angular/material' as mat;
+@use '@moryx/ngx-web-framework/styles/theme-colors' as *;
+
+html.dark-theme {
+  @include mat.theme((
+    color: (
+      primary: $primary-palette,
```
1nf0rmagician, line 9: is it intended not to have set the secondary palette here?

src/Moryx.Launcher/app/src/environments/environment.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the Apache License, Version 2.0
+*/
+
+// This file can be replaced during build by using the `fileReplacements` array.
+// `ng build` replaces `environment.ts` with `environment.prod.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+  production: false,
+  assets: "/_content/Moryx.Launcher/",
```
1nf0rmagician: why do you have the _content directory in the dev environment? That should only exist in the package shouldn't it?

src/Moryx.Launcher/app/src/assets/favicon.ico
1nf0rmagician, line 1: If this is moved to the angular app's assets directory, it is not going to be placed in the _content directory of the package later or am i missing a copying mechanism somewhere?

src/Moryx.Launcher/app/src/assets/languages/de.json
```diff
@@ -0,0 +1,63 @@
...
+    "SERVER_TIME": "Serverzeit",
+    "CLOSE": "Schließen",
+    "LABEL_TITLE": "Titel",
+    "LABEL_PRODUCT": "Produkt",
+    "LABEL_DESCRIPTION": "Beschreibung",
+    "LABEL_VERSION": "Version",
+    "LABEL_INFORMATIONAL_VERSION": "Informationsversion",
+    "LABEL_COMPANY": "Unternehmen",
+    "LABEL_CONFIGURATION": "Konfiguration",
+    "LABEL_COPYRIGHT": "Urheberrecht",
```
1nf0rmagician: "Urheberrechtshinweis" or "Copyright" as well

src/Moryx.Launcher/app/src/app/app.ts
```diff
@@ -0,0 +1,90 @@
...
+  private shortcutService = inject(ShortcutService);
+  private languageService = inject(LanguageService);
+  private translateService = inject(TranslateService);
+  private searchService = inject(SearchService);
+  private destroyRef = inject(DestroyRef);
+
+  // Web component inputs
+  webModuleItems = input.required<WebModuleItem[]>();
+  externalModuleItems = input.required<ExternalModuleItem[]>();
+  supportedCultures = input.required<CultureModel[]>();
```
1nf0rmagician, line 47: Just out of interest, how does the required input work if the content is serialized and set on the element in the razor page 🤔

src/Moryx.Launcher/app/src/app/app.ts
```diff
@@ -0,0 +1,90 @@
...
+      this.destroyRef
+    );
+
+    effect(() => {
+      this.moduleService.modules.set([...this.webModuleItems(), ...this.externalModuleItems()]);
+      this.cultureService.supportedCultures.set(this.supportedCultures());
+
+      const authBaseAddress = this.authBaseAddress();
+      this.authService.authBaseAddress = authBaseAddress;
+      this.authService.authConfigured.set(!!authBaseAddress && authBaseAddress.length > 0);
```
1nf0rmagician: Did you intentionally put all this in the same effect and not split it to prevent multiple updates for each?

Also since you set signals on the different services it would be a good idea to place that part in an untracked section preventing possible unintended triggers down the line, isn't it?

src/Moryx.Launcher/app/src/app/app.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    this.translateService.addLangs([
+      TranslationConstants.LANGUAGES.EN,
+      TranslationConstants.LANGUAGES.DE,
+      TranslationConstants.LANGUAGES.IT,
+      TranslationConstants.LANGUAGES.ZH,
+    ]);
+    this.translateService.setFallbackLang("en");
+
+    console.log("Language: " + this.languageService.getFallbackLang());
```
1nf0rmagician: Intended to keep this log?

src/Moryx.Launcher/app/src/app/app.scss
```diff
@@ -0,0 +1,37 @@
+.material-symbols-outlined {
+  font-family: 'Material Symbols Outlined';
+}
```
1nf0rmagician, line 4: Why do you need this class defninition 🤔

src/Moryx.Launcher/app/src/app/app.config.ts
```diff
@@ -0,0 +1,36 @@
...
+import { provideTranslateHttpLoader } from '@ngx-translate/http-loader';
+import { environment } from '../environments/environment';
+
+import { routes } from './app.routes';
+import { LocationPersistenceService } from './services/location-persistence.service';
+
+export const appConfig: ApplicationConfig = {
+  providers: [
+    provideBrowserGlobalErrorListeners(),
+    provideRouter(routes, withHashLocation()),
```
1nf0rmagician: Can be removed, same as `app.routes.ts`

src/Moryx.Launcher/app/src/app/services/module.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    let best: string | null = null;
+    for (const module of this.modules()) {
+      const route = module.route;
+      const idx = pathname.indexOf(route);
+      if (idx < 0) {
+        continue;
+      }
+      const afterChar = pathname[idx + route.length];
+      if (afterChar === undefined || afterChar === '/' || afterChar === '?') {
+        if (!best || route.length > best.length) {
```
1nf0rmagician, line 44: Combine ifs?

src/Moryx.Launcher/app/src/app/services/shortcut.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  ctrl?: boolean;
+  shift?: boolean;
+  alt?: boolean;
+  action: () => void;
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+/** Global keyboard shortcut registry. Bindings are auto-removed when the registering component is destroyed. */
```
1nf0rmagician, line 25: Am I right in assuming you want to make this service somehow available from within the loaded modules as well?

src/Moryx.Launcher/app/src/app/services/theme.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  mode = signal<ThemeMode>(this.getStoredMode());
+
+  constructor() {
+    this.systemDarkQuery.addEventListener('change', () => this.applyTheme());
+    this.applyTheme();
+  }
+
+  setMode(mode: ThemeMode): void {
+    this.mode.set(mode);
+    window.localStorage.setItem(this.storageKey, mode);
```
1nf0rmagician: Since you added the "which regions to show"-information to the launcher state stored in the local storage, I think this would go well within a "LauncherState" as well, don't you think?

src/Moryx.Launcher/app/src/app/services/theme.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  private getStoredMode(): ThemeMode {
+    const stored = window.localStorage.getItem(this.storageKey);
+    if (stored === 'light' || stored === 'dark' || stored === 'system') {
+      return stored;
+    }
+    return 'system';
+  }
+
+  private applyTheme(): void {
```
1nf0rmagician: Do you want to put this in an effect of the mode signal? Maybe also for storing the mode it?

src/Moryx.Launcher/app/src/app/about-dialog/about-dialog.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  protected TranslationConstants = TranslationConstants;
+  isMac = navigator.platform.toUpperCase().includes('MAC');
+
+  shortcuts = [
+    { label: TranslationConstants.ABOUT.SHORTCUT_SPOTLIGHT, keys: { mac: '⌘ ⌥ K', other: 'Ctrl+Alt+K' } },
+    { label: TranslationConstants.ABOUT.SHORTCUT_FULL_MODE, keys: { mac: '⌘ ⌥ 1', other: 'Ctrl+Alt+1' } },
+    { label: TranslationConstants.ABOUT.SHORTCUT_OPERATOR_MODE, keys: { mac: '⌘ ⌥ 2', other: 'Ctrl+Alt+2' } },
+    { label: TranslationConstants.ABOUT.SHORTCUT_FULLSCREEN_MODE, keys: { mac: '⌘ ⌥ 3', other: 'Ctrl+Alt+3' } },
+  ];
```
1nf0rmagician: Do you want to add the string representation of the shortcuts to the `ShortcutBinding` in the shortcut service and take them from there? 

If my assumption that you plan to populate the service from nested applications is correct, it would allow the about dialog to show all bindings and it would keep the information which shortcuts exist in one place 🙂

src/Moryx.Launcher/app/src/app/horizontal-module-nav/horizontal-module-nav.html
```diff
@@ -0,0 +1,32 @@
...
+    </div>
+    <span class="nav-title">{{ TranslationConstants.NAVIGATION.ALL | translate }}</span>
+  </button>
+
+  @for (module of visibleModules(); track module.route) {
+    <a class="nav-item" [href]="resolveHref(module)" [title]="module.description" [class.active]="module.route === activeRoute()">
+      <div class="icon-wrapper">
+        <mat-icon>{{ module.icon }}</mat-icon>
+        @if (asWeb(module); as webItem) {
+          <app-notification-badge [eventStream]="webItem.eventStream ?? ''"></app-notification-badge>
```
1nf0rmagician: you could filter the event stream is null case in the if as well

src/Moryx.Launcher/app/src/app/horizontal-module-nav/horizontal-module-nav.html
```diff
@@ -0,0 +1,32 @@
+<nav #navEl>
+  <button class="nav-item" [matMenuTriggerFor]="moduleGridMenu.moduleGridMenu">
```
1nf0rmagician: Do you want to hide the button if there are no modules, e.g. if user is not logged in?

src/Moryx.Launcher/app/src/app/navigation/horizontal-module-nav/horizontal-module-nav.scss
```diff
@@ -0,0 +1,80 @@
...
+  flex: 1;
+  gap: 2px;
+  text-decoration: none;
+  color: inherit;
+  position: relative;
+  padding: 8px 4px;
+  border-radius: 4px;
+  transition: background 0.15s ease;
+
+  &:hover {
```
1nf0rmagician, line 31: You said there is no material color variable for selections, for hovering neither?

src/Moryx.Launcher/app/src/app/horizontal-module-nav/horizontal-module-nav.ts
```diff
@@ -0,0 +1,64 @@
...
+  protected TranslationConstants = TranslationConstants;
+  private navEl = viewChild.required<ElementRef<HTMLElement>>('navEl');
+
+  private visibleCount = signal(Number.MAX_SAFE_INTEGER);
+
+  visibleModules = computed(() => this.modules().slice(0, this.visibleCount()));
+
+  private resizeObserver = new ResizeObserver(entries => {
+    const navWidth = entries[0].contentRect.width;
+    const count = Math.max(0, Math.floor(navWidth / MIN_ITEM_WIDTH) - 1);
```
1nf0rmagician: Isn't it ceil and -2 because of the grid and more menus?

src/Moryx.Launcher/app/src/app/navigation/module-grid-menu/module-grid-menu.scss
```diff
@@ -0,0 +1,45 @@
...
+  text-decoration: none;
+  color: inherit;
+  border: 1px solid var(--mat-sys-outline-variant);
+  border-radius: 6px;
+  font-size: 0.8rem;
+  font-weight: 600;
+  text-align: center;
+  transition: background 0.15s ease;
+
+  &:hover {
```
1nf0rmagician, line 31: Different hover color than on the navigation rail itself

src/Moryx.Launcher/app/src/app/module-overview/module-overview.scss
```diff
@@ -0,0 +1,75 @@
...
+  min-width: 0;
+  border: 1px solid var(--mat-sys-outline-variant);
+  border-radius: 4px;
+  padding: 1rem;
+  text-decoration: none;
+  color: inherit;
+  background: var(--mat-sys-surface-container-lowest);
+  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.12);
+
+  &:hover {
```
1nf0rmagician, line 32: Another hover color style 😅

src/Moryx.Launcher/app/src/app/module-overview/module-overview.ts
```diff
@@ -0,0 +1,27 @@
...
+@Component({
+  selector: 'app-module-overview',
+  imports: [],
+  templateUrl: './module-overview.html',
+  styleUrl: './module-overview.scss',
+})
+export class ModuleOverview {
+  private locationPersistenceService = inject(LocationPersistenceService);
+
+  webModuleItems = input.required<WebModuleItem[]>();
```
1nf0rmagician, line 21: Any reason why unlike everywhere else you are not injecting the module service here to get the modules? 🤔

src/Moryx.Launcher/app/src/app/more-menu/more-menu.html
```diff
@@ -0,0 +1,122 @@
...
+    <mat-icon>language</mat-icon>
+    <span>{{ TranslationConstants.MORE.LANGUAGE  | translate }}</span>
+  </button>
+  <button mat-menu-item [matMenuTriggerFor]="viewsMenu">
+    <mat-icon>visibility</mat-icon>
+    <span>{{ TranslationConstants.MORE.VIEWS  | translate }}</span>
+  </button>
+  <button mat-menu-item [matMenuTriggerFor]="themeMenu">
+    <mat-icon>palette</mat-icon>
+    <span>{{ TranslationConstants.MORE.THEME  | translate }}</span>
```
1nf0rmagician: spaces in the pipe operations

src/Moryx.Launcher/app/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,51 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+      .then(() => {
+        this.isLoggedIn.set(false);
+        this.userName.set('');
+        window.location.assign('/');
+      })
+      .catch((err) => console.log(err));
```
1nf0rmagician: You could use the snackbar service from the ngx-web-framework to show the error here

src/Moryx.Launcher/app/src/app/services/culture.service.ts
```diff
@@ -0,0 +1,40 @@
...
+export class CultureService {
+  private cookieService = inject(CookieService);
+
+  private readonly cookieName = '.AspNetCore.Culture';
+
+  /** Available cultures provided by the server. */
+  supportedCultures = signal<CultureModel[]>([]);
+
+  /** The currently active culture, derived from the ASP.NET culture cookie. */
+  currentCulture = computed(() => {
```
1nf0rmagician, line 23: Am I missing something or could this computed be a simple get method as it is not referencing any other signals? 🤔

src/Moryx.Launcher/app/src/app/services/culture.service.ts
```diff
@@ -0,0 +1,40 @@
...
+      return '';
+    }
+    // Cookie format: c=de-DE|uic=de-DE
+    const culture = cultureCookie.match(/c=([^|]+)/)?.[1] ?? '';
+    return culture;
+  });
+
+  selectCulture(culture: CultureModel) {
+    // Expire after 1 year
+    const expires = new Date(Date.now() + 365 * 24 * 3600 * 1000);
```
1nf0rmagician, line 35: We have feedback for the visual instructions module in the backlog that while one year is long, operators on the shopfloor usually don't expect to reconfigure this and it could very well not be changed for years. What do you think of automatically refreshing/replacing the cookie when it is retrieved or something like that?

src/Moryx.Launcher/app/src/app/services/launcher-state.service.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Licensed under the Apache License, Version 2.0
+*/
+
+import { Injectable, signal } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+/** Manages the launcher's UI state (layout, nav, regions) and persists it to localStorage. */
+export class LauncherStateService {
```
1nf0rmagician: You could keep a state object in memory rather than deserializing the object from local storage for each interaction. I know we don't expect to many calls to the signals between page refreshes, so I don't think its mandatory. But I sure would feel better about ist 😅

src/Moryx.Launcher/app/src/app/spotlight-search/spotlight-search.html
```diff
@@ -0,0 +1,50 @@
...
+      />
+      @if (query()) {
+        <button mat-icon-button (click)="onQueryChange('')" tabindex="-1" [attr.aria-label]="TranslationConstants.SEARCH.CLEAR | translate">
+          <mat-icon>close</mat-icon>
+        </button>
+      }
+    </div>
+
+    @if (suggestions().length > 0) {
+      <div class="divider"></div>
```
1nf0rmagician: why not use an actual `mat-divider`element?

src/Moryx.Launcher/app/src/app/spotlight-search/spotlight-search.scss
```diff
@@ -0,0 +1,147 @@
...
+.result-item {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  padding: 4px 8px;
+  cursor: pointer;
+  border-radius: 4px;
+  color: var(--mat-sys-on-surface);
+
+  &.active {
```
1nf0rmagician: I think active in this case means what hover means in other situations? Anyway check for consistency on the colors

src/Moryx.Launcher/app/src/app/spotlight-search/spotlight-search.html
```diff
@@ -0,0 +1,50 @@
+@if (isOpen()) {
+  <div class="backdrop" (click)="close()"></div>
+
+  <div class="panel" role="dialog" aria-modal="true" [attr.aria-label]="TranslationConstants.SEARCH.DIALOG_LABEL | translate">
+    <div class="search-row">
+      <mat-icon class="search-icon">search</mat-icon>
+      <input
```
1nf0rmagician, line 7: Did you check whether an input using the angular autocomplete options could work for us here as well? It would reduce the amount of custom styling you had to do and also the handling of visible list elements.

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
```diff
@@ -0,0 +1,34 @@
+<div class="search-container" [class.expanded]="expanded()">
```
1nf0rmagician: you could store the signal value in a `@let` statement to only have the signal event chain reference once in the template

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
```diff
@@ -0,0 +1,34 @@
...
+  <button mat-icon-button aria-label="Close" (click)="collapse()" class="close-button" tabindex="-1">
+    <mat-icon>close</mat-icon>
+  </button>
+</div>
+
+<mat-autocomplete #auto="matAutocomplete" (optionSelected)="onOptionSelected($event)" [displayWith]="displayFn">
+  @for (suggestion of suggestions(); track suggestion.text) {
+    <mat-option [value]="suggestion">
+      <mat-icon>{{ suggestion.url ? 'open_in_new' : 'search' }}</mat-icon>
+      {{ suggestion.text }}
```
1nf0rmagician, line 33: Do you want to make this an anchor element to maybe also allow preview of the link we navigate to on click?

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
```diff
@@ -0,0 +1,75 @@
...
+    const q = this.query().trim();
+    if (q) {
+      this.searchService.search(q, true);
+    }
+  }
+
+  onOptionSelected(event: MatAutocompleteSelectedEvent): void {
+    const suggestion = event.option.value as SearchSuggestion;
+    if (suggestion.url) {
+      window.location.href = suggestion.url;
```
1nf0rmagician: For consistency in another location i cannot remember where you used `window.location.assign`

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
```diff
@@ -0,0 +1,75 @@
...
+  imports: [MatIconModule, MatButtonModule, MatAutocompleteModule, TranslatePipe],
+  templateUrl: './toolbar-search.html',
+  styleUrl: './toolbar-search.scss',
+})
+export class ToolbarSearch {
+  private searchService = inject(SearchService);
+
+  @ViewChild('searchInput') searchInput!: ElementRef<HTMLInputElement>;
+
+  protected TranslationConstants = TranslationConstants;
```
1nf0rmagician, line 26: We haven't adhered to the Angular best practice/guideline of using protected for members only meant for the components template yet. Since you are using it here, do you want to consistently introduce it for such members in the components in this pr? 

In this component already there should be several cases like the signals mapped from the service for example 🙂

src/Moryx.Launcher/app/src/app/vertical-module-nav/vertical-module-nav.html
```diff
@@ -0,0 +1,17 @@
...
+  @for (moduleItem of modules(); track moduleItem.route) {
+    <div class="nav-item-wrapper">
+      <a mat-list-item [href]="resolveHref(moduleItem)" [title]="moduleItem.description" [class.active]="moduleItem.route === activeRoute()">
+        <mat-icon matListItemIcon>{{ moduleItem.icon }}</mat-icon>
+        @if (!collapsed()) {
+          <span matListItemTitle>{{ moduleItem.title }}</span>
+        }
+      </a>
+      @if (asWeb(moduleItem); as webItem) {
+        <app-notification-badge [eventStream]="webItem.eventStream ?? ''"></app-notification-badge>
```
1nf0rmagician: As mentioned above you could cover the eventstream is undefinced in the if 🙂

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
```diff
@@ -0,0 +1,34 @@
...
+    autocomplete="off"
+    spellcheck="false"
+  />
+  <button mat-icon-button aria-label="Close" (click)="collapse()" class="close-button" tabindex="-1">
+    <mat-icon>close</mat-icon>
+  </button>
+</div>
+
+<mat-autocomplete #auto="matAutocomplete" (optionSelected)="onOptionSelected($event)" [displayWith]="displayFn">
+  @for (suggestion of suggestions(); track suggestion.text) {
```
1nf0rmagician, line 30: I think if you want you could also add and `@empty` here for a no results message similar to the spotlight search where you did that on the custom list 🤔

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
```diff
@@ -0,0 +1,63 @@
...
+  position: relative;
+}
+
+a[mat-list-item] {
+  border-radius: 4px;
+  position: relative;
+  margin-bottom: 2px;
+
+  &:hover {
+    background-color: rgba($primary, 0.3);
```
1nf0rmagician, line 32: Also here different hover and active foreground/background colours

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
```diff
@@ -0,0 +1,63 @@
+@use "@moryx/ngx-web-framework/styles/variables" as *;
+
+:host {
```
1nf0rmagician, line 3: What is your solution when the modules exceed the available space? We had a request for a "click/touch" rather than a "touch-gesture" scroll solution in the past, that is why the old ui had this scroll buttons

src/Moryx.Launcher/app/src/app/vertical-module-nav/vertical-module-nav.ts
```diff
@@ -0,0 +1,44 @@
...
+@Component({
+  selector: 'app-vertical-module-nav',
+  imports: [MatListModule, MatIconModule, TranslatePipe, NotificationBadge],
+  templateUrl: './vertical-module-nav.html',
+  styleUrl: './vertical-module-nav.scss',
+  host: {
+    '[class.collapsed]': 'collapsed()',
+  }
+})
+export class VerticalModuleNav {
```
1nf0rmagician: You could use a common base class for the ver4tical and horizontal nav. If I'm not mistaken all of the code in here should be duplicated

src/Moryx.Launcher/app/src/app/layouts/full-layout/full-layout.scss
```diff
@@ -0,0 +1,57 @@
...
+  overflow: hidden;
+}
+
+.main-area {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  min-width: 0;
+  overflow: hidden;
+}
```
1nf0rmagician: Can we check once whether the scrolling of elements in the modules still works without setting width or height explicitly 🙈

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.html
```diff
@@ -0,0 +1,38 @@
+<div class="main-area">
+  <div class="top-bar">
+    @if (showTopRegion()) {
+      <slot name="topRegion" (slotchange)="onTopRegionSlotChange($event)"></slot>
+    } @else {
+      <slot name="topRegion" (slotchange)="onTopRegionSlotChange($event)" hidden></slot>
```
1nf0rmagician: having the slots for the regions still in the dom, only hidden, does this leave an empty space there or does the flex layout make the main part grow over the elements?

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.html
```diff
@@ -0,0 +1,38 @@
...
+    }
+  </div>
+
+  <div class="content">
+    <slot name="app"></slot>
+  </div>
+
+  <div class="floating-toolbar" [class.visible]="showExitButton()">
```
1nf0rmagician, line 14: For my understanding is there a reason why you made this visible/invisible over having it not in the dom at all? 🤔

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.scss
```diff
@@ -0,0 +1,50 @@
...
+  overflow: hidden;
+}
+
+.main-area {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  min-width: 0;
+  overflow: hidden;
+}
```
1nf0rmagician: Want to extract at least this and maybe others into a shared scss file next to the layout base?

src/Moryx.Launcher/app/src/app/layouts/_layout.scss
```diff
@@ -0,0 +1,9 @@
+.right-region {
+  display: flex;
+  flex-direction: column;
+  min-width: 200px;
+  position: relative;
+  z-index: 1;
+  border-left: 1px solid var(--mat-sys-outline-variant);
+  box-shadow: -1px 0 2px color-mix(in srgb, var(--mat-sys-shadow) 6%, transparent);
```
1nf0rmagician: I think the border and shadow are different from the top bar in the full layout. Would it make sense/look good if they were the same?

src/Moryx.Launcher/app/src/app/layouts/layout-base.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  onTopRegionSlotChange(event: Event): void {
+    const slot = event.target as HTMLSlotElement;
+    this.launcherStateService.topRegionAvailable.set(slot.assignedNodes().length > 0);
+  }
+
+  onRightRegionSlotChange(event: Event): void {
+    const slot = event.target as HTMLSlotElement;
+    this.launcherStateService.rightRegionAvailable.set(slot.assignedNodes().length > 0);
+  }
```
1nf0rmagician, line 28: Maybe if we are discussing other points anyway you could quickly explain to me what this does 🙈

src/Moryx.Launcher/app/src/app/more-menu/more-menu.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  @ViewChild('appMenu') appMenu!: MatMenu;
+  protected readonly LauncherLayout = LauncherLayout;
+
+  setLayout(layout: LauncherLayout) {
+    this.launcherStateService.updateLayout(layout);
+  }
+
+  openAbout() {
+    this.dialog.open(AboutDialog);
```
1nf0rmagician: There are material design guidelines for dialog size which are not the angular material standard I think. 

https://m3.material.io/components/dialogs/specs#9a8c226b-19fa-4d6b-894e-e7d5ca9203e8

1nf0rmagician (changes requested):

src/Moryx.Resources.Web/app/src/app/app.scss
dbeuchler: because of the image  it stays the same in dark-mode and white is the only option here. changed that from white to the var in the angular-22 pr but reverted here do be compatible with dark-mode.

src/Moryx.Resources.Web/app/src/app/components/resource-tree/resource-tree.scss
dbeuchler: in a next PR in ngx-web-framework I want to introduce a --moryx-selected color, which is also changing on dark-mode. Then we can use primary-light again and in dark-mode only primary.

src/Moryx.Products.Web/app/src/environments/environment.ts
dbeuchler: thanks - I often switch between the servers (have a full machine running on a container) and sometimes that happens

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-recipes/product-recipes.html
dbeuchler: Yes, thats true, but Im going to leave it as is for now. Cases like that appear very frequently in the code. I updated the variables using auto-rename to reduce redundancy and use shorter, clearer names.

src/Moryx.Resources.Web/app/src/app/components/resource-tree/resource-tree.scss
dbeuchler: Same as in other components like the tab-nav too.. Is good for light and Dark-Mode
Screenshot 2026-06-30 at 07 46 10

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-parts/product-parts.scss
dbeuchler: Same as in other components like the tab-nav too.. Is good for light and Dark-Mode
Screenshot 2026-06-30 at 07 46 10

src/Moryx.Media.Web/app/src/app/components/variant-overview/variant-overview.scss
dbeuchler: yes, they are obsolete - sometimes I use another IDE which introduces them automatically - good catch

src/Moryx.Launcher/Moryx.Launcher.csproj
dbeuchler: Fonts are a bit special and as far as I know the browser-developers have no common implementation here. @font-face declarations inside Shadow DOM are ignored (or handled inconsistently) by browsers. Fonts must be registered at the document scope (Light DOM /  <head>). So Fonts are a global browser resource. Shadow DOM encapsulates styles, but font registration must always happen "outside".

Our modules do that by referencing the styles.css from the module in their razor page. The launcher is doing the same too. So Roboto, Material-Icons are registered already. A short look in the inspector, has confirmed that the icons e.g. are only loaded once.

src/Moryx.Launcher/Pages/Index.cshtml
dbeuchler: No, also discussed with @MathoMathiasCamara that we want to provide an endpoint (maybe minor-conform) in following PRs to more and more get rid of the razor-mix.

src/Moryx.Launcher/Models/CultureModel.cs
dbeuchler: I also thought about that but to keep the changes in launcher limited, I don't touched them. After your comment now, I moved them to the folder but do not provide the namespace anymore since it would be a breaking change for the Moryx.Launcher-project. Added a TODO on each definition.

src/Moryx.Launcher/app/package.json
dbeuchler: oh, this happened because the launcher app was created completely from scratch. I will harmonize them over all projects in a separate commit

src/Moryx.Launcher/Pages/Shared/_Layout.cshtml
dbeuchler: comes from the launcher-app project - so only styles are necessary - it contains all definitions

src/Moryx.Products.Web/app/src/app/components/products-details-view/product-recipes/product-recipes-details/product-recipes-details.scss
dbeuchler: no difference if I add or remove it

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.html
dbeuchler: Its for the slide-in/slide-out animation. The toolbar slides up from below the viewport when showExitButton() becomes true. Using visibility + transform with CSS transitions allows the enter/exit animation.

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.scss
dbeuchler: extracted, the host too..

src/Moryx.Launcher/app/src/app/layouts/layout-base.ts
dbeuchler: simple: this is how you chan check if the slot is assigned in the html or not. Region slots are not added in the razor if they are not configured.

src/Moryx.Launcher/app/src/app/more-menu/more-menu.ts
dbeuchler: what's wrong with this line?

src/Moryx.Launcher/app/src/app/layouts/_layout.scss
dbeuchler: The borders match, just different sides. But the box-shadow values differ - the toolbar has a two-layer shadow while .right-region has a single shadow -> harmonized them.

src/Moryx.Launcher/app/package.json
dbeuchler: as discussed each module can define its own fonts by importing them in the styles.scss ans reference them in the razor styles section as before. There is no need to explicitly keep material-icons sind its 3mb on every page load.

src/Moryx.Launcher/app/src/styles.scss
dbeuchler: This is correct for Angular Material 3. The M3 theme system automatically derives the secondary color from the palettes merged into $_rest. Since secondary is included in $_rest, and $_rest is merged into both $primary-palette and $tertiary-palette, the secondary palette is available to the theme.

src/Moryx.Launcher/app/src/environments/environment.ts
dbeuchler: yes the launcher is an exception for that because you cannot run the launcher angular project without any module for a full test. The easiest way is to run the StartProject.Asp and run the launcher as watch. But you are right, this is uncommon for our web projects and therefore I introduced a watch:prod for the launcher which uses the prod config for full tests. Changed the environments to the defaults we use.

src/Moryx.Launcher/app/src/app/app.ts
dbeuchler: input.required() dos not mean the value must be present at element creation time. It means Angular will throw if the component tries to read the signal before it is been set. Since the <script> block runs synchronously right after the element is in the DOM, the properties are assigned before Angular bootstraps

src/Moryx.Launcher/app/src/app/app.ts
dbeuchler: .set() is a write, it does not register a read dependency, so wrapping it in untracked() has no effect on this dependency tracking. The effect  will only re-run based on the input signal reads (this.webModuleItems(), etc.), not because of the .set() calls.

If someone later adds a read inside the callback (e.g., reading moduleService.modules(), untracked would prevent accidentally creating an unwanted dependency or feedback loop.

I have splitted the effects and added untracked to be more defensive.

src/Moryx.Launcher/app/src/app/app.config.ts
dbeuchler: mhmm yes I don't have removed them as in other web projects too for maybe adding routes to the launcher. But if they are required we can add them. Removed.

src/Moryx.Launcher/app/src/app/services/module.service.ts
dbeuchler: no, because of readability. First if checks the char, second if checks the route. So I would like to keep it separate.

src/Moryx.Launcher/app/src/app/services/shortcut.service.ts
dbeuchler: nop, for now it is only for the launcher. But interesting idea for later (service of the moryx shell) :-)

src/Moryx.Launcher/app/src/app/services/theme.service.ts
dbeuchler: no I don't want to mix that. LauncherStateService manages layout/nav/region state. ThemeService manages theme mode and applies CSS classes to the document root. Combining them just because both use localStorage would be coupling by implementation detail, not by domain.

I used that comment as a trigger for renaming the LauncherStateService to LauncherLayoutService.

src/Moryx.Launcher/app/src/app/services/theme.service.ts
dbeuchler: ok

src/Moryx.Launcher/app/src/app/about-dialog/about-dialog.ts
dbeuchler: > If my assumption that you plan to populate the service from nested applications is correct,

as already said, good idea but was not my intention at the first point. But we should follow the idea in further PRs. I moved the label to the ShortcutBinding and just read them in the about-dialog.

src/Moryx.Launcher/app/src/app/navigation/horizontal-module-nav/horizontal-module-nav.scss
dbeuchler: not for our color. Therefore I used 0,3 of primary. We need moryx-selected and others in ngx-web if we want to generalize that.

src/Moryx.Launcher/app/src/app/navigation/module-grid-menu/module-grid-menu.scss
dbeuchler: intended, other hovers or selections are on a white background. The module grid uses the default material background for drop-downs. That would not look good.

src/Moryx.Launcher/app/src/app/module-overview/module-overview.scss
dbeuchler: rgba($primary, 0.3); is correct. A border is used here to restore the card style as before with tw (Index.cshtml)

src/Moryx.Launcher/app/src/app/module-overview/module-overview.ts
dbeuchler: because it is a third web component which is hosted in the Index.cshtml to provide the module-overview. Its not used inside the launcher, it is a virtual module you can say.

src/Moryx.Launcher/app/src/app/services/auth.service.ts
dbeuchler: moved the code from the dev. No changes made by me -> would introduce that in a followup pr

src/Moryx.Launcher/app/src/app/services/culture.service.ts
dbeuchler: moved the code from the dev. No changes made by me -> would introduce that in a followup pr

src/Moryx.Launcher/app/src/app/spotlight-search/spotlight-search.scss
dbeuchler: always rgba($primary, 0.3)..  this is not hovered if you use your keyboard for selection. Renamed to selected.

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
dbeuchler: is there a guideline? Im a big fan of access modifiers but I am more a fan of consistency ;-) I think removing protected from TanslationConstants would be consistent for now?

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
dbeuchler: what do you mean? Its always rgba($primary, 0.3)

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
dbeuchler: The normal ui behavior is a scrollbar. The old ui had nothing.

Old: 
Screenshot 2026-06-30 at 15 09 51

New:
Screenshot 2026-06-30 at 15 10 42

src/Moryx.Launcher/app/src/app/vertical-module-nav/vertical-module-nav.ts
dbeuchler: At the first view it looks like that but its only property definitions and type check. From my pov it would only over-complicate it with a base class. The horizontal-modules is doing the resize thing...

src/Moryx.Launcher/app/src/app/layouts/full-layout/full-layout.scss
dbeuchler: I cannot find any issues.

src/Moryx.Launcher/app/src/app/layouts/fullscreen-layout/fullscreen-layout.html
dbeuchler: > The hidden [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes) is an [enumerated](https://developer.mozilla.org/en-US/docs/Glossary/Enumerated) attribute indicating that the browser should not render the contents of the element. For example, it can be used to hide elements of the page that can't be used until the login process has been completed.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden

src/Moryx.Launcher/.gitignore
dbeuchler: every asset is added from the angular build. Therefore I ignored the whole wwwroot.

src/Moryx.Launcher/app/src/assets/favicon.ico
dbeuchler: the assets are copied to wwwroot/assets by the angular build process and the favicon is referenced from there, or do you mean something different?

src/Moryx.Launcher/app/src/styles.scss
1nf0rmagician: That is wild, thanks for the clarification 😅

src/Moryx.Launcher/app/src/app/services/module.service.ts
1nf0rmagician: Ahh ok I just didn't read it that way, I was only confused whether there shoudl be an else somewhere. But alright to keep it 👍

src/Moryx.Launcher/app/src/app/services/shortcut.service.ts
1nf0rmagician: I was a bit confused because I didn't see why you would handover the destroyref from service to service. I don't think I like that very much especially because the remote registration of the shortcuts only happens from two other services. 

I see that it is somewhat nicely splitted to have the service managing the shortcuts and the other services defining them, but it also means having to gather the information about the shortcuts from these places...

Let's leave it for now and if the service is somehow used by the modules itself it would need to be defined this way anyway 🙂

src/Moryx.Launcher/app/src/app/services/theme.service.ts
1nf0rmagician: Ok if we split LauncherLayout and LauncherTheme I'm fine with that. I would argue both are part of the LauncherState though and hence could be grouped by the domain 😅

src/Moryx.Launcher/app/src/app/module-overview/module-overview.ts
1nf0rmagician: Ahh I didn't get that from just reading the PR 🙈

src/Moryx.Launcher/app/src/app/services/culture.service.ts
1nf0rmagician: I was not awara that our culture cookie resets once a year 😬😅

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
1nf0rmagician: Yip there is a Style guide by angular for exactly that point, but I only stumbled upon this specific point when I read through the 22 migration guide. Hence it is not applied in most modules yet 🙈

https://angular.dev/style-guide#use-protected-on-class-members-that-are-only-used-by-a-components-template

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
1nf0rmagician: Sorry maybe that was a part where I got a bit tired 🙈 I was under the impression that text and background colors where not always applied consistently. Must have misread

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
1nf0rmagician: Whaaat 👀 @MathoMathiasCamara do you know when the scroll-buttons where removed?

src/Moryx.Launcher/app/src/app/vertical-module-nav/vertical-module-nav.ts
1nf0rmagician: I mean I don't have a stronmg opinion, just thought the vertical nav woudl basically be empty then 😅

src/Moryx.Launcher/app/src/app/layouts/layout-base.ts
1nf0rmagician: Ahh now I get it...

src/Moryx.Launcher/app/src/app/more-menu/more-menu.ts
1nf0rmagician: My (maybe outdated) information is that opening an angular material dialog does not limit its width to the material desgin limits of 
<html>
<body>

Container width | Min 280dp; Max 560dp

</body>
</html>

Could you check that?

src/Moryx.Launcher/app/src/app/app.scss
dbeuchler: The launcher uses ViewEncapsulation.ShadowDom. Shadow DOM creates an isolated DOM subtree with its own style scope; global styles from the document (classes defined in styles.scss) don't go into the shadow dom.

So the .material-symbols-outlined is needed here to re-declare the font-family inside the shadow DOM, otherwise <mat-icon> elements within this component tree wouldn't resolve to the correct icon font.

I will write a comment to that

src/Moryx.Launcher/app/src/app/services/culture.service.ts
dbeuchler: I tried to change that to a method but the computed signal feels better here because it provides a assignable reference that can be passed around (like currentCulture = this.cultureService.currentCulture in more-menu) without losing this context; something a plain method can't do without an extra wrapper. In later PRs we could also support language updates without hard-reload.

src/Moryx.Launcher/app/src/app/services/launcher-state.service.ts
dbeuchler: Also thought about that but localStorage is synchronous and fast. It is already an in-memory lookup with persistence. The multiple getStoredState() calls during initialization each parse the same JSON string, but this only happens once at service construction. The cost is trivial. Caching it would be a micro-optimization with no measurable benefit. Adding a local field or something similar would only make it more complicated. So reading cookies or reading for the local storage is like reading from memory.

src/Moryx.Launcher/app/src/app/spotlight-search/spotlight-search.html
dbeuchler: yes but it is designed as a dropdown overlay attached to a matInput inside a mat-form-field. The autocomplete would render its own overlay/dropdown below the input -> mat-autocomplete is designed to fill the input value on selection, not navigate

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
dbeuchler: good point - done

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
dbeuchler: mat-option manages its own click handling internally, placing an `<a>` inside it would cause conflicts. And you can't replace mat-option with a custom anchor element because mat-autocomplete requires mat-option children.  open_in_new icon already signals that the option will navigate

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
dbeuchler: harmonized

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.ts
dbeuchler: Added here https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/1061 to make it consistent, thanks

src/Moryx.Launcher/app/src/app/toolbar-search/toolbar-search.html
dbeuchler: good idea, done

src/Moryx.Launcher/app/src/app/navigation/vertical-module-nav/vertical-module-nav.scss
dbeuchler: So if we want to reintroduce, we can do that in follow-up PRs ;-)

src/Moryx.Launcher/app/src/app/more-menu/more-menu.ts
dbeuchler: added Material Design container limits: min 280px, max 560px (falling back to 92vw on very small screens).

src/Moryx.Launcher/app/src/app/horizontal-module-nav/horizontal-module-nav.ts
dbeuchler: upps, implemented before more button 👍  -> changed

src/Moryx.Launcher/app/src/app/vertical-module-nav/vertical-module-nav.ts
dbeuchler: convinced

src/Moryx.Workplans.Web/app/src/app/app.scss
```diff
@@ -2,58 +2,71 @@
...
+  display: flex;
+  flex-flow: column nowrap;
+  align-items: stretch;
+  background-color: var(--mat-sys-surface-container-lowest);
+
+  mat-toolbar {
+    color: white;
+
+    button {
+      color: white;
```
1nf0rmagician, line 26: Did you intend to set it to white and not a material color? I think the resource and products sidebars are already adjusted to the dark theme 😬

dbeuchler: yes, same argument why resources and products also use white as hard-value: They use this background-image and it does not matter if light or dark mode, the image is the same and only white color matches here.

1nf0rmagician (approved):

PR #1291, opened 2026-06-29
Tweaks in ValueProviderExecutor and config providers

andreniggemann:
Optimize ValueProviderExecutor to not iterate over chars in string

Additionally fix a bug with the ValueProviders created by configurationIntegration = true.
Due to the ActivationProvider being first lists wouldn't be null and thus not created by the MicrosoftExtensionConfigProvider
even if keys for that List exist

### DIFF
--- a/docs/articles/framework/configuration.md
+++ b/docs/articles/framework/configuration.md
@@ -142,6 +142,17 @@ or alternatively
 }
 ```
 
+> [!CAUTION]
+> The collection integration works **if and only if** the list is null or if all the needed entries in the list already exist and can be modified.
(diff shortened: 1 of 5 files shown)


dbeuchler (commented):
Is it worth to extend the tests?

andreniggemann:
Good points. I could for the part with the order changes. 
For the part with strings I guess it would be hard, because it's an internal optimization only.
If you have an idea how to test the changes to the ValueProviderExecutor, please let me know.

dbeuchler (approved):

PR #1297, opened 2026-07-02
Add custom serialization for visual instruction to support all EntrySerialize attribute types in Inputs

MathoMathiasCamara:
**Previously** Any custom PossibleValuesAttribute that needed access to local or general container threw an exception because they were null.

**Now** added a new VisualInstructionsSerialization that take value of the local and general container.


src/Moryx.VisualInstructions.Endpoints/VisualInstructionsSerialization.cs
1nf0rmagician, line 1: Since this is actually a 1:1 copy of the code in the resources we shoulf find a spot were we can reuse it. There is the `Moryx.AspNetCore` package which could be used I think. Maybe @dbeuchler has another idea?

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
```diff
@@ -31,8 +32,11 @@ public class VisualInstructionsController : ControllerBase
         Converters = { new JsonStringEnumConverter() }
     };
 
-    public VisualInstructionsController(IVisualInstructions visualInstructions)
-        => _visualInstructions = visualInstructions;
+    public VisualInstructionsController(IVisualInstructions visualInstructions, IModuleManager moduleManager, IServiceProvider serviceProvider)
+    {
+        _visualInstructions = visualInstructions;
+        Converter._serialization = new VisualInstructionsSerialization(moduleManager.AllModules.FirstOrDefault(module => module is IFacadeContainer<IVisualInstructions>)?.Container, serviceProvider);
```
1nf0rmagician: Instead of overriding it you could correctly initialize it 🙈 The default initializiation [here](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/1552a70ba8257a343101cf0f88403ae7ad637177/src/Moryx.VisualInstructions.Endpoints/Models/Converter.cs#L12) does not work of course. And maybe you could also find out why we need this member filter at all and cannot just use the `PossibleValuesSerialization` base class 🤔

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsSerialization.cs
dbeuchler: `Moryx.AspNetCore` does not have any reference to Moryx.Runtime and is more a core-core project. For me the implementation is somehow the same but all are one-liner which can be made more specific in the future.

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
dbeuchler: I also do not like the override of the static. Same question from me: Why not use the PossibleValuesSerialization directly?

MathoMathiasCamara: Mmm that's a good question. I will use the PossibleValueSerialization

dbeuchler (dismissed):
i dont like this FirstOrDefault to find the module but I have currently also no better idea. We should think about MORYX 12 how this could be solved.

dbeuchler (approved):

PR #1304, opened 2026-07-06
Fixed cancellationToken naming

dbeuchler:


src/Moryx.Operators.Endpoints/OperatorManagementController.cs
```diff
@@ -205,16 +205,16 @@ public ActionResult SignOut(string operatorIdentifier, long resourceId)
     /// </item>
     /// </list>
     /// </summary>
-    /// <param name="token"></param>
+    /// <param name="cancellationToken"></param>
     /// <returns></returns>
     [HttpGet("stream")]
-    public IResult Stream(CancellationToken token)
+    public IResult Stream(CancellationToken cancellationToken)
```
dbeuchler, line 211: @1nf0rmagician @andreniggemann here is one change were I need your "go". Is changing the argument name. This is not a problem from semantic versioning view: 
1. Parameter names on CancellationToken parameters are irrelevant - ASP.NET Core binds the cancellation token from the request automatically, not from query string or route values. No client ever sends this parameter by name.                                                 
  2. Even if someone called this method directly in C# using named arguments (Stream(token: cts.Token)), controller actions are not typically invoked that way. Its not a library API consumers call directly.                                                         
  3. CancellationToken parameters are excluded from the generated API spec by default.

It would only be a problem if someone would use reflection to find an argument with the name. So for me I would say we should neglect that.

andreniggemann: I agree that this is negligable

andreniggemann (approved):

1nf0rmagician (approved):

PR #1306, opened 2026-07-06
Consistent naming of StartUpHooks and code improvements

dbeuchler:
Improve StartupHook base classes and ModelStartupHook for #1202                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                       
  Changes                                                                                                                                                                                                        

  - Fix deadlock in ModuleStartupHook.RunAsync: The event handler used Task.Run(...).GetAwaiter().GetResult(), which synchronously blocked the calling thread waiting for async work to complete. Replaced with a fire-and-forget using a separate async Task method.
  - The ModuleStateChanged event handler is now unsubscribed when the cancellation token is triggered.
  - Use CancellationToken, previously the parameter was accepted but never used.
  - Changed Priority to get-only / init-only to prevent mutation after initialization, matching the IStartupHook interface contract.
  - Replace protected fields with properties. Protected fields expose internal state without any encapsulation; properties allow subclasses to access the values while preserving the ability to other behavior in the future without breaking the API. They also follow C# naming and design conventions (CA1051).
  - Guard against nitializationResult.Error being null when logging initialization failures.
  - Fix typos


src/Moryx.Runtime/Modules/StartupHooks/ModuleStartupStartHookBase.cs
```diff
@@ -4,29 +4,28 @@
...
 
 /// <summary>
-/// Base class for StarupHooks that react to module starts.
+/// Base class for StartupHooks that react to module starts.
 /// </summary>
 /// <typeparam name="TFacade">The facade of the Module. Is used to select the correct module</typeparam>
 /// <typeparam name="TConfig">The type of config this hook requires</typeparam>
-public abstract class ModuleStartHook<TFacade, TConfig>(IModuleManager moduleManager, IConfigManager configManager, ILogger logger)
-    : ModuleHook<TFacade, TConfig>(moduleManager, configManager, logger) where TConfig : ConfigBase, new()
+public abstract class ModuleStartupStartHookBase<TFacade, TConfig>(IModuleManager moduleManager, IConfigManager configManager, ILogger logger)
```
andreniggemann, line 14: I am sorry, but the name ModuleStartupStartHookBase is far from ideal

src/Moryx.Runtime/Modules/StartupHooks/ModuleStartupHookBase.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            else
+            {
+                Logger.LogInformation("Initialization failed: {message}. Not running this hook", InitializationResult.Error?.Message);
+            }
+            return Task.CompletedTask;
+        }
+
+        ModuleManager.ModuleStateChanged += OnModuleStateChanged;
+        cancellationToken.Register(() => ModuleManager.ModuleStateChanged -= OnModuleStateChanged);
```
andreniggemann, line 86: Unregistering here is a bad idea from my perspective.

Let's say somebody set's a timeout on the CancellationToken when calling StartupHookExtensions. RunHooksAsync with the intention to not hang the application indefinitely, while the startup is waiting for other processes to start or database migrations to apply etc.

That does not mean they don't want these events to be registered anymore.

andreniggemann (changes requested):
Thanks for taking the time to work on this. I am not really a huge fan of having StartupHook everywhere, but I would go with it if you insist.
I would really like to change the name of the ModuleStartupStartHookBase class though.

andreniggemann:
> * Fix deadlock in ModuleStartupHook.RunAsync: The event handler used Task.Run(...).GetAwaiter().GetResult(), which synchronously blocked the calling thread waiting for async work to complete. Replaced with a fire-and-forget using a separate async Task method

I don't think there is a dead lock there and it was placed intentionally. It was added while the ResourceHook was still planned to be included, because it would block other modules from starting. 
This was very helpful, because the modules depending on the ResourceManagement would fail, when the ResourceManagement was reincarnated, while they were starting and iterating over the Resources they care about.

This might not be strictly necessary without the ResourceManagement, but it might be a good option to have.

dbeuchler:
> I don't think there is a dead lock there and it was placed intentionally. It was added while the ResourceHook was still planned to be included, because it would block other modules from starting. This was very helpful, because the modules depending on the ResourceManagement would fail, when the ResourceManagement was reincarnated, while they were starting and iterating over the Resources they care about.
> 
> This might not be strictly necessary without the ResourceManagement, but it might be a good option to have.

You would block other modules from starting with hooks?

andreniggemann:
> > I don't think there is a dead lock there and it was placed intentionally. It was added while the ResourceHook was still planned to be included, because it would block other modules from starting. This was very helpful, because the modules depending on the ResourceManagement would fail, when the ResourceManagement was reincarnated, while they were starting and iterating over the Resources they care about.
> > This might not be strictly necessary without the ResourceManagement, but it might be a good option to have.
> 
> You would block other modules from starting with hooks?

Like I said it was at least beneficial in this one case. I could probably come up with other occasions, if I tried hard enough.
As these hooks are specifically helpful in tests, influencing the timing of module starts might be a necessity to test certain scenarios.

Contrived example: I could hook into the ModuleStarting event to stop my MES Adapter from starting correctly so that I can study, if the other components still work as intended and collect data to be published once the adapter becomes ready.

PR #1316, opened 2026-07-09
Merge dev to future

dbeuchler:
only conflicts solved and launchers models moved to Models namespace. Otherwise only conflicts solved.


dbeuchler:
@1nf0rmagician I think my merge base was wrong... will fix that

1nf0rmagician (dismissed):

PR #1334, opened 2026-07-12
Fix several compiler warning groups

1nf0rmagician:
Applies non-behaviour changing fixes for compiler warnings reducing number by ~150.

@dbeuchler, @seveneleven fixing some of the warnings would/could introduce breaking changes (behaviour or source). I checked the changes myself but please verify them attentively before we merge them.


src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
```diff
@@ -160,12 +160,8 @@ protected virtual void ConvertReferenceRecursion(Resource resource, ResourceMode
     private static ICollection<Type> MergeTypeConstraints(PropertyInfo property, Type targetType, IDictionary<string, List<Type>> referenceOverrides)
     {
         // If there are no overrides the only limitation is the target type
-        if (!referenceOverrides.ContainsKey(property.Name))
+        if (!referenceOverrides.TryGetValue(property.Name, out var myOverrides))
             return [targetType];
-
-        // Otherwise find all types that limit the reference type without redundancies. This means eliminating all types
-        // represented by another type
```
dbeuchler, line 167: Deliberately removed?

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
```diff
@@ -185,8 +181,8 @@ internal ResourceTypeModel ConvertType(IResourceTypeNode node, ResourceTypeModel
     {
         var resType = node.ResourceType;
 
-        if (TypeCache.ContainsKey(node.Name))
-            return TypeCache[node.Name];
+        if (TypeCache.TryGetValue(node.Name, out var value))
+            return value;
```
dbeuchler: I don't like `value` because it conflicts somehow with value implicit param on properties... maybe `cached` in this sample?

I started commenting each change with `value` but removed the comments... maybe you search for 'out var value' ;-)

src/Moryx.Orders.Management/Implementation/Pool/OperationDataPool.cs
```diff
@@ -43,7 +43,7 @@ public async Task StartAsync(CancellationToken cancellationToken = default)
         var operationRepo = uow.GetRepository<IOperationEntityRepository>();
         // Restore only operations which are not completed
         var restored = await operationRepo.Linq.Active()
-            .Where(o => o.State < OperationDataStateBase.CompletedKey).ToArrayAsync();
+            .Where(o => o.State < OperationDataStateBase.CompletedKey).ToArrayAsync(cancellationToken: cancellationToken);
```
dbeuchler: Do you have followed just an IDE suggestion? In MORYX 10, the asynchronous APIs were introduced BUT: not fully applied. So just accepting IDE suggestions does not make sense every time, that's the reason it was not finalized in MORYX 10 rework of the APIs. Each call must be cancelled at a defined point in state, where cancellation makes sense. This comment is not regarding exactly this line, but for all lines where the cancellation token is now passed.

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
1nf0rmagician: The comment no, that was part of the ide action that i did not expext 😅

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
1nf0rmagician: It is the IDE default when applying the warning's fix

src/Moryx.Orders.Management/Implementation/Pool/OperationDataPool.cs
1nf0rmagician: This actually only applied the IDE auto-fix for the respective warning. I'm not sure whether it would give any benefit to keep the warnings as a reminder to come back to

> Each call must be cancelled at a defined point in state, where cancellation makes sense

but from my perspective handing over the cancellation token in all places where called methods would accept them is a better state than keeping the warnings. 

If we fully resolve this issue, we would need to check for all methods accepting a token anyway and could not focus on the warnings alone.

So I'm with you this does not really improve the asynchronous APIs we have, it just gets rid of warnings which on the other side did not have a benefit when keeping them either 🙂

dbeuchler: > but from my perspective handing over the cancellation token in all places where called methods would accept them is a better state than keeping the warnings.

Think about a scenario: In product storage you do multiple steps to save a product (save parts, save this, safe that) if you cancel it, the database state is broken if you hand over the cancellation token in every place. This was the reason I explicitly NOT handed over the cancellationToken "just because the IDE says that".

More general: In complex logic, there is often a "point of no return." Once a multi-step process begins writing to the database, you usually want it to finish to avoid partial states, unless you have implemented a full rollback mechanism.

Its totally okay for me to fix that but then you must be aware about the whole call chain and what might can be go wrong. If you want get rid of the warning we could pass `CancellationToken.None`. This drops the IDE warning, documents our current decision, and guarantees the writes won't be interrupted halfway through.

src/Moryx.AbstractionLayer.Resources.Endpoints/Converter/ResourceToModelConverter.cs
dbeuchler: My ide marks "value" and then I can write a good name ;-)

src/Moryx.Orders.Management/Implementation/Pool/OperationDataPool.cs
1nf0rmagician: >Its totally okay for me to fix that but then you must be aware about the whole call chain and what might can be go wrong. If you want get rid of the warning we could pass CancellationToken.None. This drops the IDE warning, documents our current decision, and guarantees the writes won't be interrupted halfway through.

I like that suggestion, will take a look where the IDE suggestion acutally made sense and otherwise hand over `CancellationToken.None`

1nf0rmagician: As it were actually only 4 changes for this specific warning I took the time to think about how to correctly handle the specific cases. Two tokens are still just handed through with the warnings actually reflecting issues in these cases. The other to options should be reviewed again @dbeuchler 🙂

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -24,7 +24,7 @@ namespace Moryx.Drivers.OpcUa;
 
 /// <summary>
 /// Driver to communicate via Opc Ua . It is able to write and read nodes
-/// and subscribe to value changes of nodes in a session
+/// and subscribe to node changes of nodes in a session
```
dbeuchler: value of node changes... so value sounds better here.

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -854,7 +854,7 @@ internal async Task OnWriteNode(OpcUaNode node, object payload, CancellationToke
         {
             if (writeResult.Results.First().Code != 0)
             {
-                Logger.Log(LogLevel.Warning, "There was an error when trying to write a value to node {NodeId}", node.NodeId);
+                Logger.Log(LogLevel.Warning, "There was an error when trying to write a node to node {NodeId}", node.NodeId);
```
dbeuchler: value was better here

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -908,7 +908,7 @@ internal async Task<DataValueResult> OnReadValueOfNode(string identifier, Cancel
     }
 
     /// <summary>
-    /// Write a value to a node using the driver directly
+    /// Write a node to a node using the driver directly
```
dbeuchler: value

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -922,7 +922,7 @@ public void Send(object payload)
         }
 
         var node = State.GetNode(msg.Identifier);
-        const string errorMsg = "When trying to read the value of the node, ";
+        const string errorMsg = "When trying to read the node of the node, ";
```
dbeuchler: value

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -968,7 +968,7 @@ internal string ReadNodeAsString(string nodeId)
             var value = ReadNodeDataValue(nodeId);
             if (value == null)
             {
-                return "There was an error, when trying to read the value of the node. Please look into the log for further information";
+                return "There was an error, when trying to read the node of the node. Please look into the log for further information";
```
dbeuchler: value

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -990,7 +990,7 @@ internal string ReadNodeAsString(string nodeId)
     internal async Task WriteNode(string identifier, string valueString, CancellationToken cancellationToken = default)
     {
         var node = State.GetNode(identifier);
-        var errormsg = "When trying to read the value of the node, ";
+        var errormsg = "When trying to read the node of the node, ";
```
dbeuchler: value

src/Moryx.Drivers.OpcUa/OpcUaDriver.cs
```diff
@@ -1050,7 +1050,7 @@ private object CreateValue(BuiltInType type, string stringValue)
         {
             Logger.Log(
                 LogLevel.Error,
-                "An error occured, when trying to cast the value for the OPC UA node to the corresponding type: {Message}",
+                "An error occured, when trying to cast the node for the OPC UA node to the corresponding type: {Message}",
```
dbeuchler: value?

src/Moryx.Drivers.Simulation/InOutDriver/SimulatedInput.cs
```diff
@@ -11,9 +12,9 @@ namespace Moryx.Drivers.Simulation.InOutDriver;
 public class SimulatedInput : IInput
 {
     /// <summary>
-    /// Direkt access to a single value
+    /// Direkt access to a single indexedValue
```
dbeuchler: value

src/Moryx.Drivers.Simulation/InOutDriver/SimulatedInput.cs
```diff
@@ -32,15 +33,15 @@ public object this[int index]
...
+    public object this[string key] => Values.TryGetValue(key, out var keyedValue) ? keyedValue : default;
 
     /// <summary>
     /// All access types supported
     /// </summary>
     public SupportedAccess Access => SupportedAccess.Single | SupportedAccess.Index | SupportedAccess.Key | SupportedAccess.Event;
 
     /// <summary>
-    /// Internal value dictionary
+    /// Internal singleValue dictionary
```
dbeuchler: value

src/Moryx.Orders.Management/Implementation/Pool/OperationDataPool.cs
dbeuchler: why named? ToArrayAsync only have one parameter

src/Moryx.ProcessData.SpreadsheetsListener/CsvStructure.cs
```diff
@@ -59,7 +59,7 @@ private async Task AdjustHeaderAddRow(IList<string> header, CsvRow row)
             tempHeader = _header.ToList();
 
         var addHeadings = header.Except(tempHeader);
-        if (addHeadings.Count() > 0)
+        if (addHeadings.Any())
```
dbeuchler: header returns IEnumerable, so: Possible multiple enumeration, could you fix that too?

src/Moryx.Products.Management/Implementation/RecipeManagement.cs
```diff
@@ -156,22 +157,25 @@ public async Task<long> SaveWorkplanAsync(Workplan workplan, CancellationToken c
...
+        // No changes are executed if cancelled
+        cancellationToken.ThrowIfCancellationRequested();
+
         foreach (var recipe in affectedRecipes)
         {
             recipe.Workplan = entity;
         }
 
-        uow.SaveChanges();
+        await uow.SaveChangesAsync(CancellationToken.None);
```
dbeuchler, line 172: handled very well 👍

src/Moryx.Resources.Management/Resources/ResourceGraph.cs
```diff
@@ -121,7 +121,7 @@ public Resource Instantiate(string type)
         var template = new ResourceEntityAccessor { Type = type };
         var instance = template.Instantiate(TypeController, this);
 
-        // Initially set name to value of DisplayNameAttribute if available
+        // Initially set name to resource of DisplayNameAttribute if available
```
dbeuchler: value

src/Moryx.Runtime.Kernel/Configuration/ValueProvider/MicrosoftExtensionsConfigProvider.cs
```diff
@@ -54,7 +54,7 @@ public ValueProviderResult Handle(Stack<ExecutorLevel> levels)
 
         var currentValue = prop.GetValue(config);
         var defaultValue = GetDefaultValue(prop.PropertyType);
-        if (!Equals(currentValue, defaultValue)) // value is already assigned
+        if (!Equals(currentValue, defaultValue)) // index is already assigned
```
dbeuchler: value

src/Moryx.Runtime.Kernel/Modules/Components/ModuleDependencyManager.cs
```diff
@@ -48,7 +48,7 @@ public IModuleDependencyTree GetDependencyTree()
     /// </summary>
     public IModuleDependency GetDependencyBranch(IServerModule plugin)
     {
-        return _cache.ContainsKey(plugin) ? _cache[plugin] : null;
+        return _cache.TryGetValue(plugin, out var pluginsDependencies) ? pluginsDependencies : null;
```
dbeuchler, line 51: I love that you or your agent kept the name `plugin` which was Marvin 0.x - we should keep this for historic throwbacks 🤟

src/Moryx.Runtime.Kernel/Modules/Components/ModuleDependencyManager.cs
```diff
@@ -219,7 +219,7 @@ private IReadOnlyList<IServerModule> StartDependencies(IServerModule module, IDi
 
             // when there is no facade provider for the dependency,
             // add a missing dependency for the property type
-            if (dependencyProviders.Count() == 0)
+            if (!dependencyProviders.Any())
```
dbeuchler: maybe you could also fix the `Possible multiple enumeration` here....

src/Moryx.Runtime.Kernel/Modules/Components/ModuleStarter.cs
```diff
@@ -115,13 +115,13 @@ private async Task ModuleChangedState(IServerModule module, ServerModuleState ne
         // Now we start every service waiting on this service to return
         await _waitingModulesSemaphore.ExecuteAsync(async () =>
         {
-            if (!WaitingModules.TryGetValue(module, out var value))
+            if (!WaitingModules.TryGetValue(module, out var previouslyWaitingModule))
```
dbeuchler: `waitingModules` -> its a collection

src/Tests/Moryx.Resources.Management.Tests/ResourceEntityAccessorTests.cs
```diff
@@ -149,9 +149,9 @@ public async Task SaveEntityReturnsAValidResource(bool createNew)
 
     private class ExtensionDataTestBase
     {
-        public int Value1 => 1;
+        public static int Value1 => 1;
 
-        public string Value2 => "MyVal";
+        public static string Value2 => "MyVal";
     }
```
dbeuchler, line 157: In this case I would keep the non-static because the test checks if it is restored correctly, Newtonsoft is using a lot of reflection and maybe the behavior is different and do not match the test case anymore.

src/Tests/Moryx.Tests/Configuration/Transformer/TransformerTestConfig.cs
```diff
@@ -95,7 +95,7 @@ public TransformerTestConfig()
     /// <summary>
     /// Gets the read only property.
     /// </summary>
-    public string ReadOnlyProperty => "Test";
+    public static string ReadOnlyProperty => "Test";
```
dbeuchler: same here, test-case, keep non-static

src/Tests/Moryx.Tests/Serialization/EntrySerializeDummies.cs
```diff
@@ -49,7 +49,7 @@ public class EntrySerialize_NotClassMixed
     [EntrySerialize(EntrySerializeMode.Never)]
     public string NeverMethod1() => "1234";
 
-    private bool NullMethod1() => true;
+    private static bool NullMethod1() => true;
```
dbeuchler: same here, test-case, keep non-static

src/Tests/Moryx.Tests/Serialization/EntrySerializeDummies.cs
```diff
@@ -143,7 +143,7 @@ public void InvocablePublic(int intValue, string stringValue1, string stringValu
     }
 
     [EntrySerialize]
-    internal void InvocableInternal()
+    internal static void InvocableInternal()
```
dbeuchler: same here, test-case, keep non-static, maybe add static as another test-case

src/Tests/Moryx.Tests/Serialization/EntrySerializeDummies.cs
```diff
@@ -155,7 +155,7 @@ protected void NonInvocableProtected()
     }
 
     [EntrySerialize]
-    private void NonInvocablePrivate()
+    private static void NonInvocablePrivate()
```
dbeuchler: same here, test-case, keep non-static, maybe add static as another test-case

src/Tests/Moryx.VisualInstructions.Endpoints.Tests/VisualInstructionsFacadeMock.cs
```diff
@@ -43,9 +43,8 @@ public void CompleteInstruction(string identifier, long instructionId, string re
     }
 
     public IReadOnlyList<ActiveInstruction> GetInstructions(string identifier)
-        => _instructions.ContainsKey(identifier)
-            ? _instructions[identifier]
-            : Array.Empty<ActiveInstruction>();
+        => _instructions.TryGetValue(identifier, out var activeInstruction)
+            ? activeInstruction : Array.Empty<ActiveInstruction>();
```
dbeuchler: collection initializer? `[]`?

src/Moryx.Runtime.Kernel/Modules/Components/ModuleDependencyManager.cs
1nf0rmagician: This was actually me but i didn't realize thge mismatch to the method parameter, will keep it for now xD

src/Moryx.Benchmarking/BenchmarkParameters.cs
```diff
@@ -36,6 +37,6 @@ protected override void Populate(Process process, Parameters instance)
     {
         var parameters = (BenchmarkParameters)instance;
         parameters.Step = Step;
-        parameters.Instructions = [string.Format(Instructions[0].Content, Step).AsInstruction()];
+        parameters.Instructions = [string.Format(CultureInfo.InvariantCulture, Instructions[0].Content, Step).AsInstruction()];
```
dbeuchler: CurrentCulture? I know its int but for user

src/Moryx.ControlSystem.Simulator/ModuleController/ModuleConfig.cs
```diff
@@ -57,9 +58,9 @@ public class ExecutionTimeDefinition
 
     public override string ToString()
     {
-        var cellString = CellId == 0 ? "" : CellId.ToString() + ": ";
+        var cellString = CellId == 0 ? "" : CellId.ToString(CultureInfo.InvariantCulture) + ": ";
```
dbeuchler: CurrentCulture? I know its int but for "user"

src/Moryx.ControlSystem/Jobs/JobSchedulerBase.cs
```diff
@@ -56,19 +56,18 @@ public virtual void Stop()
     /// <returns></returns>
     protected IReadOnlyList<Job> Dependencies(Job target)
     {
-        return _dependencies.ContainsKey(target)
-            ? (IReadOnlyList<Job>)_dependencies[target]
-            : EmptyDependencies;
+        return _dependencies.TryGetValue(target, out var targetsDependencies)
+            ? (IReadOnlyList<Job>)targetsDependencies : EmptyDependencies;
```
dbeuchler: maybe [] instead of EmptyDependencies? Not really part of the PR but when you touch the place, its an easy change?

src/Moryx.FactoryMonitor.Endpoints/FactoryMonitorController.cs
```diff
@@ -390,7 +391,7 @@ await _resourceManager.ModifyUnsafeAsync(manufacturingConfig.Id, r =>
     public ActionResult<Dictionary<string, CellPropertySettings>> GetCellPropertiesSettings(string identifier)
     {
         var cellLocation = _resourceManager.GetResources<IMachineLocation>()?
-            .FirstOrDefault(x => x.Machine?.Id.ToString() == identifier);
+            .FirstOrDefault(x => x.Machine?.Id.ToString(CultureInfo.InvariantCulture) == identifier);
```
dbeuchler: for user -> CurrentCulture

src/Moryx.AspNetCore.Mqtt/Endpoints/MqttEndpointContext.cs
```diff
@@ -71,17 +71,17 @@ public class MqttEndpointContext(MqttApplicationMessage request, RouteValueDicti
     }
 
     /// <summary>
-    /// Given a <paramref name="parameterName"/>, returns the matching value from the current <see cref="RequestMessage"/> topic.
+    /// Given a <paramref name="parameterName"/>, returns the matching parameterValue from the current <see cref="RequestMessage"/> topic.
```
dbeuchler: Keep comments

dbeuchler (approved):

PR #1330, opened 2026-07-10
Replace ngOnInit by appropriate signal handling

dbeuchler:
for #1061

Migrated all `BehaviorSubject`, `Subject`, and `ReplaySubject` patterns to Angular signals to replace ngOnInit by appropriate signal handling with `effect`. Replaced Observable-based patterns (`toSignal`, `.subscribe()`) with direct signal references, `computed()`, and `effect()`. 

- Replaced all `.update(_ => value)` calls (where the previous value is unused) with `.set(value)` 
- Standardized all `effect()` and `untracked()` calls to multi-line block.
- Removed `ngOnInit`, `ngOnDestroy`, `OnInit`, `OnDestroy`, and `Subscription` where effects or signals made them unnecessary.

PR #1335, opened 2026-07-12
Use promise for API calls and remove custom extensions

dbeuchler:
for #1061
based on #1330

The ng-openapi-gen configuration has been switched to `"promises": true` across all API services. This aligns the generated code with how it was already consumed. The most calls was wrapping the Observable in `lastValueFrom()` or `.toAsync()` anyway.

The custom `observable.extensions.ts` files (which added `.toAsync()` to `Observable`) have been deleted.

Unnecessary `async/await` wrappers on single-statement `.then()/.catch()` callbacks have been cleane. Same idea as avoiding redundant state machines in C#.


src/Moryx.FactoryMonitor.Web/app/src/app/api/models/internal-operation-classification.ts
```diff
@@ -8,11 +8,13 @@
 
 export enum InternalOperationClassification {
   Initial = 'Initial',
+  Assigning = 'Assigning',
```
1nf0rmagician, line 11: Waoh... how did we never update this...

1nf0rmagician (approved):

PR #1342, opened 2026-07-16
Dialog context is now a component, that is used by different dialogs.

ArthurPxC:
**Dialog Context UI**

1) Extracted dialog context UI clones from various order related dialogs into an own component.
2) Order and product information is now displayed inside separate rows, which allows displaying longer product names and order numbers.
3) Order and product values are now shown with ellipses, when they are still too long. 

image

**Begin Button Text**

1) The Begin Button now shows the text "Change Amount" in the states Running and Amount Reached. In all other states it contains the text "Begin". This is also true for the Title and the Buttons inside the Dialog, which opens, when the button is pressed.

image


src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
```diff
@@ -0,0 +1,30 @@
+import { DestroyRef, Directive, ElementRef, inject, signal } from "@angular/core";
```
dbeuchler: missing copyright

src/Moryx.Orders.Web/app/src/app/components/dialog-context/dialog-context.html
```diff
@@ -0,0 +1,34 @@
...
+
+      @let orderValue = model.order + ' - ' + model.number;
+
+      <span class="info-value"
+            appOverflowDetector
+            #overflowDetectorRef="appOverflowDetector"
+            [matTooltip]="orderValue"
+            [matTooltipDisabled]="!overflowDetectorRef.overflowing()"
+            matTooltipPosition="below">
+        {{ orderValue }}
```
dbeuchler: format -> keep in same line until 120 chars are reached

src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    update();
+
+    const observer = new ResizeObserver(update);
+    observer.observe(element);
+
+    this.destroyRef.onDestroy(() => {
+      observer.disconnect();
+    });
+  }
```
dbeuchler: why not using css elipses instead of this whole detector?

dbeuchler (commented):
Lots of conflicts in this branch

src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
ArthurPxC: I use css ellipses, the detector is only there to control the tooltip visibility.

src/Moryx.Orders.Web/app/src/app/components/dialog-context/dialog-context.ts
```diff
@@ -0,0 +1,29 @@
...
+  imports: [
+    TranslatePipe,
+    CommonModule,
+    MatTooltip,
+    OverflowDetector
+  ]
+})
+export class DialogContext {
+  protected TranslationConstants = TranslationConstants;
+  public operationModel = input<OperationModel>();
```
dbeuchler: why not using required? Then the whole If in the template can be removed

src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
dbeuchler: could you describe that?

ArthurPxC: Yes, a common pattern from other UI frameworks I worked on was to allow the user to read the whole text of text blocks which display ellipses because of overflowing text.

For this, the whole text is displayed in a tooltip, when the mouse hovers over the overflowing text block. Because tooltips that have the same content as the text block do not make any sense and distract the user, when there is no overflowing of text, these tooltips are not there, when there is no overflowing.

Exactly this behavior is realized by the directive. No overflowing => no tooltip, overflowing => show whole text in tooltip on hovering, so that the user sees, what is "behind" the ellipses of the span.

dbeuchler: > Because tooltips that have the same content as the text block 

Are you sure? The text is always the full text but the three dots are added by css to the control and not to the tooltip.

src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Licensed under the Apache License, Version 2.0
+*/
+
+import { DestroyRef, Directive, ElementRef, inject, signal } from "@angular/core";
+
+@Directive({
+  selector: '[appOverflowDetector]',
+  exportAs: 'appOverflowDetector'
+})
+export class OverflowDetector {
```
1nf0rmagician: I first wanted to suggest to just make the `element.scrollWidth > element.clientWidth` calculation inline (which I still think makes sense, if we only use this once). Instead I was wondering if you want to prevent the line wrap in the expandable elements in the main view as well? 🤔
The line wrap looks a bit weird from my opinion...

However, the order & operation number is crucial to the meaning of the elements, which under accessibility standards means it should not be truncated there. However, Since this can be arbitrarily long depending on the using system, I'm open for discussing that point 🤔
Maybe ellipsis on the collapsed state only? Not sure whether that can reasonably placed in this PR or should be done separately though.

src/Moryx.Orders.Web/app/src/app/components/dialog-context/overflow-detector.ts
1nf0rmagician: The M3 specifically states that truncated text should be avialable in some form and lists tooltips as an option [here](https://m3.material.io/foundations/writing/text-truncation).
While the M§ not directly answers whether the tooltips should be disabled, if no truncation happens, I would argue it can be derived from the usage guidline on tooltips over buttons with texts, where they shouldn't be used as described [here](https://m3.material.io/components/tooltips/guidelines).

I actually was under the impression we were already using this... well, I was mistaken 😅
I would accept it as a solution here, but either @ArthurPxC you are motivated to skim through the UIs and unify the behaviour by putting this directive in the ngs-web-framework, or we create a ticket to do that in a subsequent PR 🙃

In any case comments on the directive will surely help then 😅

ArthurPxC:
I have removed the directive to shorten the discussion.

src/Moryx.Orders.Web/app/src/app/components/dialog-context/dialog-context.ts
```diff
@@ -0,0 +1,27 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [
+    TranslatePipe,
+    CommonModule,
+    MatTooltip
+  ]
+})
+export class DialogContext {
+  protected TranslationConstants = TranslationConstants;
+  public operationModel = input.required<OperationModel>();
```
dbeuchler: We do not use public, without modifier they are public.

dbeuchler (approved):

PR #1360, opened 2026-07-20
Improved order dialogs

ArthurPxC:
1) Removed a flicker in the begin order dialog causd by late operator field visibility evaluation
2) Moved confirmation radio group in the report dialog to the end
3) Updated spacing between mat form fields and comparable controls to be the same across dialogs

image

image

image

image

### DIFF
--- a/src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
+++ b/src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
@@ -128,11 +128,8 @@
     }
   </mat-form-field>
 
-  <app-operator-selector class="full-width"
-    [isEnabled]="canBegin"
-    [(selectedOperatorId)]="selectedOperatorId"
-    (creatingOperatorFailed)="creatingOperatorFailed.set($event)">
-  </app-operator-selector>
+  <app-operator-selector class="full-width top-spaced" [isEnabled]="canBegin"
+    [(selectedOperatorId)]="selectedOperatorId" (creatingOperatorFailed)="creatingOperatorFailed.set($event)"/>
 </mat-dialog-content>
 
 <div mat-dialog-actions align="end">
--- a/src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
+++ b/src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
@@ -30,7 +30,7 @@
     </div>
   }
 
-  <app-dialog-context [operationModel]="operation.model"></app-dialog-context>
+  <app-dialog-context class="full-width dialog-context" [operationModel]="operation.model"></app-dialog-context>
 
   <mat-grid-list cols="3" rowHeight="50px">
     <mat-grid-tile>
@@ -128,7 +128,7 @@
     }
   </mat-form-field>
 
-  <app-operator-selector class="operator-selector"
+  <app-operator-selector class="full-width"
     [isEnabled]="canBegin"
     [(selectedOperatorId)]="selectedOperatorId"
     (creatingOperatorFailed)="creatingOperatorFailed.set($event)">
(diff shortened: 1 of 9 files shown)


dbeuchler:
Could you check the spacing between "The target amount cannot further reduced" and the operator selection? 

- After the comment in report-dialog there is somehow too much space.
- Between the amount and comment field is no space

src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
```diff
@@ -30,7 +30,7 @@ <h2 mat-dialog-title>
     </div>
   }
 
-  <app-dialog-context [operationModel]="operation.model"></app-dialog-context>
+  <app-dialog-context class="full-width dialog-context" [operationModel]="operation.model"></app-dialog-context>
```
1nf0rmagician, line 33: (Optional) Instead of adding the styling classes on each mention, you could apply the styling to the `:host{}` element in the dialog-context.scss file

src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
```diff
@@ -128,7 +128,7 @@ <h2 mat-dialog-title>
     }
   </mat-form-field>
 
-  <app-operator-selector class="operator-selector"
+  <app-operator-selector class="full-width"
```
1nf0rmagician: Same here

src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
ArthurPxC: I know, but margins and arrangements do not belong into the control's default style. They belong into the context, where the control is used. For example a button should never come with a bottom margin of 8px inside it's default style, because it dependends on where the button is used. In horizontal flex container or as last item in a vertical flex container, this would disturb the layout. And whether a control is stretched or not is again dependent on the button's surrounding. A control's style should always desribe the inner properties, which start at the border, go inside over the padding into the content.

src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
ArthurPxC: See above

ArthurPxC:
> Could you check the spacing between "The target amount cannot further reduced" and the operator selection?
> 
> * After the comment in report-dialog there is somehow too much space.
> * Between the amount and comment field is no space

I have checked this. The spacing for the "The target..." message remained the same:

image

For mat form fields we have a lack of definition. Either we add extra margin between each of them, or we add no extra space. I sticked to adding no extra space, as adding extra space had always to be reverted in the past:

image

src/Moryx.Orders.Web/app/src/app/dialogs/begin-dialog/begin-dialog.html
1nf0rmagician: Good point, hadn't thought about it that way 🤔

dbeuchler:
There is no lack of definition, we add spacing 8px in form fields.. see entry editor

1nf0rmagician:
To provide the concrete source for future reference 
https://github.com/PHOENIXCONTACT/ngx-moryx-web/blob/dev/projects/ngx-web-framework/entry-editor/entry-editor.scss#L23

image

dbeuchler:
> To provide the concrete source for future reference 
> 
> https://github.com/PHOENIXCONTACT/ngx-moryx-web/blob/dev/projects/ngx-web-framework/entry-editor/entry-editor.scss#L23
> 
> ...

This!

ArthurPxC:
I will adapt this, but it will make some dialogs to have more space between elements as before, especially visible, when no hint is shown, as currently there is no margin like in here:

image

ArthurPxC:
Adapted margins, new layouts:

image

image

image

image

1nf0rmagician (approved):

PR #1337, opened 2026-07-13
Additional web-project optimizations

dbeuchler:
for https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/1061
based on #1335

PR #1362, opened 2026-07-21
Fix automatic reconnect in visual instruction stream & notification stream 

andreniggemann:
### Summary

Fix automatic reconnect in visual instruction stream and Notification stream, by creating the stream independent of module state,
Fix duplicated event registration for Visual Instructions
Add loading spinner to visualInstructions UI to show that the stream in not connected

### Checklist for Submitter

### Review

**Typical tasks**

**Clean code**


dbeuchler (dismissed):

dbeuchler:
I already ask me, if there was a property 😂

andreniggemann:
@dbeuchler thanks for the review, but that's not really ready yet. I thought I solved it yesterday but in testing today I figured there are more problems than anticipated. It needs more work

dbeuchler:
Have some one of you checked for a possible middleware instead of all the local solutions?

andreniggemann:
> Have some one of you checked for a possible middleware instead of all the local solutions?

I doubt this is possible in a meaningful way. The underlying issue is the EventSource implementation in the Browser.
When the connection to the server fails completely the connection will be retried correctly, but as soon as any error response is returned (regardless of a retry header and whether the status code is 4xx or 5xx) the browser will stop retrying the connection.

We could try to add a middleware that hooks into each endpoint, try to figure out if the endpoint should return a stream and if it should return a stream and returns an error code instead we abort the connection without sending the response to the client.

I don't like this approach, because
a) it hides the error information from the client and thus from the developer
b) we would probably find cases where returning the error is intentional i.e. in VisualInstructions when no identifier is set.

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction.service.ts
```diff
@@ -25,14 +25,26 @@ export class InstructionService {
...
     this.eventSource.onmessage = event => {
       const instructions = JSON.parse(event.data);
       this._instructions.set(instructions);
+      if(!this.connected()){
+        this.connected.set(true);
+      }
     };
+      this.eventSource.onerror = event => {
+      // const instructions = JSON.parse(event.data);
+      //this._instructions.set(instructions);
```
dbeuchler: remove these lines

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction.service.ts
```diff
@@ -25,14 +25,26 @@ export class InstructionService {
...
+      if(this.connected()){
+        this.connected.set(false);
+      }
```
dbeuchler: Angular signals only notify consumers when the value actually changes (using equality checking), so setting the same value again is not a problem and handled internally.

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction.service.ts
```diff
@@ -25,14 +25,26 @@ export class InstructionService {
...
   readonly instructions = this._instructions.asReadonly();
 
   public connect() {
     this.eventSource = new EventSource(this.visualInstructionsService.rootUrl + '/api/moryx/instructions/stream', {withCredentials: !environment.production});
     this.eventSource.onmessage = event => {
       const instructions = JSON.parse(event.data);
       this._instructions.set(instructions);
+      if(!this.connected()){
+        this.connected.set(true);
+      }
```
dbeuchler: same here

src/Moryx.VisualInstructions.Web/app/src/app/components/worker-instructions/worker-instructions.ts
```diff
@@ -66,6 +68,7 @@ export class WorkerInstructions {
   protected displayedInstruction = signal<InstructionModel | undefined>(undefined);
   protected mediaItemsContent = signal<DisplayedMediaContent[]>([]);
   protected textItems = signal<InstructionItemModel[]>([]);
+  readonly connected = this.instructionService.connected.asReadonly();
```
dbeuchler: use protected here: https://angular.dev/style-guide#use-protected-on-class-members-that-are-only-used-by-a-components-template

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction.service.ts
```diff
@@ -25,14 +25,26 @@ export class InstructionService {
   private eventSource?: EventSource;
 
   private readonly _instructions = signal<InstructionModel[]>([]);
+  readonly connected = signal<boolean>(false);
```
dbeuchler: use readonly signal here, reference: https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1337/changes/3f2a0c23d65a07ebaa72fd0e76e8508c90e63f13

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
```diff
@@ -54,17 +55,34 @@ public async Task InstructionStream(CancellationToken cancellationToken)
             return;
         }
 
+        var channel = Channel.CreateUnbounded<string>();
```
dbeuchler, line 58: So moving out channel creation out of subscribe registers event handlers for each subscriptions seems to be cleaner. Can we apply that to all of our other SSEs too? Currently they are all consistent.

src/Moryx.VisualInstructions.Web/app/src/app/components/worker-instructions/worker-instructions.ts
andreniggemann: Thanks for the heads up. I did not know

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
andreniggemann: I think we should, but I would prefer to do it in another PR.

The initial task was to do it for Visual Instructions and Notifications, because they are the UIs that are most likely to be on screens with no user input and will thus not be regularly refreshed.

Marcel asked us to extend the guidelines that are currently being written with our findings
https://github.com/PHOENIXCONTACT/MORYX-Home/blob/feat/api-best-practices/development/architecture/api-best-practices.md

After that we could subsequently update each of the other SSE Endpoints

src/Moryx.Notifications.Endpoints/NotificationPublisherController.cs
```diff
@@ -89,31 +102,35 @@ public async Task NotificationStream(CancellationToken cancellationToken)
...
-            finally
+            catch (HealthStateException)
             {
-                _notificationStreamSubscribers.TryRemove(id, out _);
+                // Ignore if module is not ready yet.
+            }
+            yield return new SseItem<string>(JsonSerializer.Serialize(initialNotifications, _serializerOptions));
+
+
+
```
1nf0rmagician, line 130: Multiple empty lines should be marked by the IDE as violating editorconfig

src/Moryx.VisualInstructions.Controller/Facades/VisualInstructionsFacade.cs
```diff
@@ -26,17 +25,19 @@ public VisualInstructionsFacade()
     #endregion
 
     /// <inheritdoc cref="IFacadeControl"/>
-    public void Activate()
+    public override void Activate()
     {
         Controller.InstructionAdded += ParallelOperations.DecoupleListener<InstructionEventArgs>(OnInstructionAdded);
         Controller.InstructionCleared += ParallelOperations.DecoupleListener<InstructionEventArgs>(OnInstructionCleared);
+        base.Activate();
```
1nf0rmagician, line 32: Not sure whether it is an issue, but it feels like if the deactivation happens after the event subscriptions, the activation should happen beforehand?

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
1nf0rmagician: Will you or @KrzysztofMarchewkaPxC start working on it right away or create an issue as to not forget about it? 😬

andreniggemann: I won’t get to it for at least two weeks.
@KrzysztofMarchewkaPxC if you have some time, feel free to tackle this

src/Moryx.Notifications.Publisher/Facade/NotificationPublisherFacade.cs
```diff
@@ -5,24 +5,22 @@
...
 
     public NotificationPublisherFacade()
     {
         ValidateHealthState = InitialValidateHealthState;
     }
 
-    public void Activate()
+    public override void Activate()
     {
         NotificationManager.Published += OnPublished;
```
dbeuchler, line 19: Missing base call

src/Moryx.Notifications.Publisher/Facade/NotificationPublisherFacade.cs
```diff
@@ -5,24 +5,22 @@
...
         NotificationManager.Acknowledged += OnAcknowledged;
     }
 
-    public void Deactivate()
+    public override void Deactivate()
     {
         NotificationManager.Published -= OnPublished;
```
dbeuchler, line 27: Missing base call

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
```diff
@@ -54,17 +55,34 @@ public async Task InstructionStream(CancellationToken cancellationToken)
...
-            Broadcast(e.Identifier));
+        EventHandler<InstructionEventArgs> eventHandler = (_, e) =>
+        {
+            if (e.Identifier == identifier)
+            {
+                Broadcast();
+            }
+        };
+
+        EventHandler<bool> stateChangedEventHandler = (args, ready) =>
```
dbeuchler: parameter naming (args, ready) should be (sender, eventArgs or ready) to follow .NET conventions. The sender is not used here and can be _.

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
```diff
@@ -26,17 +27,17 @@ public class VisualInstructionsController : ControllerBase
...
     {
         PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
         Converters = { new JsonStringEnumConverter() }
     };
     private readonly Converter _converter;
-    public VisualInstructionsController(IVisualInstructions visualInstructions, IModuleManager moduleManager, IServiceProvider serviceProvider)
+    public VisualInstructionsController(IVisualInstructions visualInstructions, IModuleManager moduleManager, IServiceProvider serviceProvider, ILogger<VisualInstructionsController> logger)
     {
         _visualInstructions = visualInstructions;
+        _logger = logger;
```
dbeuchler: `ILogger<VisualInstructionsController>` never used. Either use it (e.g., log reconnect events) or remove it. Its more interesting on UI, so I would remove it here.

src/Moryx.VisualInstructions.Endpoints/VisualInstructionsController.cs
```diff
@@ -76,45 +94,42 @@ public async Task InstructionStream(CancellationToken cancellationToken)
...
+        async IAsyncEnumerable<SseItem<string>> Subscribe([EnumeratorCancellation] CancellationToken cancelToken)
         {
-            var channel = Channel.CreateUnbounded<string>();
-            var id = Guid.NewGuid();
-            _instructionStreamSubscribers[id] = (identifier, channel);
-
-            // Send all instructions as first item
-            var initialInstructions = _visualInstructions.GetInstructions(identifier).Select(_converter.ToModel).ToArray();
-            yield return JsonSerializer.Serialize(initialInstructions, _serializerOptions);
+
```
dbeuchler: blank line

dbeuchler (changes requested):

1nf0rmagician (approved):

dbeuchler (approved):

PR #1363, opened 2026-07-23
Remove wrong logging in ResourceSynchronizationService

LucasGnoerich:
The service sent a log message whenever a resource was created that was not set as synchronizable simply because the equation with null for SynchronizationTypeId matches when no attribute was set at all. When creating new processholders for example instead of recycling existing ones because they are not supposed to be set at all times will just fill the log rather frequently if they are not marked as synchronizable.
Additionally added logging for updating and removing local resources because we noticed that in one instance the synchronisation was suddenly not registered anymore for some reason after orders were newly pulled and assigned. We want to see those instances more clearly in our log. This will be watched further internally.

### DIFF
--- a/src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
+++ b/src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
@@ -271,6 +271,7 @@ public class ResourceSynchronizationService : IMqttService
     {
         return resourceManagement.ModifyUnsafeAsync(matchingResource.Id, resource =>
         {
+            _logger.LogInformation("Updating local resource with identity '{identity}'", (resource as IIdentifiableObject)?.Identity?.Identifier); 
             SetInstanceValue(jsonProperties, resource);
             return Task.FromResult(true);
         });
@@ -280,6 +281,7 @@ public class ResourceSynchronizationService : IMqttService
     {
         var message = MqttMessageSerialization.GetJsonPayload(matchingResource, _options.JsonSerializerOptions);
         TryAddToMessageCache(GetHashString(message));
+        _logger.LogInformation("Removing local resource with identity '{identity}'", (matchingResource as IIdentifiableObject)?.Identity?.Identifier);
 
         return resourceManagement.DeleteAsync(matchingResource.Id);
     }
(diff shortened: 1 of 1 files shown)


src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
```diff
@@ -271,6 +271,7 @@ private Task UpdateLocalInstanceAsync(IResourceManagement resourceManagement, Js
     {
         return resourceManagement.ModifyUnsafeAsync(matchingResource.Id, resource =>
         {
+            _logger.LogInformation("Updating local resource with identity '{identity}'", (resource as IIdentifiableObject)?.Identity?.Identifier);
```
dbeuchler, line 274: A resource is an IPersistentObject and not IIdentifiableObject. Use Id here. Maybe the better place is also the internal component to log instead of the endpoint. Other parts of the application using the facade would also benefit from the explicit log message. With our child-logger structure it is also easier to filter them if not necessary or explicit necessary.

src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
```diff
@@ -280,6 +281,7 @@ private Task RemoveLocalInstanceAsync(IResourceManagement resourceManagement, IR
     {
         var message = MqttMessageSerialization.GetJsonPayload(matchingResource, _options.JsonSerializerOptions);
         TryAddToMessageCache(GetHashString(message));
+        _logger.LogInformation("Removing local resource with identity '{identity}'", (matchingResource as IIdentifiableObject)?.Identity?.Identifier);
```
dbeuchler, line 284: Same here

dbeuchler (changes requested):

src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
1nf0rmagician: While we can discuss logging on the facade level of the resource management as well, this is logging for the resource synchronization specifically. The resource synchronization works with the `IIdentifiableObject` to synchronize resource throughout multiple instances of the resource management. Hence, logging of changes for these resources through the synchronization should log the identifier 😊

1nf0rmagician (approved):

src/Moryx.AbstractionLayer.Resources.Mqtt.Endpoints/ResourceSynchronizationService.cs
dbeuchler: Upps

dbeuchler (approved):

PR #1364, opened 2026-07-23
Fix generic JSON persistence and text column deserialization

fmoebert:
## Summary

This MR fixes several issues in the generic product persistence layer.

### GenericEntityMapper
- reserve JsonColumn exclusively for JSON content
- prevent JsonColumn from being used as a regular mapped property
- store unmapped properties in the JSON fallback column
- write default/unset values as explicit null entries in JSON

### AutoConfigurator
- exclude JsonColumn from automatically generated property mappings
- exclude JsonColumn from selectable generic columns

### TextColumnMapper
- handle null and empty text values when deserializing complex JSON-backed properties
- prevent ArgumentNullException for newly introduced optional complex properties stored in text columns

~~### FloatColumnMapper~~
~~round floating point values before persistence~~
~~avoid floating-point representation artifacts in stored values~~

### Samples
- added dedicated sample product types for reproducing and validating JSON fallback and text-column JSON behavior

### Tests
- Save and load generic json type
- Save and load complex property
- Save and load type with complex property is null
- Load type with plain text in json column
- Load typ with empty json object

## Notes

The ShellNavigator / Microsoft.Build startup warning is already present on the base dev branch and is unrelated to the changes in this MR.

### DIFF
--- a/src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
+++ b/src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
 // Licensed under the Apache License, Version 2.0
 
+using System.Reflection;
 using Moryx.AbstractionLayer.Products;
 using Moryx.AbstractionLayer.Recipes;
 using Moryx.Configuration;
@@ -40,8 +41,10 @@ internal class AutoConfigurator
         var productTypes = ReflectionTool.GetPublicClasses<ProductType>();
         var product = productTypes.FirstOrDefault(p => p.FullName == productType);
         if (product == null)
+        {
             return $"Found no product type {productType ?? string.Empty}!\n" +
                    "Available types: " + string.Join(", ", productTypes.Select(t => t.FullName));
+        }
 
         var result = string.Empty;
         // First try to select the best type strategy
@@ -78,7 +81,10 @@ internal class AutoConfigurator
             {
                 var linkType = link.PropertyType;
                 if (typeof(IEnumerable<ProductPartLink>).IsAssignableFrom(linkType))
+                {
                     linkType = linkType.GetGenericArguments()[0];
+                }
+
                 var linkConfig = StrategyConfig<IProductLinkStrategy, ProductLinkConfiguration, ProductPartLink>(linkType);
                 if (linkConfig == null)
                 {
@@ -105,8 +111,10 @@ internal class AutoConfigurator
         var types = ReflectionTool.GetPublicClasses<ProductInstance>();
         var instance = types.FirstOrDefault(p => p.FullName == instanceType);
         if (instance == null)
+        {
             return $"Found no instance type {instanceType ?? string.Empty}!\n" +
                    "Available types: " + string.Join(", ", types.Select(t => t.FullName));
+        }
 
         var result = string.Empty;
         if (Config.InstanceStrategies.Any(s => s.TargetType == instanceType))
@@ -138,8 +146,10 @@ internal class AutoConfigurator
         var types = ReflectionTool.GetPublicClasses<IProductRecipe>();
         var recipe = types.FirstOrDefault(p => p.FullName == recipeType);
         if (recipe == null)
+        {
             return $"Found no recipe type {recipeType ?? string.Empty}!\n" +
                    "Available types: " + string.Join(", ", types.Select(t => t.FullName));
+        }
 
         var result = string.Empty;
         if (Config.RecipeStrategies.Any(s => s.TargetType == recipeType))
@@ -169,7 +179,9 @@ internal class AutoConfigurator
     {
         var tuple = CreateConfig<TStrategy, TConfig>(targetType);
         if (tuple == null)
+        {
             return null;
+        }
 
         var config = tuple.Item2;
         config.TargetType = targetType.FullName;
@@ -179,28 +191,49 @@ internal class AutoConfigurator
 
         // Optionally try to configure property mappers
         if (config is not IPropertyMappedConfiguration propertyMapperConfig)
+        {
             return config;
+        }
 
-        var remainingColumns = typeof(IGenericColumns).GetProperties()
-            .OrderBy(p => p.Name).ToList();
+        var jsonColumn = (config as IGenericMapperConfiguration)?.JsonColumn;
+
+        var remainingColumns = typeof(IGenericColumns)
+            .GetProperties()
+            .Where(p => string.IsNullOrEmpty(jsonColumn) ||
+                        !string.Equals(p.Name, jsonColumn, StringComparison.OrdinalIgnoreCase))
+            .OrderBy(p => p.Name)
+            .ToList();
 
         var baseTypeWrapper = Storage.GetTypeWrapper(typeof(TBaseType).FullName);
-        var baseProperties = baseTypeWrapper != null ? baseTypeWrapper.Properties.ToArray() : typeof(TBaseType).GetProperties();
+        var baseProperties = baseTypeWrapper?.Properties.ToArray() ?? typeof(TBaseType).GetProperties();
 
         var targetTypeWrapper = Storage.GetTypeWrapper(targetType.FullName);
-        var targetProperties = targetTypeWrapper != null ? targetTypeWrapper.Properties.ToArray() : targetType.GetProperties();
-        var filteredProperties = targetProperties
-            .Where(p => baseProperties.All(bp => bp.Name != p.Name))
-            .Where(p => p.GetSetMethod() != null)
-            .Where(p => !typeof(ProductPartLink).IsAssignableFrom(p.PropertyType) & !typeof(IEnumerable<ProductPartLink>).IsAssignableFrom(p.PropertyType))
-            .Where(p => !typeof(ProductInstance).IsAssignableFrom(p.PropertyType) & !typeof(IEnumerable<ProductInstance>).IsAssignableFrom(p.PropertyType))
+        var targetProperties = targetTypeWrapper?.Properties.ToArray() ?? targetType.GetProperties();
+
+        var excludedTypes = new[]
+        {
+            typeof(ProductPartLink),
+            typeof(IEnumerable<ProductPartLink>),
+            typeof(ProductInstance),
+            typeof(IEnumerable<ProductInstance>)
+        };
+
+        bool IsExcluded(PropertyInfo property) => excludedTypes.Any(t => t.IsAssignableFrom(property.PropertyType));
+        var filteredProperties = targetProperties.Where(p =>
+            baseProperties.All(bp => bp.Name != p.Name)
+            && p.GetSetMethod() != null
+            && !IsExcluded(p)
+            && (string.IsNullOrEmpty(jsonColumn)
+            || !string.Equals(p.Name, jsonColumn, StringComparison.OrdinalIgnoreCase)))
             .ToList();
 
         foreach (var property in filteredProperties)
         {
             var propertyTuple = CreateConfig<IPropertyMapper, PropertyMapperConfig>(property.PropertyType);
             if (propertyTuple == null)
+            {
                 continue;
+            }
 
             var strategy = propertyTuple.Item1;
             var propertyConfig = propertyTuple.Item2;
@@ -210,7 +243,9 @@ internal class AutoConfigurator
             var columnType = strategy.GetCustomAttribute<PropertyStrategyConfigurationAttribute>()?.ColumnType ?? typeof(string);
             var column = remainingColumns.FirstOrDefault(rc => rc.PropertyType == columnType);
             if (column == null)
+            {
                 continue;
+            }
 
             remainingColumns.Remove(column);
             propertyConfig.Column = column.Name;
@@ -227,7 +262,9 @@ internal class AutoConfigurator
         var strategies = Container.GetRegisteredImplementations(typeof(TStrategy));
         var typeStrategy = SelectStrategy(strategies, targetType);
         if (typeStrategy == null)
+        {
             return null;
+        }
 
         var configType = typeStrategy.GetCustomAttribute<ExpectedConfigAttribute>()?.ExpectedConfigType
                          ?? typeof(TConfig);
--- a/src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
+++ b/src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
@@ -2,11 +2,13 @@
 // Licensed under the Apache License, Version 2.0
 
 using System.Linq.Expressions;
+using Microsoft.Extensions.Logging;
 using Moryx.Container;
 using Moryx.Products.Management.Model;
 using Moryx.Serialization;
 using Moryx.Tools;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using static Moryx.Products.Management.ProductExpressionHelpers;
 
 namespace Moryx.Products.Management;
@@ -24,7 +26,10 @@ internal class GenericEntityMapper<TBase, TReference> : IGenericMapper
     /// </summary>
     public IPropertyMapperFactory MapperFactory { get; set; }
 
+    public ILogger Logger { get; set; }
+
     private IPropertyMapper[] _configuredMappers;
+
 
     private JsonSerializerSettings _jsonSettings;
 
@@ -37,25 +42,47 @@ internal class GenericEntityMapper<TBase, TReference> : IGenericMapper
         _jsonAccessor = ReflectionTool.PropertyAccessor<IGenericColumns, string>(jsonColumn);
 
         var baseProperties = typeof(TBase).GetProperties().Select(p => p.Name).ToArray();
-        var configuredProperties = config.PropertyConfigs.Select(cm => cm.PropertyName);
+
+        // Legacy compatibility:
+        // Historically some configurations used the JsonColumn both as JSON storage
+        // and as a regular property mapping. This remains supported until the next
+        // major release but should no longer be generated automatically.
+        // TODO: MORYX 12: Maybe remove support for using JsonColumn as a regular property mapping.
+        if (config.PropertyConfigs.Any(pc => string.Equals(pc.PropertyName, config.JsonColumn, StringComparison.OrdinalIgnoreCase)))
+        {
+            Logger.LogWarning(
+                "Detected a PropertyConfig for JsonColumn '{JsonColumn}'. " +
+                "This configuration is deprecated and may no longer be supported in a future major release.",
+                config.JsonColumn);
+        }
+
+        var configuredProperties = config.PropertyConfigs
+            .Select(cm => cm.PropertyName)
+            .ToArray();
 
         var readOnlyProperties = concreteType.GetProperties()
-            .Where(p => p.GetSetMethod() == null).Select(p => p.Name).ToArray();
+            .Where(p => p.GetSetMethod() == null)
+            .Select(p => p.Name)
+            .ToArray();
 
-        // The json should not contain base, configured nor readonly properties
         var jsonIgnoredProperties = baseProperties
             .Concat(configuredProperties)
-            .Concat(readOnlyProperties).ToArray();
+            .Concat(readOnlyProperties)
+            .ToArray();
 
         _jsonSettings = JsonSettings.Minimal
-            .Overwrite(j => j.ContractResolver = new DifferentialContractResolver<TReference>(jsonIgnoredProperties));
+            .Overwrite(j => j.ContractResolver =
+                new DifferentialContractResolver<TReference>(jsonIgnoredProperties));
 
         // Properties where no mapper should be created for: base and read only properties
         var mapperIgnoredProperties = baseProperties
-            .Concat(readOnlyProperties).ToArray();
+            .Concat(readOnlyProperties)
+            .ToArray();
 
-        _configuredMappers = config.PropertyConfigs.Where(pc => !mapperIgnoredProperties.Contains(pc.PropertyName))
-            .Select(pc => MapperFactory.Create(pc, concreteType)).ToArray();
+        _configuredMappers = config.PropertyConfigs
+            .Where(pc => !mapperIgnoredProperties.Contains(pc.PropertyName))
+            .Select(pc => MapperFactory.Create(pc, concreteType))
+            .ToArray();
     }
 
     public bool HasChanged(IGenericColumns storage, object instance)
@@ -116,32 +143,127 @@ internal class GenericEntityMapper<TBase, TReference> : IGenericMapper
         return Expression.Lambda(body, columnParam) as Expression<Func<IGenericColumns, bool>>;
     }
 
+    /// <inheritdoc />
     public void ReadValue(IGenericColumns source, object target)
     {
-        // Use all configured mappers
-        var properties = source;
+        // Read all directly mapped properties
         foreach (var mapper in _configuredMappers)
         {
-            mapper.ReadValue(properties, target);
+            mapper.ReadValue(source, target);
         }
 
-        // Fill the rest from JSON
+        // Reading JSON from the JsonColumn
         var json = _jsonAccessor.ReadProperty(source);
-        if (!string.IsNullOrEmpty(json))
-            JsonConvert.PopulateObject(json, target, _jsonSettings);
+        if (string.IsNullOrEmpty(json))
+        {
+            return;
+        }
 
+        // Legacy compatibility:
+        // Older systems may contain plain text in the JsonColumn.
+        // In this case simply ignore the content and continue loading
+        // the directly mapped properties.
+        if (!(json.StartsWith("{") || json.StartsWith("[")))
+        {
+            Logger.LogWarning("Ignoring non-JSON content in JsonColumn '{Column}'. Value: '{Value}'", _jsonAccessor.Property.Name, json);
+            return;
+        }
+
+        // Populate remaining properties from the JsonColumn JSON payload
+        try
+        {
+            JsonConvert.PopulateObject(json, target, _jsonSettings);
+        }
+        catch (JsonException ex)
+        {
+            throw new InvalidOperationException(
+                $"Failed to deserialize JSON from column '{_jsonAccessor.Property.Name}' for target type '{target.GetType().FullName}'. " +
+                $"Stored value: '{json}'", ex);
+        }
     }
 
+    /// <inheritdoc/>>
     public void WriteValue(object source, IGenericColumns target)
     {
-        // Convert and write JSON
-        var json = JsonConvert.SerializeObject(source, _jsonSettings);
-        _jsonAccessor.WriteProperty(target, json);
-
-        // Execute property mappers
+        // First, write all the directly mapped properties
         foreach (var mapper in _configuredMappers)
         {
             mapper.WriteValue(source, target);
         }
+
+        // Generate JSON from the rest of the object
+        var mappedNames = _configuredMappers
+            .Select(m => m.Property.Name)
+            .ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+        // JSON for all other props
+        var jsonObject = new JObject();
+
+        foreach (var prop in source.GetType().GetProperties())
+        {
+            if (!prop.CanRead)
+            {
+                continue;
+            }
+
+            if (!prop.CanWrite)
+            {
+                continue;
+            }
+
+            // Do not include base, read-only, or directly mapped properties in the JSON
+            if (mappedNames.Contains(prop.Name))
+            {
+                continue;
+            }
+
+            // jump Indexer 
+            if (prop.GetIndexParameters().Length > 0)
+            {
+                continue;
+            }
+
+            // jump Read-only Props
+            if (prop.GetSetMethod() == null)
+            {
+                continue;
+            }
+
+            var value = prop.GetValue(source);
+
+            // "Not set yet" -> null in JSON
+            if (IsDefaultOrNull(prop.PropertyType, value))
+            {
+                jsonObject[prop.Name] = JValue.CreateNull();
+            }
+            else
+            {
+                jsonObject[prop.Name] = JToken.FromObject(value, JsonSerializer.Create(_jsonSettings));
+            }
+        }
+
+        // Write JSON to the Json - Column
+        _jsonAccessor.WriteProperty(target, jsonObject.ToString(Formatting.None));
+    }
+
+    private static bool IsDefaultOrNull(Type propertyType, object value)
+    {
+        if (value == null)
+        {
+            return true;
+        }
+
+        var underlyingType = Nullable.GetUnderlyingType(propertyType);
+        var effectiveType = underlyingType ?? propertyType;
+
+        // Reference type
+        if (!effectiveType.IsValueType)
+        {
+            return false;
+        }
+
+        // Value type: handle Default as "not set yet"
+        var defaultValue = Activator.CreateInstance(effectiveType);
+        return Equals(value, defaultValue);
     }
 }
--- a/src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
+++ b/src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
@@ -18,35 +18,64 @@ internal class TextColumnMapper : ColumnMapper<string>
 {
     public TextColumnMapper(Type targetType) : base(targetType)
     {
-
     }
 
     protected override IPropertyAccessor<object, string> CreatePropertyAccessor(PropertyInfo objectProp)
     {
         var propType = objectProp.PropertyType;
-        // Convert return value to string
+
+        // Convert Guid to string
         if (propType == typeof(Guid))
-            return new ConversionAccessor<string, Guid>(objectProp,
+        {
+            return new ConversionAccessor<string, Guid>(
+                objectProp,
                 guid => guid != Guid.Empty ? guid.ToString() : null,
                 str => !string.IsNullOrEmpty(str) ? Guid.Parse(str) : Guid.Empty);
+        }
 
-        if ((propType.IsClass | propType.IsInterface) && propType != typeof(string))
-            return new ConversionAccessor<string, object>(objectProp,
+        // Complex reference types -> JSON
+        if ((propType.IsClass || propType.IsInterface) && propType != typeof(string))
+        {
+            return new ConversionAccessor<string, object>(
+                objectProp,
                 o => JsonConvert.SerializeObject(o, Property.PropertyType, JsonSettings.Minimal),
-                s => JsonConvert.DeserializeObject(s, Property.PropertyType, JsonSettings.Minimal));
+                s =>
+                {
+                    if (string.IsNullOrWhiteSpace(s))
+                    {
+                        var effectiveType = Nullable.GetUnderlyingType(Property.PropertyType) ?? Property.PropertyType;
 
+                        if (!effectiveType.IsValueType)
+                            return null;
+
+                        return Activator.CreateInstance(effectiveType);
+                    }
+
+                    return JsonConvert.DeserializeObject(s, Property.PropertyType, JsonSettings.Minimal);
+                });
+        }
+
+        // Normal string property
         return base.CreatePropertyAccessor(objectProp);
     }
 
     protected override object ToExpressionValue(object value)
     {
         var propType = Property.PropertyType;
+
+        // Guid -> string
         if (propType == typeof(Guid))
+        {
             return ((Guid)value).ToString();
+        }
 
-        if ((propType.IsClass | propType.IsInterface) && propType != typeof(string))
-            return JsonConvert.SerializeObject(value, Property.PropertyType, JsonSettings.Minimal);
+        // Complex reference types -> JSON
+        if ((propType.IsClass || propType.IsInterface) && propType != typeof(string))
+        {
+            return value == null ? null : JsonConvert.SerializeObject(value, Property.PropertyType, JsonSettings.Minimal);
+        }
 
+        // Normal string / primitive conversion
         return base.ToExpressionValue(value);
     }
 }
--- a/src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
+++ b/src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
@@ -1,0 +1,71 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+using Moryx.AbstractionLayer.Products;
+using Moryx.Modules;
+
+namespace Moryx.Products.TestProducts;
+
+[ExpectedConfig(typeof(GenericJsonTestProductTypeImporterConfig))]
+[ProductImporter(nameof(GenericJsonTestProductTypeImporter))]
+public class GenericJsonTestProductTypeImporter : ProductImporterBase<GenericJsonTestProductTypeImporterConfig, SpecializedJsonTestProductTypeParameters>
+{
+    protected override Task<ProductImporterResult> ImportAsync(ProductImportContext context, SpecializedJsonTestProductTypeParameters parameters,
+        CancellationToken cancellationToken)
+    {
+        var product = new GenericJsonTestProductType()
+        {
+            Name = parameters.Name,
+            Identity = new ProductIdentity(parameters.Identifier, parameters.Revision),
+
+            //Text1 = "T1",
+            Text2 = "T2",
+            Text3 = "T3",
+            Text4 = "T4",
+            Text5 = "T5",
+            Text6 = "T6",
+            Text7 = "T7",
+            Text8 = "T8",
+            Text9 = "T9",
+            Text10 = "T10",
+
+            Integer1 = 1,
+            Integer2 = 2,
+            Integer3 = 3,
+            Integer4 = 4,
+            Integer5 = 5,
+            Integer6 = 6,
+            Integer7 = 7,
+            Integer8 = 8,
+            Integer9 = 9,
+            Integer10 = 10,
+
+            Float1 = 1.23f,
+            Float2 = 2.34f,
+            Float3 = 3.45f,
+            Float4 = 4.56f,
+            Float5 = 5.67f,
+            Float6 = 6.78f,
+            Float7 = 7.89f,
+            Float8 = 8.90f,
+            Float9 = 9.01f,
+            Float10 = 10.11f,
+
+            ComplexData1 = new ComplexData()
+        };
+
+        var result = new ProductImporterResult
+        {
+            ImportedTypes = [product]
+        };
+
+        return Task.FromResult(new ProductImporterResult
+        {
+            ImportedTypes =
+                [product]
+        });
+    }
+}
+
+public class SpecializedJsonTestProductTypeParameters : PrototypeParameters
+{
+}
--- a/src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
+++ b/src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
@@ -1,0 +1,35 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.TestProducts;
+
+public class TextColumnMapperTestProductTypeImporter : ProductImporterBase<TextColumnMapperTestProductTypeImporterConfig, TextColumnMapperTestProductTypeParameters>
+{
+    protected override Task<ProductImporterResult> ImportAsync(ProductImportContext context, TextColumnMapperTestProductTypeParameters parameters, CancellationToken cancellationToken)
+    {
+        var product = new TextColumnMapperTestProductType
+        {
+            Name = parameters.Name,
+            Identity = new ProductIdentity(parameters.Identifier, parameters.Revision),
+
+            ComplexData1 = new ComplexData
+            {
+                PropertyName = "First",
+                Number = 123,
+                Weight = 1.23f
+            },
+        };
+
+        return Task.FromResult(new ProductImporterResult
+        {
+            ImportedTypes = [product]
+        });
+    }
+}
+
+public class TextColumnMapperTestProductTypeParameters
+    : PrototypeParameters
+{
+}
--- a/src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporterConfig.cs
+++ b/src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporterConfig.cs
@@ -1,0 +1,13 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System.Runtime.Serialization;
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.TestProducts;
+
+[DataContract]
+public class TextColumnMapperTestProductTypeImporterConfig : ProductImporterConfig
+{
+    public override string PluginName => nameof(TextColumnMapperTestProductTypeImporter);
+}
--- a/src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
+++ b/src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
@@ -1,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Moryx.AbstractionLayer\Moryx.AbstractionLayer.csproj" />
+    <ProjectReference Include="..\Moryx.Products.Management\Moryx.Products.Management.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/Moryx.Products.TestProducts/ProductInstances/GenericJsonTestProductTypeInstance.cs
+++ b/src/Moryx.Products.TestProducts/ProductInstances/GenericJsonTestProductTypeInstance.cs
@@ -1,0 +1,10 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.TestProducts;
+
+public class GenericJsonTestProductTypeInstance : ProductInstance<GenericJsonTestProductType>
+{
+}
--- a/src/Moryx.Products.TestProducts/ProductInstances/TextColumnMapperTestProductTypeInstance.cs
+++ b/src/Moryx.Products.TestProducts/ProductInstances/TextColumnMapperTestProductTypeInstance.cs
@@ -1,0 +1,10 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.TestProducts;
+
+public class TextColumnMapperTestProductTypeInstance : ProductInstance<TextColumnMapperTestProductType>
+{
+}
--- a/src/Moryx.Products.TestProducts/ProductTypes/ComplexData.cs
+++ b/src/Moryx.Products.TestProducts/ProductTypes/ComplexData.cs
@@ -1,0 +1,21 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+namespace Moryx.Products.TestProducts;
+
+/// <summary>
+/// To test complex data in text columns
+/// </summary>
+public class ComplexData
+{
+    // ToDo: Property name "Name" is reserved and is covered by Parent class property so it can not be used here
+
+    public string Content { get; set; }
+
+    public string PropertyName { get; set; }
+
+    public int Number { get; set; }
+
+    public float Weight { get; set; }
+}
+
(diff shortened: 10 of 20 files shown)


src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
```diff
@@ -12,7 +15,46 @@ namespace Moryx.Products.Management;
 [Component(LifeCycle.Transient, typeof(IPropertyMapper), Name = nameof(FloatColumnMapper))]
 internal class FloatColumnMapper : ColumnMapper<double>
 {
+    private const int Precision = 8;
```
dbeuchler: We cannot change this behavior, we must store the full double/decimal/float. We must ensure that save and restore have the same values. A storage should not be responsible for the precision.

dbeuchler (changes requested):
There are a lot of behavioral changes we must explicitly check, some of them bloat the database, some break behavior, some round data. This week I habe no time to look at the specific lines but next...

src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
```diff
@@ -179,28 +190,43 @@ private TConfig StrategyConfig<TStrategy, TConfig, TBaseType>(Type targetType)
...
         var filteredProperties = targetProperties
             .Where(p => baseProperties.All(bp => bp.Name != p.Name))
             .Where(p => p.GetSetMethod() != null)
-            .Where(p => !typeof(ProductPartLink).IsAssignableFrom(p.PropertyType) & !typeof(IEnumerable<ProductPartLink>).IsAssignableFrom(p.PropertyType))
-            .Where(p => !typeof(ProductInstance).IsAssignableFrom(p.PropertyType) & !typeof(IEnumerable<ProductInstance>).IsAssignableFrom(p.PropertyType))
+            .Where(p => !typeof(ProductPartLink).IsAssignableFrom(p.PropertyType))
+            .Where(p => !typeof(IEnumerable<ProductPartLink>).IsAssignableFrom(p.PropertyType))
+            .Where(p => !typeof(ProductInstance).IsAssignableFrom(p.PropertyType))
+            .Where(p => !typeof(IEnumerable<ProductInstance>).IsAssignableFrom(p.PropertyType))
+            .Where(p => string.IsNullOrEmpty(jsonColumn) ||
```
1nf0rmagician, line 219: I'm not sure I fully understand this line, does it filter out properties that would actually be called Text8? 🤔🙈

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
```diff
@@ -12,7 +15,46 @@ namespace Moryx.Products.Management;
...
+    /// 
+    /// </summary>
+    /// <param name="targetType"></param>
     public FloatColumnMapper(Type targetType) : base(targetType)
     {
     }
-}
\ No newline at end of file
+
+    protected override IPropertyAccessor<object, double> CreatePropertyAccessor(PropertyInfo objectProp)
```
1nf0rmagician: Why was it necessary to introduce this from your pov? 🤔

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -37,15 +40,21 @@ public void Initialize(Type concreteType, IGenericMapperConfiguration config)
         _jsonAccessor = ReflectionTool.PropertyAccessor<IGenericColumns, string>(jsonColumn);
 
         var baseProperties = typeof(TBase).GetProperties().Select(p => p.Name).ToArray();
-        var configuredProperties = config.PropertyConfigs.Select(cm => cm.PropertyName);
+
+        //FMÖ: 22 07 2026
+        var configuredProperties = config.PropertyConfigs
+            .Select(cm => cm.PropertyName)
+            .Where(p => !string.Equals(p, config.JsonColumn, StringComparison.OrdinalIgnoreCase))
```
1nf0rmagician: I think making this class an `ILoggingComponent` and adding a warning message in the log, that we detected a property pointing to the configured JSON column would be nice here 😊

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -100,7 +113,7 @@ public Expression<Func<IGenericColumns, bool>> TransformSelector<T>(Expression<F
                     }
 
                     callValue = ExtractExpressionValue(call.Arguments.First());
-                    return Convert(((MemberExpression)call.Object).Member.Name, ExpressionType.Equal, callValue);
+                    return Convert(((MemberExpression)call.Object)?.Member.Name, ExpressionType.Equal, callValue);
```
1nf0rmagician: Out of interest: Did you actually provoke an exception here? If yes I would be interested to here how 😅

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +129,138 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
         return Expression.Lambda(body, columnParam) as Expression<Func<IGenericColumns, bool>>;
     }
 
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="source"></param>
+    /// <param name="target"></param>
+    /// <exception cref="ArgumentNullException"></exception>
+    /// <exception cref="InvalidOperationException"></exception>
```
1nf0rmagician: Use `inheritdoc` if you don't need a specific xml comment over the method here 😊

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +129,138 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
+        }
 
-    }
+        // JsonColumn must contain valid JSON
+        var trimmed = json.Trim();
+        if (!(trimmed.StartsWith("{") || trimmed.StartsWith("[")))
+        {
+            throw new InvalidOperationException(
+                $"The JSON column '{_jsonAccessor.Property.Name}' contains plain text instead of JSON. " +
+                $"Value: '{json}'. This usually means the JsonColumn is also configured as a normal property mapping.");
```
1nf0rmagician: > This usually means the JsonColumn is also configured as a normal property mapping.

Is this still true, with your changes from above? I thought the column would be blocked for other content now 🤔

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +129,138 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
+                $"Stored value: '{json}'", ex);
+        }
+    }
+    
+    /// <summary>
+    /// Also write null-values to the JSON
+    /// </summary>
+    /// <param name="source"></param>
+    /// <param name="target"></param>
+    /// <exception cref="ArgumentNullException"></exception>
```
1nf0rmagician: Could use inheritdoc here as well 😊

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +129,138 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
     public void WriteValue(object source, IGenericColumns target)
     {
-        // Convert and write JSON
-        var json = JsonConvert.SerializeObject(source, _jsonSettings);
```
1nf0rmagician, line 138: Do I get this correctly: Before, we wrote the mapped properties into the respective columns, but they also appeared in the json column 🤔

src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
```diff
@@ -18,35 +18,67 @@ internal class TextColumnMapper : ColumnMapper<string>
...
+                        var effectiveType = Nullable.GetUnderlyingType(Property.PropertyType) ?? Property.PropertyType;
+
+                        if (!effectiveType.IsValueType)
+                            return null;
+
+                        return Activator.CreateInstance(effectiveType);
+                    }
 
+                    return JsonConvert.DeserializeObject(s, Property.PropertyType, JsonSettings.Minimal);
+                });
```
1nf0rmagician, line 55: Nice that should actually fix the second bug we talked about 👍

src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
```diff
@@ -18,35 +18,67 @@ internal class TextColumnMapper : ColumnMapper<string>
...
         if (propType == typeof(Guid))
+        {
             return ((Guid)value).ToString();
+        }
+
+        // Complex reference types -> JSON
+        if ((propType.IsClass || propType.IsInterface) && propType != typeof(string))
+        {
+            if (value == null)
+                return null;
```
1nf0rmagician: Here I am unsure, isn't this covered by `JsonConvert.SerializeObject(value, Property.PropertyType, JsonSettings.Minimal)` automatically? Not sure what is written for null values in the json in the db, but something was already and it didn't cause issues did it? 🤔

src/Moryx.Products.Management/Moryx.Products.Management.csproj
```diff
@@ -7,6 +7,10 @@
     <PackageTags>MORYX;IIoT;IoT;Product;Manufacturing;Management;API</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Plugins\GenericStrategies\___FloatColumnMapper.cs" />
+  </ItemGroup>
+
```
1nf0rmagician: I guess this was commited by accident

src/Moryx.Products.Samples/Importer/GenericJsonTestProductTypeImporter.cs
```diff
@@ -0,0 +1,71 @@
...
+        };
+
+        return result;
+    }
+}
+
+public class SpecializedJsonTestProductTypeParameters : PrototypeParameters
+{
+    //[Required]
+    //public string Test { get; set; }
```
1nf0rmagician: Remove comment

src/Moryx.Products.Samples/ProductTypes/ComplexData.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Moryx.Products.Samples;
+
+
+/// <summary>
+/// To test complex data in text columns
+/// </summary>
+public class ComplexData
+{
+    // Property name "Name" is reserved: do not use!
+    //public string Name { get; set; }
```
1nf0rmagician: I though this was only an issue in the UI 😮 It should work to do this on DTO's held by a product type.

src/Moryx.Products.TestProducts/ProductTypes/TextColumnMapperTestProductType.cs
1nf0rmagician, line 1: I see the samples are very useful for manual testing. Could we bring these into test cases in the `ProductStorageTests` now? It uses an sqlite based inmemory db and might need some modifications to the test setup and test models, but then we would have regression tests on this 😊

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
fmoebert: I have reverted now all changes in FloatColumnMapper ans PropertyAccesor.
Investigation showed that the observed values originate from the exact IEEE754 representation of Single (float) properties when converted to the generic double storage columns.
The rounding logic has therefore been removed again. Persisting a rounded value would modify the original value and break save/load symmetry.
A separate discussion may be required whether generic float columns should continue to store float values via double columns or whether a different representation strategy should be introduced in a future release.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -54,8 +63,12 @@ public void Initialize(Type concreteType, IGenericMapperConfiguration config)
         var mapperIgnoredProperties = baseProperties
             .Concat(readOnlyProperties).ToArray();
 
-        _configuredMappers = config.PropertyConfigs.Where(pc => !mapperIgnoredProperties.Contains(pc.PropertyName))
-            .Select(pc => MapperFactory.Create(pc, concreteType)).ToArray();
+        //FMÖ: 22 07 2026
```
seveneleven: This type of comment can be removed, because the commit itself already provides this information.

src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
fmoebert: Yes, this intentionally filters out a property whose name matches the configured JsonColumn (e.g. "Text8").
The issue I observed was that the JsonColumn could be configured twice:
1. as a regular mapped property via PropertyConfigs
2. as the dedicated JSON fallback column
In that situation the same storage column was used both as plain text storage and as JSON storage. During loading the content of the column was interpreted as JSON, even when it previously contained a regular text value, which resulted in the JsonReaderException we investigated.
Therefore the JsonColumn must be reserved exclusively for JSON fallback storage and must not participate in the normal property mapping process.
The filter does not exclude every property backed by a text column. It only excludes the property whose name equals the configured JsonColumn.

Maybe it has a reference to SMAF-5422?

Without this filter, AutoConfigurator creates a regular mapping for Text8 and GenericEntityMapper also uses Text8 as JSON storage. This dual usage caused the persistence inconsistency that led to the exception.

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
fmoebert: Reverted FloatColumnMapper. So this point is obsolete

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: No, that's just because Visual Studio likes to refer to it as "conditional access."

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Thanks, I agree. The filter is intended as a safeguard for existing or manually edited configurations. Since such a configuration is inconsistent, I added a warning log entry when a PropertyConfig targeting the configured JsonColumn is detected. The mapping is then ignored to avoid using the same column both as regular string and JSON fallback storage.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: check!

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: The original cause should now be resolved thanks to the changes made to the “AutoConfigurator” and the “GenericEntityMapper.”
However, the exception may still occur if the data was saved before the fix, if database content was manually modified, or if other components write invalid JSON to the configured “JsonColumn.”
The catch block therefore no longer protects against the original configuration issue, but it still provides a much more meaningful error message than the plain `JsonException` when invalid JSON content is detected.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: check!

src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
fmoebert: That was the intention ;)

src/Moryx.Products.Management/Plugins/GenericStrategies/TextColumnMapper.cs
fmoebert: Good point. The explicit null handling was introduced while I was experimenting with keeping all properties visible in the persisted JSON structure.
However, the actual bug was related to deserialization of empty values, not serialization of null values. I agree that JsonConvert.SerializeObject() already handles null values correctly and the explicit check is not required for fixing the original issue.
My intention was to make JSON content more self-describing for manual inspection and editing, but that would require a different approach than this null check.

src/Moryx.Products.Management/Moryx.Products.Management.csproj
fmoebert: Yes, was an accident. I'm surprised that VS grabs and adds (backup) files copied in Explorer. The file doesn't even exist anymore.

src/Moryx.Products.Samples/ProductTypes/ComplexData.cs
fmoebert: Unfortunately, in my tests, I was actually unable to add a property named "Name" to ComplexData. The property is always overwritten. I posted this comment so that no one else would stumble across it.
But really, we need to figure out a way to solve this somehow

src/Moryx.Products.TestProducts/ProductTypes/TextColumnMapperTestProductType.cs
fmoebert: Yes, that will be a follow up

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: I tend to use these comments more for my own reference, especially when I'm not yet thoroughly familiar with the subject matter. But yes, of course they should be deleted before the final commit.
I've deleted all of them now

src/Moryx.Products.Samples/Importer/GenericJsonTestProductTypeImporter.cs
fmoebert: check!

src/Moryx/Tools/PropertyAccessor.cs
```diff
@@ -136,7 +137,18 @@ public ConversionAccessor(PropertyInfo property) : base(property)
     public TValue ReadProperty(TBase instance)
     {
         var value = (object)PropertyGetter((TConcrete)instance);
-        return (TValue)(value is TValue ? value : Convert.ChangeType(value, typeof(TValue)));
+        //return (TValue)(value is TValue ? value : Convert.ChangeType(value, typeof(TValue)));
```
1nf0rmagician: I agree this is more readable, just remove the comment please 😊

fmoebert: check!

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Not exactly.
The mapped properties were still written to their configured columns. However, the JsonColumn could also participate in the normal property mapping process. This could lead to the same storage column being used both as a regular mapped column and as JSON fallback storage.
As a consequence, properties that should have been exclusively stored in the JsonColumn were not handled correctly and, depending on the configuration, loading could later fail because the JsonColumn no longer contained valid JSON content.
Now: 
- The JsonColumn (currently Text8) is now reserved exclusively for the JSON fallback storage used by GenericEntityMapper.
- Regular property mappings are no longer assigned to the configured JsonColumn.
- Other text columns may still be used by TextColumnMapper. Depending on the mapped property type they can contain either:
  - plain text values (string), or 
  - JSON serialized complex objects.

1nf0rmagician (dismissed):
From my POV this looks good now, thank you 😊 

As discussed, we will proceed with the the mentioned changes for MORYX 12 on the future branch right after this branch was merged.

@dbeuchler to be able to merge you will have to take a look again as you have requested changes 😊

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
```diff
@@ -12,7 +16,11 @@ namespace Moryx.Products.Management;
 [Component(LifeCycle.Transient, typeof(IPropertyMapper), Name = nameof(FloatColumnMapper))]
 internal class FloatColumnMapper : ColumnMapper<double>
 {
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="targetType"></param>
```
dbeuchler: Fill or remove

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -37,25 +44,50 @@ public void Initialize(Type concreteType, IGenericMapperConfiguration config)
...
+        // Historically some configurations used the JsonColumn both as JSON storage
+        // and as a regular property mapping. This remains supported until the next
+        // major release but should no longer be generated automatically.
+        if (config.PropertyConfigs.Any(pc =>
+                string.Equals(
+                    pc.PropertyName,
+                    config.JsonColumn,
+                    StringComparison.OrdinalIgnoreCase)))
+        {
+            Logger?.LogWarning(
```
dbeuchler: Logger cannot be null here -> remove `?`

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -37,25 +44,50 @@ public void Initialize(Type concreteType, IGenericMapperConfiguration config)
...
+
+        // Legacy compatibility:
+        // Historically some configurations used the JsonColumn both as JSON storage
+        // and as a regular property mapping. This remains supported until the next
+        // major release but should no longer be generated automatically.
+        if (config.PropertyConfigs.Any(pc =>
+                string.Equals(
+                    pc.PropertyName,
+                    config.JsonColumn,
+                    StringComparison.OrdinalIgnoreCase)))
```
dbeuchler: Format, keep in one line until 120 characters

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +148,133 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
         return Expression.Lambda(body, columnParam) as Expression<Func<IGenericColumns, bool>>;
     }
 
+    /// <inheritdoc />
     public void ReadValue(IGenericColumns source, object target)
     {
-        // Use all configured mappers
-        var properties = source;
+        ArgumentNullException.ThrowIfNull(source);
+        ArgumentNullException.ThrowIfNull(target);
```
dbeuchler: We follow the principle "trust yourself", so system bondaries should be validated, but then we trust that we dont get nulls. This internal mapper should not be aware of nulls for the source and target here.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +148,133 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
+        // Older systems may contain plain text in the JsonColumn.
+        // In this case simply ignore the content and continue loading
+        // the directly mapped properties.
+        var trimmed = json.Trim();
+        if (!(trimmed.StartsWith("{") || trimmed.StartsWith("[")))
+        {
+            Logger?.LogWarning(
+                "Ignoring non-JSON content in JsonColumn '{Column}'. Value: '{Value}'",
+                _jsonAccessor.Property.Name,
+                json);
```
dbeuchler: Same here, Remove `?`, keep 120 chars per line

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +148,133 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
+            }
+
+            // jump Indexer 
+            if (prop.GetIndexParameters().Length > 0)
+            {
+                continue;
+            }
+
+            // jump Read-only Props
+            if (prop.GetSetMethod() == null)
```
dbeuchler, line 228: CanWrite?

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+using Moryx.Modules;
+
+// ReSharper disable AsyncMethodWithoutAwait
```
dbeuchler: dont make it async, return Task.Result instead.

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Moryx.Products.Samples;
+
+[ExpectedConfig(typeof(GenericJsonTestProductTypeImporterConfig))]
+[ProductImporter(nameof(GenericJsonTestProductTypeImporter))]
+public class GenericJsonTestProductTypeImporter : ProductImporterBase<GenericJsonTestProductTypeImporterConfig, SpecializedJsonTestProductTypeParameters>
+{
+    protected override async Task<ProductImporterResult> ImportAsync(ProductImportContext context, SpecializedJsonTestProductTypeParameters parameters,
+        CancellationToken cancellationToken)
+    {
+
```
dbeuchler: empty line

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.Samples.Importer;
+
+public class TextColumnMapperTestProductTypeImporter :
+    ProductImporterBase<
+        TextColumnMapperTestProductTypeImporterConfig,
+        TextColumnMapperTestProductTypeParameters>
+{
```
dbeuchler: Format

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
```diff
@@ -0,0 +1,43 @@
...
+    protected override Task<ProductImporterResult> ImportAsync(
+        ProductImportContext context,
+        TextColumnMapperTestProductTypeParameters parameters,
+        CancellationToken cancellationToken)
```
dbeuchler: Format

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporterConfig.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Runtime.Serialization;
+using Moryx.AbstractionLayer.Products;
+using Moryx.Products.Samples.Importer;
+
+namespace Moryx.Products.Samples;
+
+[DataContract]
+public class TextColumnMapperTestProductTypeImporterConfig
+    : ProductImporterConfig
+{
```
dbeuchler: Format

src/Moryx.Products.TestProducts/ProductInstances/GenericJsonTestProductTypeInstance.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+// ReSharper disable RedundantTypeDeclarationBody
```
dbeuchler: What is redundant here?

src/Moryx.Products.TestProducts/ProductInstances/TextColumnMapperTestProductTypeInstance.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.Samples;
+
+public class TextColumnMapperTestProductTypeInstance
+    : ProductInstance<TextColumnMapperTestProductType>
```
dbeuchler: Format

src/Moryx.Products.TestProducts/ProductTypes/ComplexData.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+namespace Moryx.Products.Samples;
+
+
```
dbeuchler, line 5: Format

src/Moryx.Products.Samples/ProductTypes/ComplexData.cs
dbeuchler: There is an open ticket for it, dont let this here like that, use //TODO and refere to the github ticket.

src/Moryx.Products.TestProducts/ProductTypes/GenericJsonTestProductType.cs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using System.ComponentModel;
+using Moryx.AbstractionLayer.Products;
+
+// ReSharper disable IdentifierTypo
+// ReSharper disable UnusedMember.Global
```
dbeuchler: Check if necessary

src/Moryx.Products.TestProducts/ProductTypes/GenericJsonTestProductType.cs
```diff
@@ -0,0 +1,141 @@
...
+
+namespace Moryx.Products.Samples;
+
+[DisplayName("Generic Json Test type")]
+public class GenericJsonTestProductType : ProductType
+{
+
```
dbeuchler, line 12: empty line

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.Samples.Importer;
+
+public class TextColumnMapperTestProductTypeImporter :
```
dbeuchler: All TestColumnMapperTestProductXX , these are tests, correct? Could we move them to isolated unit/integration tests? Otherwise the `Moryx.Products.Samples` project will be blown with test-code and not with real world samples.

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
```diff
@@ -134,7 +134,38 @@ public async Task PrepareStorage()
...
+                            PropertyName = nameof(TextColumnMapperTestProductType.Integer1),
+                            Column = nameof(IGenericColumns.Integer1),
+                            PluginName = nameof(IntegerColumnMapper)
+                        },
+
+                        new PropertyMapperConfig
+                        {
+                            PropertyName = nameof(TextColumnMapperTestProductType.Text1),
+                            Column = nameof(IGenericColumns.Text1),
+                            PluginName = nameof(TextColumnMapper)
```
dbeuchler, line 169: Ah it is already part of the tests. So moving the classes to the IntegrationTests assembly would be nice.

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
```diff
@@ -952,4 +983,178 @@ public async Task SaveAndLoadInstance()
         Assert.That(byType5.Count, Is.GreaterThanOrEqualTo(1));
         Assert.That(byType6.Count, Is.GreaterThanOrEqualTo(1));
     }
-}
\ No newline at end of file
+
+
```
dbeuchler: empty line

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
```diff
@@ -952,4 +983,178 @@ public async Task SaveAndLoadInstance()
...
+            ComplexData1 = null,
+            ComplexData2 = null
+        };
+
+        var id = await _storage.SaveTypeAsync(product);
+
+        // Act
+        var loaded =
+            (TextColumnMapperTestProductType)await _storage.LoadTypeAsync(id);
```
dbeuchler: format

dbeuchler (commented):
First review comments from my personal static code analysis :D will check this deeper next...

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
1nf0rmagician: Since this component can be called by external plugins, doesn't it make this a boundary of the system? 🤔

src/Moryx.Products.Samples/ProductTypes/ComplexData.cs
fmoebert: Replaced by " //TODO: see SMAF-5082: Property 'Name' is filtered"

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
fmoebert: Removed

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Removed

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: done

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Removed the exceptions

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Done

src/Moryx.Products.TestProducts/ProductTypes/GenericJsonTestProductType.cs
fmoebert: Done

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
fmoebert: Yeah, that's a good idea. But I still need to figure out the exact process, since I'm also using these test products for "manual testing."

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
fmoebert: Done

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
fmoebert: Done

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
fmoebert: The test product types are not samples. They exist to reproduce specific bugs and to support automated and manual regression testing.
I propose moving them into a dedicated assembly **Moryx.Products.TestTypes**.
This keeps Moryx.Products.Samples focused on training and example content while still allowing the test types to be discovered by reflection and used through the UI whenever the test assembly is deployed.
Is this an approach that complies with the rules?

src/Moryx.Products.Samples/ProductTypes/ComplexData.cs
dbeuchler: do not expose internal ticket numbers, use github issue instead

fmoebert: replaced: "// ToDo: Property name "Name" is reserved and is covered by Parent class property so it can not be used here"

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
fmoebert: Moved the TestProducts from 'Moryx.Products.Samples' to 'Moryx.Products.TestProducts'.
So the Products can be enabled or disabled in StartProj.Asp.csproj by changing ProjectReferences

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
fmoebert: Moved test products to an extra projekt. See other remark

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
fmoebert: check!

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
fmoebert: check!

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
fmoebert: check!

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporterConfig.cs
fmoebert: check!

src/Moryx.Products.TestProducts/ProductInstances/GenericJsonTestProductTypeInstance.cs
fmoebert: I guess that's a remnant. Check

src/Moryx.Products.TestProducts/ProductInstances/TextColumnMapperTestProductTypeInstance.cs
fmoebert: check

src/Moryx.Products.TestProducts/ProductTypes/ComplexData.cs
fmoebert: check

src/Moryx.Products.TestProducts/ProductTypes/GenericJsonTestProductType.cs
fmoebert: I guess that's a remnant. Check

src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
```diff
@@ -179,28 +190,43 @@ private TConfig StrategyConfig<TStrategy, TConfig, TBaseType>(Type targetType)
...
+                        !string.Equals(p.Name, jsonColumn, StringComparison.OrdinalIgnoreCase))
             .ToList();
```
dbeuchler, line 221: This part gets longer and longer - maybe we simplify it a bit: Here is a proposal from me:

First option: Combine into a single Where

```csharp
var filteredProperties = targetProperties
    .Where(p => baseProperties.All(bp => bp.Name != p.Name)
        && p.GetSetMethod() != null
        && !typeof(ProductPartLink).IsAssignableFrom(p.PropertyType)
        && !typeof(IEnumerable<ProductPartLink>).IsAssignableFrom(p.PropertyType)
        && !typeof(ProductInstance).IsAssignableFrom(p.PropertyType)
        && !typeof(IEnumerable<ProductInstance>).IsAssignableFrom(p.PropertyType)
        && (string.IsNullOrEmpty(jsonColumn)
            || !string.Equals(p.Name, jsonColumn, StringComparison.OrdinalIgnoreCase)))
    .ToList();
```

Scnd option: Extract excluded types into an array

```csharp
var excludedTypes = [
    typeof(ProductPartLink), 
    typeof(IEnumerable<ProductPartLink>),
    typeof(ProductInstance), 
    typeof(IEnumerable<ProductInstance>)
];

var filteredProperties = targetProperties
    .Where(p => baseProperties.All(bp => bp.Name != p.Name)
        && p.GetSetMethod() != null
        && !excludedTypes.Any(t => t.IsAssignableFrom(p.PropertyType))
        && (string.IsNullOrEmpty(jsonColumn)
            || !string.Equals(p.Name, jsonColumn, StringComparison.OrdinalIgnoreCase)))
    .ToList();
```

The second version is more concise and easier to extend if more types need excluding.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -37,25 +44,46 @@ public void Initialize(Type concreteType, IGenericMapperConfiguration config)
         _jsonAccessor = ReflectionTool.PropertyAccessor<IGenericColumns, string>(jsonColumn);
 
         var baseProperties = typeof(TBase).GetProperties().Select(p => p.Name).ToArray();
-        var configuredProperties = config.PropertyConfigs.Select(cm => cm.PropertyName);
+
+        // Legacy compatibility:
+        // Historically some configurations used the JsonColumn both as JSON storage
+        // and as a regular property mapping. This remains supported until the next
+        // major release but should no longer be generated automatically.
```
dbeuchler, line 51: Could you add a TODO here, that we can find this place in MORYX 12?

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +144,127 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
-            mapper.ReadValue(properties, target);
+            mapper.ReadValue(source, target);
         }
 
-        // Fill the rest from JSON
+        // Reading JSON from the JsonColumn
         var json = _jsonAccessor.ReadProperty(source);
-        if (!string.IsNullOrEmpty(json))
-            JsonConvert.PopulateObject(json, target, _jsonSettings);
+        if (string.IsNullOrWhiteSpace(json))
```
dbeuchler, line 158: I would use `IsNullOrEmpty` here as before. If there are whitespace characters, it can be part of the a content.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +144,127 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
+        {
+            return;
+        }
+
+        // Legacy compatibility:
+        // Older systems may contain plain text in the JsonColumn.
+        // In this case simply ignore the content and continue loading
+        // the directly mapped properties.
+        var trimmed = json.Trim();
```
dbeuchler, line 167: As I said, I would treat whitespace characters as content. So use `IsNullOrEmpty` and without trim sounds the best compatible for me.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -116,32 +144,127 @@ private Expression<Func<IGenericColumns, bool>> Convert(string memberName, Expre
...
     }
 
+    /// <inheritdoc/>>
     public void WriteValue(object source, IGenericColumns target)
     {
-        // Convert and write JSON
-        var json = JsonConvert.SerializeObject(source, _jsonSettings);
-        _jsonAccessor.WriteProperty(target, json);
+        ArgumentNullException.ThrowIfNull(source);
+        ArgumentNullException.ThrowIfNull(target);
```
dbeuchler, line 192: Same here, follow the strategy "Trust yourself", no Argument-Checks here.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
dbeuchler: > The JsonColumn (currently Text8) is now reserved exclusively for the JSON fallback storage used by GenericEntityMapper.

Where do I find this limit? You meant that you cannot use the JsonColumn and property mapping for the same column, correct? But the JsonColumn can still be configured, right?

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -2,12 +2,16 @@
...
+using Microsoft.Extensions.Logging;
 using Moryx.Container;
 using Moryx.Products.Management.Model;
 using Moryx.Serialization;
 using Moryx.Tools;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using static Moryx.Products.Management.ProductExpressionHelpers;
+// ReSharper disable GrammarMistakeInComment
+// ReSharper disable MergeIntoPattern
```
dbeuchler, line 14: Can you check these disabled checks?

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+using Moryx.Modules;
+
+
```
dbeuchler, line 7: empty line

src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
```
dbeuchler, line 5: Could you remove this line and add summaries to the classes?

src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Moryx.AbstractionLayer\Moryx.AbstractionLayer.csproj" />
+    <ProjectReference Include="..\Moryx.Products.Management\Moryx.Products.Management.csproj" />
+    <ProjectReference Include="..\Moryx.TestModule\Moryx.TestModule.csproj" />
```
dbeuchler, line 12: Why is the TestModule referenced?

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: added
// TODO: MORYX 12: Maybe remove support for using JsonColumn as a regular property mapping.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: jepp, done

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: done

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Done

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: removed the checks

src/Moryx.Products.TestProducts/Importer/GenericJsonTestProductTypeImporter.cs
fmoebert: done

src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
fmoebert: Don't know why I did this....
Removed it

src/Moryx.Products.TestProducts/Moryx.Products.TestProducts.csproj
fmoebert: Yes, did it now

src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
fmoebert: Please check

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Yes, that's exactly what I meant.
The JsonColumn itself can still be configured and is still used by GenericEntityMapper for JSON fallback storage.
The restriction applies only to using the same column both as a configured JsonColumn and as a regular property mapping. Existing configurations are still supported for backward compatibility, but this combination should no longer be automatically generated.

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
fmoebert: Since that's in the write method, I've now added "if (!prop.CanWrite)"

src/Moryx.Products.TestProducts/Importer/TextColumnMapperTestProductTypeImporter.cs
fmoebert: please check

src/Moryx.Products.Management/Implementation/Storage/StrategyBase.cs
```diff
@@ -24,7 +24,17 @@ public class StrategyBase<TConfig, TConfigBase> : IAsyncConfiguredInitializable<
...
+        if (config is not TConfig typedConfig)
+        {
+            throw new InvalidOperationException(
+                $"Configuration mismatch for plugin '{config.PluginName}' " +
+                $"and target type '{config.TargetType}'. " +
+                $"Expected configuration type '{typeof(TConfig).Name}', " +
+                $"but received '{config.GetType().Name}'. " +
+                $"Please recreate the strategy configuration in the ControlCenter.");
+        }
+
```
dbeuchler: This was already a part of a different PR and was already reworked again, could you check that?

fmoebert: Oh, I think you mean https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1380/changes/6c1c24517bcee4bef576de29d79533019711aa3c?
I guess I just carried that over / didn't rebase it properly to dev. Should I undo that here?

src/Moryx.Products.Management/Implementation/Storage/AutoConfigurator.cs
```diff
@@ -169,48 +179,73 @@ private TConfig StrategyConfig<TStrategy, TConfig, TBaseType>(Type targetType)
...
             if (propertyTuple == null)
+            {
                 continue;
+            }
 
             var strategy = propertyTuple.Item1;
             var propertyConfig = propertyTuple.Item2;
             propertyConfig.PropertyName = property.Name;
-            propertyConfig.PluginName = strategy.GetCustomAttribute<ComponentAttribute>().Name;
+            propertyConfig.PluginName = CustomAttributeProviderExtensions.GetCustomAttribute<ComponentAttribute>(strategy).Name;
```
1nf0rmagician: I don't know why, but for some reason, the use of the extension method as an extension method was changes here as well 🤔 Search for other occasions of the `CustomAttributeProviderExtensions` it happened multiple times 🙈

src/Moryx.Products.Management/Implementation/Storage/StrategyBase.cs
1nf0rmagician: Yeah please remove it here. #1418 fiexed this for all initialized plugins 😊

1nf0rmagician (changes requested):

src/Moryx.Products.Management/Plugins/GenericStrategies/FloatColumnMapper.cs
1nf0rmagician, line 1: Not sure why, but this slipped in as well I guess

1nf0rmagician (approved):

dbeuchler (approved):

Issue #911, opened 2025-12-09
Support custom IIdentity in ProductManagement


dbeuchler:
**Is your feature request related to a problem? Please describe.**
A developer wants to use a custom IIdentity with their products. The ProductType is already based on IIdentity, but the facade and storage api is restricted to IIdentity.

PR #1384, opened 2026-08-04
Changed the Regex of Identifier in ProductIdentity

fmoebert:
For his project, Sascha needs the ability to specify an alphanumeric identifier for a customer-specific item after the 7-digit item/material number. This identifier should be separated from the item number by a hyphen.
The change now allows for 
- a numeric item/material number with a minimum of 7 digits and a maximum of 12 digits
- optionally followed by a hyphen and one to twenty alphanumeric characters, excluding punctuation or special characters
The revision of the product type remains unaffected by this change

### DIFF
--- a/src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
+++ b/src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
@@ -16,10 +16,8 @@ public class PrototypeParameters
     /// </summary>
     [DefaultValue("2901234")]
     [Description("Identifier of the new product"), Required]
-    //[StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
-    [StringLength(35)]
-    [RegularExpression(@"^\d{7,12}(-[A-Za-z0-9]{1,20})?$")]
-    public string Identifier { get; set; }
+    [StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
+    public virtual string Identifier { get; set; }
 
     /// <summary>
     /// Revision of the new product
--- a/src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
+++ b/src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
@@ -16,7 +16,9 @@ public class PrototypeParameters
     /// </summary>
     [DefaultValue("2901234")]
     [Description("Identifier of the new product"), Required]
-    [StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
+    //[StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
+    [StringLength(35)]
+    [RegularExpression(@"^\d{7,12}(-[A-Za-z0-9]{1,20})?$")]
     public string Identifier { get; set; }
 
     /// <summary>
--- a/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
+++ b/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
@@ -40,11 +40,24 @@ public class ProductIdentity : IIdentity
     /// <returns></returns>
     public static ProductIdentity Parse(string identityString)
     {
-        Regex rx = new Regex(@"(?<identifier>\w+)-(?<revision>\d+)");
-        if (!rx.IsMatch(identityString))
-            throw new FormatException("identityString should consist of <identity>-<revision> instead of " + identityString);
-        var groups = rx.Match(identityString).Groups;
-        return new ProductIdentity(groups["identifier"].Value, Convert.ToInt16(groups["revision"].Value, CultureInfo.InvariantCulture));
+        var separator = identityString.LastIndexOf('-');
+
+        if (separator < 1 || separator == identityString.Length - 1)
+        {
+            throw new FormatException(
+                $"identityString should consist of <identity>-<revision> instead of {identityString}");
+        }
+
+        var identifier = identityString[..separator];
+        var revisionPart = identityString[(separator + 1)..];
+
+        if (!short.TryParse(revisionPart, out var revision))
+        {
+            throw new FormatException(
+                $"identityString should consist of <identity>-<revision> instead of {identityString}");
+        }
+
+        return new ProductIdentity(identifier, revision);
     }
 
     /// <summary>
--- a/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
+++ b/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
@@ -10,7 +10,7 @@ namespace Moryx.AbstractionLayer.Products;
 /// <summary>
 /// Identity for products consisting of identifier and revision
 /// </summary>
-public class ProductIdentity : IIdentity
+public partial class ProductIdentity : IIdentity
 {
     /// <summary>
     /// Const value representing the latest revision of a product. This is intended for later changes of this constant.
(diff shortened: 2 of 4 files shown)


dbeuchler (commented):
this is a behavior change which we cannot introduce in 10.x and I would not do that in 12 too because its limits for internal requirements. Use inherited parameters instead.

src/Moryx.Products.Web/package-lock.json
1nf0rmagician, line 1: Please remove this file 😊

src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
```diff
@@ -16,7 +16,9 @@ public class PrototypeParameters
     /// </summary>
     [DefaultValue("2901234")]
     [Description("Identifier of the new product"), Required]
-    [StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
+    //[StringLength(7, MinimumLength = 7), RegularExpression(@"\d+")]
+    [StringLength(35)]
+    [RegularExpression(@"^\d{7,12}(-[A-Za-z0-9]{1,20})?$")]
```
1nf0rmagician, line 21: **TLDR: We should change something, but not in the currently proposed way**

While I agree with @dbeuchler that this is a behaviour change we cannot introduce in this way (putting limitations on this string that are more strict than before), I disagree with the statement that a change of this behaviour is generally undesirable.

The following arguments (and subsequent conclusion) come to my mind:
1. The limitation to require a minimum of 7 digits and allow only for digits in the default importer of this repository is very arbitrary. Especially considering, that this limitation is not at all reflected in the [ProductIdentity](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs?plain=1#L13)
2. The [Parse method](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs?plain=1#L43) actually already allows for any word characters in its regex
3. This limitation is however only applied when parsing the identity, while setting it in the first place allows for any arbitrary string
4. The [ProductContext](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.Products.Management/Model/Entities/ProductTypeEntity.cs#L13) and underlying [IIdentity](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.AbstractionLayer/Identity/IIdentity.cs) actually do not impose any limitations aside from the fact that it must be a string (even letting aside that the ProductContext is an implementation detail of the module not the public API)
5. The [ProductImporterBase](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.AbstractionLayer/Products/Import/ProductImporterBase.cs) does not even require to make use of the `PrototypeImportParameters`, even though it is probably used in most places as the underlying base.

_Note: I'm not even sure I captured all places where we impose contradicting limitations on a `ProductIdentity`_

Hence, I conclude that the most general but optional parameter base class does not set limitations on possible product identifiers with good reason and we should soften this limitation.  
I propose that...
- the parse method in the `ProductIdentity` should accept any sanitized string as a product identifier only relying on the last `-` character to denote that the revision is now starting
- the `PrototypeParameters` in turn should also allow for any sanitized string to be set as product identifer

Even if we can agree on this conclusion based on the arguments above, I see that there is an argument to be made for applications relying on either of the two limitation.
- An application could inherit from the `PrototypeParameters` for their importer and rely on the UI validation limiting the possible input to be at least 7 characters long and contain only-digits.  
_I would agree with this argument and hence leave the limitation on the parameters as is until MORYX 12_
- An application could rely on the fact that the parse method only allows word characters in the identifier and take the first `-` to be the sign for a starting revision  
_I would NOT agree with this argument, as the limitation is only imposed on parsing an identifier not when setting it in the first place in the `ProductIdentity`. I would allow for changing this regex._

An option to make all proposed changes already in MORYX 10 (preferable if you ask me) could consider the fact that #911 already asks for allowing application specific `IIdentity` implementations. With this issue in mind I propose a configurable `ProductIdentityValidation` plugin, which verifies the `IIdentity` on a ProductType before saving it to the database (we could have a seperate argument whether by default in MORYX 10 a validation for identifiers only containing word characters should be put in place). 
In addition, making the `PrototypeImportParameters.Identifier` property virtual would be a non breaking change that allows applying different validation on the field itself for the entryserialize based in field validation in the UI.

1nf0rmagician (changes requested):

src/Moryx.Products.Web/package-lock.json
fmoebert: check

src/Moryx.AbstractionLayer/Products/Import/PrototypeImportParameters.cs
dbeuchler: Thanks a lot for taking the time to break it all down ;the PR was missing that context. I hope I have understood everything correctly:

- We keep the limitations on `PrototypeImportParameters` in MORYX 10 ✅
- We make the `Identifier` of `PrototypeImportParameters` `virtual` to make it overidable in apps ✅
- We soften the limits in the `Parse` method of `ProductIdentity` ✅
- Some component which could validate an `ProductIdentity` in MORYX 12 - to be discussed how -> new Issue

Sounds good so far. @fmoebert Would this be okay for you too?

fmoebert: Personally, that's perfectly fine with me!

fmoebert: Now PrototypeParameters Identifier is overwritable, Identifier-Parsing is away from RegEx and searches ffor last '-'.
Tests are available for
- Parsing Custom Identifiers OK
- Parsing Custom Identifiers KO -> Exception
- TryParse Custom Identifier

src/Moryx.Container/ComponentSelectors/ConfigBasedComponentSelector.cs
```diff
@@ -127,9 +127,8 @@ private static void ExecuteInitialize(Type componentType, Type configType, objec
         }
         catch (Exception e)
         {
-            kernel.Logger.ErrorFormat(e, CultureInfo.InvariantCulture,
-                "Error during initialization of component {0} with config {1}: {2}",
-                componentType.FullName, configType.FullName);
+            var invocationException = HandleInitializeInvocationException(componentType, configType, kernel, e);
```
dbeuchler, line 130: Now this pr has a lot of changes of other branches. Can you remove them?

src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
```diff
@@ -40,11 +40,24 @@ public ProductIdentity(string identifier, short revision)
...
+        var identifier = identityString[..separator];
+        var revisionPart = identityString[(separator + 1)..];
+
+        if (!short.TryParse(revisionPart, out var revision))
+        {
+            throw new FormatException(
+                $"identityString should consist of <identity>-<revision> instead of {identityString}");
+        }
+
+        return new ProductIdentity(identifier, revision);
```
dbeuchler: You dropped CultureInfo.InvariantCulture. The new short.TryParse(revisionPart, out var revision) uses CurrentCulture and default NumberStyles.Integer, so it's culture-sensitive and now accepts "abc- 5" and "abc-+5". Adding NumberStyles.None, CultureInfo.InvariantCulture fixes that.

An more enhanced but simpler implementation could be: 

````csharp
[GeneratedRegex(@"^(?<identifier>.+)-(?<revision>\d+)$")]
  private static partial Regex IdentityPattern();

  public static ProductIdentity Parse(string identityString)
  {
      var match = IdentityPattern().Match(identityString);
      if (!match.Success || !short.TryParse(match.Groups["revision"].Value,
              NumberStyles.None, CultureInfo.InvariantCulture, out var revision))
          throw new FormatException($"identityString should consist of <identity>-<revision> instead of {identityString}");

      return new ProductIdentity(match.Groups["identifier"].Value, revision);
  }
````

dbeuchler (changes requested):

src/Moryx.Container/ComponentSelectors/ConfigBasedComponentSelector.cs
fmoebert: I have checked out the actual dev before I worked on that. Maybe that's a problem?
Edit: Now it shoud be clean rebased.

dbeuchler (changes requested):
Was it a mistake that you removed the tests from ProductStorageTests? Maybe we should move the in separate test file "ProductIdentityTests"?

dbeuchler (approved):

1nf0rmagician (approved):

Issue #418, opened 2024-05-08
ProductQuery does not return full type information


dbeuchler:
**Describe the bug**
Given is the following product query: 

````cs
var productTypes = ProductStorage.LoadTypes(new ProductQuery()
{
    Type = typeof(SomeType).FullName
}).Cast<SomeType>().ToArray();
````

I expect that the product types have all values from the storage but only the base type information are filled. I found the following code in the [ProductStorage](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/5a5eda7770c0075fade909d10a63b2f8f623a37b/src/Moryx.Products.Management/Implementation/Storage/ProductStorage.cs#L383)

````cs
return products.Where(p => TypeInformation.ContainsKey(p.TypeName)).Select(p =>
{
    return TypeInformation[p.TypeName].CreateTypeFromEntity(p);
}).ToList();
````

This is only loading the base type information. All of the other `LoadType` methods return a full product with all information. 

Proposal:
- A: Return full type information
- B: Extend `ProductQuery` to return full type information

**Additional context**
There is no comment nor hint that this method only returns the base type information instead of full product.

dbeuchler:
done by #1396

PR #1396, opened 2026-08-06
Support full type information on LoadTypes

fmoebert:
SMAF-5161
LoadTypesAsync(ProductQuery) returned partially initialized ProductType instances by calling CreateTypeFromEntity().
Other load methods return fully populated products via Transform().
Replaced the lightweight creation path with the existing Transform() logic and added an integration test that verifies type-specific properties are available after querying products.


dbeuchler (changes requested):
For reference, here is the public issue this relates to: #418.  It iss generally more helpful for other contributors if you link publicly available tickets rather than internal ticket numbers, so everyone can follow the context.

From what I can see, the change now always returns the full type information. Thats good so far! My only thought is that it would be nice to make it optional. I know of applications where the types themselves are quite large, and loading all the full type information there can take a long time. Also the UI requesting the product types only need the basic information for the tree.

What do you think about adding an option to the product query to control whether the full type information is returned (proposed in #418)? The default would keep the current behavior, and we could flip it in MORYX 12.

---

There is also an issue in your commits, it contains a lot of changes from other branches. Maybe a cleanup is necessary.

fmoebert:
> For reference, here is the public issue this relates to: #418. It iss generally more helpful for other contributors if you link publicly available tickets rather than internal ticket numbers, so everyone can follow the context.
> 
> From what I can see, the change now always returns the full type information. Thats good so far! My only thought is that it would be nice to make it optional. I know of applications where the types themselves are quite large, and loading all the full type information there can take a long time. Also the UI requesting the product types only need the basic information for the tree.
> 
> What do you think about adding an option to the product query to control whether the full type information is returned (proposed in #418)? The default would keep the current behavior, and we could flip it in MORYX 12.
> 
I agree that large product hierarchies may benefit from lightweight queries.
For this fix I intentionally implemented option A from #418 because it is the smallest change that restores the behavior expected by the issue and by the other LoadType methods.
I would prefer discussing a lightweight/full-load option separately because it changes the public query contract and introduces new API behavior beyond the bugfix itself.
If the team prefers option B, I can adjust the implementation accordingly after my vacation.



> There is also an issue in your commits, it contains a lot of changes from other branches. Maybe a cleanup is necessary.

I used "my" TestProducts for Integration Test, so you might see that in this way.
Now I changed the Test to a product of the samples and deleted my Moryx.TestProducts-Project

fmoebert:
While validating the last change I noticed an intermittent failure in **DriverResult_ThrowsException_ShouldNotDisruptModule**.
The test passes when executed in isolation most of the time, but occasionally fails with:
**CountdownEvent.AddCount()**: The event is already signaled and cannot be incremented.
The failure occurs in **NotSoParallelOps.ScheduleExecution()** and appears unrelated to the **LoadTypesAsync(ProductQuery)** changes.

dbeuchler:
> While validating the last change I noticed an intermittent failure in DriverResult_ThrowsException_ShouldNotDisruptModule.
The test passes when executed in isolation most of the time, but occasionally fails with:
CountdownEvent.AddCount(): The event is already signaled and cannot be incremented.
The failure occurs in NotSoParallelOps.ScheduleExecution() and appears unrelated to the LoadTypesAsync(ProductQuery) changes.

Known problem, just restart the tests

dbeuchler:
> I would prefer discussing a lightweight/full-load option separately because it changes the public query contract and introduces new API behavior beyond the bugfix itself.

When we always fully load the type, this would mean that the client gets all information instead of only the basic head. Sample: Loading the heads in my product tree of an application takes ~150ms. The full load takes >4 seconds.

Extending the ProductQuery type should be not much effort?

fmoebert:
Good point. I initially implemented Option A as the simplest solution. I've now switched to Option B from #418 and added a query flag so that the previous behavior remains unchanged, while callers can explicitly request fully populated product types.

src/Moryx.AbstractionLayer/Products/ProductQuery.cs
```diff
@@ -77,6 +77,9 @@ public class ProductQuery
     /// </summary>
     [DataMember]
     public List<PropertyFilter> PropertyFilters { get; set; }
+
+    [DataMember]
+    public bool LoadFullTypeInformation { get; set; }
 }
```
dbeuchler, line 87: Add summary, update UI API

src/Moryx.ControlSystem.ProcessEngine.Web/app/package.json
dbeuchler, line 26: Could you check all the package updates? seems to come from another pull request/branch ....

dbeuchler (changes requested):
also conflicts must be resolved to be mergable to dev

src/Moryx.AbstractionLayer/Products/ProductQuery.cs
fmoebert: I added the XML summary to ProductQuery.
Regarding the "update UI API" comment: ProductQuery is used directly by the types/query endpoint and the new property is part of the data contract. I did not find a dedicated UI API model for this query.
Did you have a specific API model or generated client in mind that should be updated as well?

src/Moryx.ControlSystem.ProcessEngine.Web/app/package.json
fmoebert: There was a test according to "Changing the product Instance strategy in CommandCenter thrown an exception". I deleted it

src/Moryx.AbstractionLayer/Products/ProductQuery.cs
dbeuchler: Yes its directly used in the endpoint, so the UI API should also be updated. run `npm run generate-api` for the products-web project. Maybe you have to adjust the command temporary when working in cmd or powershell.

fmoebert: The generated ProductQuery model now also contains loadFullTypeInformation.

src/Moryx.Products.Management/Implementation/Storage/ProductStorage.cs
```diff
@@ -371,10 +371,21 @@ public Task<IReadOnlyList<ProductType>> LoadTypesAsync(ProductQuery query, Cance
...
 
-        var results = products.Where(p => _typeInformation.ContainsKey(p.TypeName))
-            .Select(p => _typeInformation[p.TypeName].CreateTypeFromEntity(p)).ToArray();
+        if (query.LoadFullTypeInformation)
+        {
+            var transformed = products
+                .Where(p => _typeInformation.ContainsKey(p.TypeName))
+                .Select(p => Transform(uow, p, true, cancellationToken));
+
+            return await Task.WhenAll(transformed);
```
dbeuchler, line 381: is there a reason why you do not return the task directly and await it?

src/Moryx.Products.Web/app/src/app/api/models/product-importer.ts
```diff
@@ -8,6 +8,7 @@ import { Entry } from '@moryx/ngx-web-framework/entry-editor';
 
 /* Code generated by ng-openapi-gen DO NOT EDIT. */
 
+
```
dbeuchler: Do you know why there are so much empty lines?

src/Moryx.Products.Web/app/src/app/api/models.ts
```diff
@@ -6,6 +6,13 @@
 /* eslint-disable */
 /* Code generated by ng-openapi-gen DO NOT EDIT. */
 
+
+
+
+
+
+
+
```
dbeuchler: same here, empty lines... issue in some generator?

src/Moryx.Products.Web/app/src/app/api/models/product-importer.ts
fmoebert: The additional blank lines were introduced by the API regeneration step. Looking at the diff, every generated file got an additional copyright header block, which also introduced the extra spacing.

The only intended API change is the new loadFullTypeInformation property in product-query.ts. I'll investigate whether the formatting changes are caused by ng-openapi-gen or by fix-openapi-gen.js and whether they can be removed from the PR.

src/Moryx.Products.Web/app/src/app/api/models.ts
fmoebert: I think, same problem

dbeuchler: maybe there is a fix in the js but its okay for me for now

dbeuchler (approved):
close #418

PR #1410, opened 2026-08-07
increase visual instruction result `button` on big screens

MathoMathiasCamara:
Increase the size of result buttons in the Visual instruction view on big screens

increase button in visual instructuinbs

### DIFF
--- a/src/Moryx.VisualInstructions.Web/app/src/app/components/worker-instructions/worker-instructions.html
+++ b/src/Moryx.VisualInstructions.Web/app/src/app/components/worker-instructions/worker-instructions.html
@@ -54,7 +54,7 @@
         <div class="result-actions">
           @if(displayedInstruction()?.type === InstructionType.Execute){
             @for (result of displayedInstruction()?.results; track result) {
-              <button  mat-stroked-button
+              <button  mat-flat-button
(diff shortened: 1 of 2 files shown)


1nf0rmagician:
Wouldn't it be necessary to increase the size of all of the content in this case? 🤔 I don't have an idea for a good concept yet but using the input fields in this case seams to also be an issue.

Also we should consider that the OS and the browser already have options to increase the zoom/scaling of the displayed content, so does it really make sense to change the size-ratio between the elements here?

ArthurPxC:
Normally, there is a concept of device independent pixels. On Hi DPI displays a device independent pixel is recalculated to multiple physical pixels, to keep a control's size the same in relation to overall layout. I would not change a control's size in an isolated way. Instead, maybe there is a way to scale everything up on Hi DPI displays.

dbeuchler:
Our UIs are all DPI-aware, and the screen DPI can easily be set to 150% on 4K displays. I would also keep the control size as it is.

MathoMathiasCamara:
> Our UIs are all DPI-aware, and the screen DPI can easily be set to 150% on 4K displays. I would also keep the control size as it is.

That's a fair point. I will close the PR and let @dacky179 and Simon know about the conclusion.

PR #1420, opened 2026-08-11
Add an `IActivity.Abort` method allowing activities to define their abort results

1nf0rmagician:
Currently an activity only supports to `Complete` with a given result number (for which you need to know the possible results of the activity type and hence the activity type) and to `Fail`. While the fact that `ICell.ProcessAborting` exists is a detail of a different namespace and should not lead implementation directly, this function would allow for more generic and activity independent cell implementations. And since aborting an activity and having activity specific result creation has semantic meaning, this extension is proposed.

Example:
```csharp
void ICell.ProcessAborting(Activity affectedActivity)
{
    var result = affectedActivity.Abort();
    PublishActivityCompleted(_currentActivityStart.CreateResult());
}
```

### DIFF
--- a/src/Moryx.AbstractionLayer/Activities/Activity.cs
+++ b/src/Moryx.AbstractionLayer/Activities/Activity.cs
@@ -65,6 +65,12 @@ public abstract class Activity : IActivity
     }
 
     /// <inheritdoc />
+    public ActivityResult Abort()
+    {
+        return Result = CreateAbortionResult();
+    }
+
+    /// <inheritdoc />
     public ActivityResult Complete(long resultNumber)
     {
         return Result = CreateResult(resultNumber);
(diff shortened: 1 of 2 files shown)


1nf0rmagician:
@dbeuchler, @andreniggemann, @seveneleven let me know what you think about the idea 😊

andreniggemann:
I like the idea, but I have an alternative idea I would like to propose.
What if we combine this abort and the existing Fail Method to receive a parameter to allow reacting to different generally applicable results?

Let's say we have different standardized results similar to the DefaultActivtyResult that includes the current values from the DefaultActivityResult, Aborted and maybe others.
This would give us a generic way to set a value according to all the standard result types.

1nf0rmagician:
>Let's say we have different standardized results similar to the DefaultActivtyResult that includes the current values from the DefaultActivityResult, Aborted and maybe others.

Sounds interesting to me. Any ideas how you would start to implement something like that @andreniggemann? Currently the results are only added as Attributes (and I'm not a big fan of changing it, as it would result in loads of changes in application code). I could not spin up an idea how to get to your proposed solution from the top of my head 🤔

Issue #1368, opened 2026-07-30
Add Material Management Module


1nf0rmagician:
This is an overarching issue summarizing the work towards a Material Management module. The leading branch for this development is 
[feature/material-management-implementation-v001](https://github.com/PHOENIXCONTACT/MORYX-Framework/tree/feature/material-management-implementation-v001).

- [x] Add feature specifications for the module as well as integrations to other modules in the system  
#1205
- [x] Add architecture design description  
#1221
  - [ ]  Reiterate on architecture description, splitting it according to modularized architecture, adding a short how to use it tutorial
- [ ] Add Material Management Module with base implementation of a `MaterialContainer`  
#1275  
#1425
#1434
  - [x] User can request material in form of an `IMaterialContainer` in the system 
  - [ ] User can announce material in form of an (known) `IMaterialContainer` in the system 
  - [x] User can register an `IMaterialContainer` in the system 
  - [ ] User can register a previously requested/announced `IMaterialContainer` in the system 
  - [x] User can pre-advice an existing `IMaterialContainer` at the system 
  - [ ] User can de-register an `IMaterialContainer` from the system 
  - [ ] User can have background updates (announcement, registration) automatically matched to existing containers (request/announcement)
  - [ ] Users can audit past container state history using registered lineage events 
  - [x] User automatically receives events from material containers on the Web interface
- [ ] Add Order Integrator Module with base implementation for `OrderLinkedMaterialContainer`  
#1381
#1440
#1441
  - [x] User can link an operation to an `OrderLinkedMaterialContainer`
  - [ ] User can link an order to an `OrderLinkedMaterialContainer` without specifying a specific operation
  - [x] Application engineers can extend on the `OrderLinkedMaterialContainer` for cutsom use cases
  - [x] Users get accurate information about the operation status and the linking state on a container
  - [ ] Users can configure `LinkingHook` plugins as part of the linking protocol which are executed
  - [ ] Application engineers can process and evaluate the `ValidationContext` results from linking hooks in the material container or connected utilizing `ILinkingRequirement`s
  - [ ] Users can audit past linking history using registered lineage events 
 
Tbc.

PR #1434, opened 2026-08-14
Add deregistering of material containers

1nf0rmagician:
This PR is in partial fullfillment of #1368

### DIFF
--- a/src/Tests/Moryx.Material.IntegrationTests/MaterialManagementFacadeTests.cs
+++ b/src/Tests/Moryx.Material.IntegrationTests/MaterialManagementFacadeTests.cs
@@ -4,7 +4,6 @@
 using Moq;
 using Moryx.AbstractionLayer.Identity;
 using Moryx.AbstractionLayer.Resources;
-using Moryx.Material.Lineage;
 using Moryx.Material.States;
 using NUnit.Framework;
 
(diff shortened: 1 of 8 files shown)


src/Tests/Moryx.Material.IntegrationTests/MaterialManagementFacadeTests.cs
```diff
@@ -133,9 +132,38 @@ public void RegisterContainerAsync_WithContainer_RegistersContainer()
...
+        _resourceManagementMock.Setup(r => r.DeleteAsync(containerId, It.IsAny<CancellationToken>()))
+            .ReturnsAsync(true);
+        _resourceManagementMock.Raise(m => m.ResourceAdded += It.IsAny<EventHandler<IResource>>(), this, container);
+        var expectedFinalQuantity = container.Quantity;
+
+        // Act
+        await _materialManagement.DeregisterContainerAsync(containerId);
+
+        // Assert
+        var deregisteredStateInformation = container.StateInformation as DeregisteredStateInformation;
```
MathoMathiasCamara, line 149: Correct me if i am wrong but when you deregister the container the proxy becomes null in reality therefore trying to access it afterward will throw an error. In the Unit test i think that should be somehow taking into account.

MathoMathiasCamara (changes requested):

src/Tests/Moryx.Material.IntegrationTests/MaterialManagementFacadeTests.cs
1nf0rmagician: While this is true and the reason I did not find the error here, taking it into account would mean to include implementation details from another module into the tests for the material management. I don't want to do that here. That rather sounds like a system test than an integration test 🙈

PR #1441, opened 2026-08-17
Adds and manages operation source in the OrderReferencePool

1nf0rmagician:
This PR is stacked on #1440 

Having the information where an operation was created from is relevant for (notifications about) later state transitions of the container properly mapping linked containers to sources of their respective operations.


src/Moryx.Material.Endpoints/MaterialManagementController.cs
```diff
@@ -145,4 +173,25 @@ void Broadcast(IMaterialContainer container) => _containerStreamSubscribers.Valu
...
     #endregion
+    #endregion
+
+    #region Order References
+    #region GET
+
+    [HttpGet("integrations/orders/available")]
+    [ProducesResponseType(typeof(bool), StatusCodes.Status200OK)]
+    [Authorize(Policy = MaterialPermissions.CanRead)]
+    public ActionResult<bool> HasOrderIntegration() => Ok(_orderIntegration is null);
```
MathoMathiasCamara: i don't think this is not correct. It's supposed to be `not null`

src/Moryx.Material.Endpoints/MaterialManagementController.cs
```diff
@@ -67,7 +69,7 @@ public ActionResult<MaterialContainerTypeModel[]> GetTypes()
     [ProducesResponseType(typeof(string), StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [Authorize(Policy = MaterialPermissions.CanUpdate)]
-    public async Task<ActionResult<MaterialContainerModel>> PreAdviceAsync(PreAdvideModel preAdvice)
+    public async Task<ActionResult<MaterialContainerModel>> PreAdviceAsync(PreAdvideModel preAdvice, CancellationToken cancellationToken)
```
MathoMathiasCamara: the Model name `PreAdvideModel` has a typo

src/Moryx.Material.Management/Components/MaterialFlowHandler.cs
```diff
@@ -77,17 +87,33 @@ public Task RegisterContainerAsync(IMaterialContainer container, CancellationTok
...
         return Task.FromResult(container);
     }
 
-    public Task DeregisterContainerAsync(IMaterialContainer container, CancellationToken cancellationToken = default)
+    public async Task DeregisterContainerAsync(IMaterialContainer container, CancellationToken cancellationToken = default)
+    {
+        var deregistration = new DeregisteredStateInformation();
+        cancellationToken.ThrowIfCancellationRequested();
+        container.TransitionTo(deregistration);
+        var isDeleted = await ResourceManagement.DeleteAsync(container.Id, CancellationToken.None);
```
MathoMathiasCamara, line 103: why not pass the token here?

src/Moryx/Tools/Extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+using System.Reflection;
+
+namespace Moryx.Tools;
+
+/// <summary>
+/// Extensions for the <see cref="Enum"/>
+/// </summary>
+public static class EnumExtensions
```
MathoMathiasCamara, line 14: In my opinion there should the a different PR just for this extension, because if i am not wrong the `EntryConvert` already does something similar in the background for flag enums and enums. @dbeuchler what do you think ?

MathoMathiasCamara (changes requested):

src/Moryx/Tools/Extensions/EnumExtensions.cs
dbeuchler: yes sounds good and also some extensions are duplicated (see CustomAttributeProviderExtensions). Every MORYX-Framework changes not related to material management should be a separate PR.

src/Moryx.Material.Endpoints/MaterialManagementController.cs
1nf0rmagician: ups... :D

src/Moryx.Material.Management/Components/MaterialFlowHandler.cs
1nf0rmagician: Because I don't want the action to be cancelled after we deregistered the container. If we got beyond the throwifcancellation we go with it no matter what 😊

src/Moryx/Tools/Extensions/EnumExtensions.cs
1nf0rmagician: Yes absolutely, there are several commits in here which were fully seperated from the other changes. I just wanted to get them through for now, but will cherry pick all relevant changes to a seperate branch and make a pr to dev for them 😊

src/Moryx.Material.Management/Facade/MaterialManagementFacade.cs
```diff
@@ -264,19 +260,17 @@ public async Task DropMaterialAnnouncementAsync(Guid announcementId, Cancellatio
...
+            throw new KeyNotFoundException($"Could not find an {nameof(IMaterialContainer)} with {nameof(IPersistentObject.Id)} '{id}'");
 
         var finalQuantity = container.Quantity;
-        //await MaterialFlowHandler.TransitionAsync(container, new DeregisteredStateInformation(), cancellationToken);
+        await MaterialFlowHandler.DeregisterContainerAsync(container, cancellationToken);
         await LineageStorage.RecordAsync(new DeregisterLineageEvent
         {
-            ContainerId = container.Id,
+            ContainerId = id,
             FinalQuantity = finalQuantity
```
MathoMathiasCamara: there is an issue here: one is double ]the other is decimal. it's missing a casting

MathoMathiasCamara (changes requested):

Issue #1071, opened 2026-02-04
TimeSpan values not correctly transmitted from entry editor


1nf0rmagician:
**Describe the bug**
`TimeSpan` and propably also `DateTime` and `DateOnly` types are not correctly serialized in the `Entry` format

**To Reproduce**

1. Go to any resource class
2. Add an entry serialize property of type `TimeSpan`  
```csharp
[EntrySerialize]
public TimeSpan timeSpan { get; set; }
```
3. Start the application and create a resource of the type
4. Try changing and saving the value. 
5. While the formatting of the value is correct, the value is not saved.

**Affected components/modules**
Might either be a problem with the ngx-web-framework or with the serialization/formatting of times. Depending on this all web modules or all web modules using the entry editor are affected

**Possible workarounds**
Create a string property and manually parse a string to a timespan

dbeuchler:
This was never implemented. DateTime is serialized as string to view-only but to modify them, it's a missing feature. Its more a Feature instead of a bug.

PR #1443, opened 2026-08-18
Support Numerics-types in JSON serialization and ProductStorage + cleanup

dbeuchler:
Adds serialization support for `System.Numerics` types (`Vector2`, `Vector3`, `Quaternion`) and date/time types (`DateOnly`, `TimeOnly`) across JSON serialization and the product storage column mappers.

The core problem is that `Vector3`, `Vector2` and `Quaternion` have public fields instead of properties, so Newtonsoft.Json silently serializes them as `{}`. See also https://github.com/JamesNK/Newtonsoft.Json/issues/2521. Similarly, `DateOnly` and `TimeOnly` are not natively supported by Newtonsoft.

Custom `JsonConverter` implementations for `Vector2`, `Vector3`, `Quaternion`, `DateOnly` and `TimeOnly` registered as singletons in `JsonSettings.Minimal` and `JsonSettings.Readable`. `DateTimeOffset` and `TimeSpan` already work natively with Newtonsoft and dont need converters.

**Product Storage**

- `TextColumnMapper` now handles non-primitive structs (in addition to classes/interfaces), serializing them as JSON into text columns. This covers types like `Vector3`, `Quaternion` and `DateTimeOffset`
- `TextStrategyConfigurationAttribute` explicitly matches non-primitive structs with a score
- `IntegerColumnMapper` supports `DateOnly` (stored as `DayNumber`) and `TimeOnly` (stored as `Ticks`)

**Test product cleanup**

While adding test types for the new features, the existing test products were cleaned up:

- Renamed `TextColumnMapperTestProductType` to `CompositeProductType` — the old name did not reflect what it actually tests (complex reference type serialization).
- Removed unused primitive properties that were set in tests but never asserted.
- Removed the redundant `ProdDataAdded` / `TestProdData` since it tested the same thing as `ComplexData1`.
- Renamed `GenericJsonTestProductType` to `JsonProductType` for consistency.
- Added `VectorProductType` and `DatedProductType` as dedicated test types for the new features.
- Merged new TestProducts-project added in #1364 with Samples. This was wrong interpret by me (thought Samples were released).

**Docs and tests**

- Updated docs with the newly supported types for each column mapper.
- Added tests for the JsonConverters

**Benefits**

Sample: Use vector directly in resources/products: 

A workaround was to create a DTO and ignore the original property:

```csharp
[DataContract]
public class VectorDto
{
    [DataMember]
    public float X { get; set; }

    [DataMember]
    public float Y { get; set; }

    [DataMember]
    public float Z { get; set; }

    public Vector3 ToVector3()
    {
        return new Vector3(X, Y, Z);
    }

    public void FromVector3(Vector3 vector3)
    {
        X = vector3.X;
        Y = vector3.Y;
        Z = vector3.Z;
    }
}

  // ...
  [DataMember, EntrySerialize]
  [EditorBrowsable(EditorBrowsableState.Never)]
  public VectorDto SomeFooDto { get; set; }

  [EntrySerialize(EntrySerializeMode.Never)]
  [JsonIgnore]
  public Vector3 SomeFoo
  {
    get => SomeFooDto?.ToVector3() ?? Vector3.Zero;
    set
      {
          SomeFooDto??= new VectorDto();
          SomeFooDto.FromVector3(value);
      }
  }
  // ...
```

In resources (full json serialization), it can now be replaced by: 

```csharp
  [DataMember, EntrySerialize]
  public Vector3 SomeFoo { get; set; }
```

or in products (mapped by text-column-mapper) by 

```csharp
  public Vector3 SomeFoo { get; set; }
```

**Limitations**

~~Vector3/Vector2/Quaternion/TimeOnly/DateOnly (#1071) )are classified as value types and treated as strings in EntryConvert. ToString() produces <1.5, 2.5, 3.5> but there's no corresponding Parse to reconstruct them. The sample from the workaround provides a navigable entry with sub-entries on the UI. If not too complex, a PR will follow later.~~

--> #1445


src/Moryx.Products.Samples/ProductTypes/DatedProductType.cs
```diff
@@ -0,0 +1,23 @@
...
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.Products.Samples;
+
+/// <summary>
+/// Test product type for DateOnly, TimeOnly and DateTimeOffset serialization.
+/// </summary>
+public class DatedProductType : ProductType
+{
+    public DateOnly ValidFrom { get; set; }
```
1nf0rmagician: While there is an ADR to use `DateTimeOffset` instead of `DateTime`, should `DateTime` be part of this test type and included in tests explicitely?

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
```diff
@@ -1212,4 +1231,100 @@ public async Task LoadTypesQueryShouldReturnBaseInformationByDefault()
...
+
+        // Assert
+        Assert.That(loaded.ProductionTime, Is.EqualTo(new TimeOnly(14, 30, 45)));
+    }
+
+    [Test]
+    public async Task SaveAndLoadTypeWithDateTimeOffset()
+    {
+        // Arrange
+        var createdAt = new DateTimeOffset(2026, 8, 18, 14, 30, 0, TimeSpan.FromHours(2));
```
1nf0rmagician, line 1315: In all other tests the 'expected' value is created twice, would be neat if all would define it in the arrange and use it in the assert like in this test

1nf0rmagician (commented):
Aside from the minor comment in the tests, I saw that for the Numeric types the dotnet namespace also contains the `Plane` and `Vector4` types, which directly fall in line with the changes you have submitted here. Would you just add the two types in this PR to have full coverage of the corresponding dotnet namespace? 😬

src/Tests/Moryx.Products.IntegrationTests/ProductStorageTests.cs
dbeuchler: could be done in a separate PR?

dbeuchler:
Plane and Vector4 could be added in a separate PR. I also don't added MatrixXxX

1nf0rmagician (approved):

PR #1444, opened 2026-08-18
Updated error responses of the ResourceModificationController to be RFC 9457 compliant

Liviafranke:
# Changes

## Standardized Error Responses
All error responses returned by `ResourceModificationController` now use the RFC 9457 format with the following fields:
- `type`
- `title`
- `status`
- `detail`
- `instance`
 
Field descriptions:
- `type` references the HTTP status code definition in RFC 9110.
- `title` provides a localized summary of the error.
- `status` contains the HTTP status code returned by the API.
- `detail` provides a localized and more detailed description of the error.
- `instance` contains the path of the request that triggered the error (`HttpContext.Request.Path`).

## Controller Error Handling
Replaced `MoryxExceptionResponse` with `ProblemDetails`. 
The controller only returns expected and explicity handled errors. For these scenarios, the standard `ProblemDetails` class is fully sufficient because no additional exception information is required. Therefore, `MoryxExceptionResponse` did not provide any additional value in this context.  
`MoryxExceptionResponse` is still used by the global `MoryxExceptionFilter`, which handles unexpected exceptions and includes the exception field in the response body to support debugging.

## Refactoring of MoryxExceptionResponse
`MoryxExceptionResponse` now inherits from `ProblemDetails`.
 
As a result:
- The response model is aligned with RFC 9457.
- The custom error title property has been removed, since `ProblemDetails` already provides the standardized `title` field.
- Existing exception-specific information can still be included where required.

## Localization
Added localization for the `title` and `detail` fields in English, German, Italian, Polish and Chinese.


src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -109,15 +109,23 @@ public ActionResult<ResourceModel[]> GetResources([FromQuery] ResourceQuery quer
...
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [Authorize(Policy = ResourcePermissions.CanViewDetails)]
     public ActionResult<ResourceModel> GetDetails(long id)
     {
         var converter = new ResourceToModelConverter(_resourceTypeTree, _serialization);
         var resourceModel = _resourceManagement.ReadUnsafe(id, r => converter.GetDetails(r));
         if (resourceModel is null)
-            return NotFound(new MoryxExceptionResponse { Title = string.Format(Strings.ResourceNotFoundException_ById_Message, id) });
+
+            return NotFound(new ProblemDetails
```
1nf0rmagician: We should be able to shorten this in an elegenat fashion. Something like 
```csharp
return NotFound().With(Title, Details, context)
```
Maybe @seveneleven has an opinion on this, designing fluent APIs is something of a habit of his 😁

If we go with the example above, you'd need to create extensions methods on, in this case, the NotFoundResult type, which returns a NotFoundObjectResult with the problemdetails object inside that you also used here. The result would be an easier to read return statement with less duplicated property assignements. 

And then you would repeat that for `BadRequest` and any other standard error response code we are currenlty using 😊

**Edit:** For the resource not found response I'm also very sure, that you can further decrease code duplication if you define an extension method within this project which already has the other properties set as well.

**Edit 2:** I just saw there is a `Problem(...)` virtual method on the controller base class which already gets the context set. So my new suggestion for the fluent API is
```csharp
// Placed in Moryx.AspNetCore project
return Problem().NotFound(Title, Details)
```
And within this project
```csharp
return Problem().NotResourceFound(id)
```

src/Moryx.AbstractionLayer.Resources.Endpoints/ResourceModificationController.cs
```diff
@@ -139,13 +147,20 @@ public ActionResult<ResourceModel> GetDetails(long id)
...
     public async Task<ActionResult<Entry>> InvokeMethod(long id, string method, Entry parameters)
     {
         if (!_resourceManagement.GetResourcesUnsafe<IResource>(r => r.Id == id).Any())
-            return NotFound(new MoryxExceptionResponse { Title = string.Format(Strings.ResourceNotFoundException_ById_Message, id) });
+            return NotFound(new ProblemDetails
+            {
+                Type = "https://www.rfc-editor.org/rfc/rfc9110.html#name-404-not-found",
+                Title = Strings.ResourceModificationController_ResourceNotFound_Title,
+                Status = StatusCodes.Status404NotFound,
+                Detail = string.Format(CultureInfo.CurrentCulture, Strings.ResourceModificationController_ResourceNotFoundException_ById_Message, id),
```
1nf0rmagician:
```suggestion
                Detail = string.Format(CultureInfo.CurrentCulture, Strings.ResourceModificationController_ResourceNotFound_ById_Message, id),
```

src/Moryx.AspNetCore/MoryxExceptionResponse.cs
```diff
@@ -1,11 +1,10 @@
...
+using Microsoft.AspNetCore.Mvc;
 
 namespace Moryx.AspNetCore;
 
-public class MoryxExceptionResponse
+public class MoryxExceptionResponse : ProblemDetails
 {
-    public string Title { get; set; }
-
     public string Exception { get; set; }
```
1nf0rmagician, line 9: As the RFC specifically states [Problem details are not a debugging tool for the underlying implementation](https://www.rfc-editor.org/rfc/rfc9457.html#section-4) we should remove this type completely and should not return exception messages or stack traces from any endpoint. 

Instead, we should log the exceptions in the endpoints together with a trace Id, which can be added to the problem details in the extensions. I think this is the place were we should start with providing an extension method to be used in the `Program.cs` as a central solution. Following is a minimal AI snippet for that. However, it does not include logging the exceptions in the controller or appending trace IDs only where there is actually a problem details response. So it only swerves as a starting point.

```csharp
using System.Diagnostics;

builder.Services.AddProblemDetails(options =>
{
    options.CustomizeProblemDetails = context =>
    {
        context.ProblemDetails.Extensions["traceId"] =
            Activity.Current?.TraceId.ToString()
            ?? context.HttpContext.TraceIdentifier;
    };
});
```

PR #1445, opened 2026-08-18
Support DateOnly, TimeOnly, DateTime, TimeSpan, DateTimeOffset and some structs in EntryConvert

### LINKED ISSUE #1071: TimeSpan values not correctly transmitted from entry editor

dbeuchler:
based on #1443

- Added DateOnly, TimeOnly, DateTime and TimeSpan (requested by #1071) to EntryUnitType for official support
- Support serialization of structs with custom serializers. Implemented: Vector2, Vector3, Quaternion
- Add support for DateTimeOffset treated as DateTime and mapped to UTC

Screenshot 2026-08-18 at 17 32 55

Screenshot 2026-08-18 at 17 33 24

close #1071

### DIFF
--- a/docs/articles/framework/Serialization/entry-convert.md
+++ b/docs/articles/framework/Serialization/entry-convert.md
@@ -17,7 +17,7 @@ The `EntryConvert` API can convert objects and types as long as they comply with
 
 * Properties not fields: All attributes of a type must be defined as properties, not public fields. Therefor `public int Foo { get; set; }` instead of `public int Foo;`
 * Public parameter-less constructor: All types within the class hierarchy need to offer a public constructor without parameters. In Generics this would be defined as `new()` or in code `public Foo() { }`. For the root object `EntryConvert` can extract Constructors as `MethodEntry`, which can be exchanged with a client and used to create instances.
-* Primitives, classes or supported structs: The reflection approach used to deserialize the entry tree to objects requires reference access. Otherwise the modifications will only take part on a copy. Therefor properties need to be either of a primitive type like int, string, enum, a class, or a supported struct (`Vector2`, `Vector3`, `Quaternion`). Supported structs are automatically decomposed into editable sub-entries.
+* Primitives, classes or supported structs: The reflection approach used to deserialize the entry tree to objects requires reference access. Otherwise the modifications will only take part on a copy. Therefor properties need to be either of a primitive type like int, string, enum, a class, or a supported struct (`Vector2`, `Vector3`, `Vector4`, `Quaternion`, `Plane`). Supported structs are automatically decomposed into editable sub-entries.
 * Dictionaries of `<Primitive, Class`: Dictionaries are only supported if the key is a primitive type like `int` or `string` and the value is a class.
 
 ## Serialize Objects
@@ -143,7 +143,7 @@ public void Deserialize(Entry entry, FileStreamDummy dummy)
 
 ## Serialize Structs
 
-`EntryConvert` supports decomposed serialization for the following `System.Numerics` struct types: `Vector2`, `Vector3` and `Quaternion`. Instead of displaying unparseable strings like `<1.5, 2.5, 3.5>`, these structs are encoded as `Struct` entries with editable sub-entries for each component (X, Y, Z, W). When you serialize a class containing one of these types, the resulting entry will have sub-entries for each component of the struct.
+`EntryConvert` supports decomposed serialization for the following `System.Numerics` struct types: `Vector2`, `Vector3`, `Vector4`, `Quaternion` and `Plane`. Instead of displaying unparseable strings like `<1.5, 2.5, 3.5>`, these structs are encoded as `Struct` entries with editable sub-entries for each component (X, Y, Z, W). When you serialize a class containing one of these types, the resulting entry will have sub-entries for each component of the struct.
 
 ````cs
 public class RobotPosition
--- a/docs/articles/framework/Serialization/entry-convert.md
+++ b/docs/articles/framework/Serialization/entry-convert.md
@@ -11,6 +11,33 @@ Because the `EntryConvert`-API can be confusing on first sight, we will split it
 * **Deserialize:** The encoded entry tree can be converted into objects after modification by the client with the overloads of `CreateInstance` and `UpdateInstance`
 * **Customization:** The behavior of both encoding and decoding can be customized to specific needs by providing a strategy implementing `ICustomSerialization`
 
+## Supported Types
+
+| .NET Type | EntryValueType | Format | Limitations |
+|-----------|---------------|--------|-------------|
+| `byte` | Byte | — | — |
+| `bool` | Boolean | — | — |
+| `short`, `ushort` | Int16, UInt16 | — | — |
+| `int`, `uint` | Int32, UInt32 | — | — |
+| `long`, `ulong` | Int64, UInt64 | — | — |
+| `float` | Single | Culture-aware with invariant fallback | — |
+| `double` | Double | Culture-aware with invariant fallback | — |
+| `decimal` | Double | Culture-aware with invariant fallback | Mapped to `Double; may lose precision beyond ~15 significant digits |
+| `string` | String | — | — |
+| `enum` | Enum | String name | — |
+| `DateTime` | DateTime | ISO 8601 round-trip (`"O"`) | Converted to UTC; original `DateTimeKind` is lost |
+| `DateTimeOffset` | DateTime | ISO 8601 round-trip (`"O"`) | Converted to UTC; original timezone offset is lost |
+| `DateOnly` | Date | ISO 8601 round-trip (`"O"`) | — |
+| `TimeOnly` | Time | ISO 8601 round-trip (`"O"`) | — |
+| `TimeSpan` | TimeSpan | Constant format (`"c"`) | — |
+| `Stream` | Stream | Base64 encoded | Limited by available memory |
+| `Vector2`, `Vector3`, `Vector4` | Struct | Decomposed into sub-entries (X, Y, Z, W) | — |
+| `Quaternion` | Struct | Decomposed into sub-entries (X, Y, Z, W) | — |
+| `Plane` | Struct | Decomposed into sub-entries (X, Y, Z, D) | — |
+| Classes | Class | Recursive sub-entries | Requires public parameter-less constructor |
+| Collections | Collection | Recursive sub-entries | — |
+| Dictionaries | Collection | Recursive sub-entries | Key must be a type supported by `ToObject` |
+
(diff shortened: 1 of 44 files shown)


docs/articles/framework/Serialization/entry-convert.md
```diff
@@ -141,6 +141,37 @@ public void Deserialize(Entry entry, FileStreamDummy dummy)
 }
 ````
 
+## Serialize Structs
+
+`EntryConvert` supports decomposed serialization for the following `System.Numerics` struct types: `Vector2`, `Vector3` and `Quaternion`. Instead of displaying unparseable strings like `<1.5, 2.5, 3.5>`, these structs are encoded as `Struct` entries with editable sub-entries for each component (X, Y, Z, W). The UI renders them as expandable entries with individual number fields — no UI changes required.
```
1nf0rmagician: >The UI renders them as expandable entries with individual number fields — no UI changes required.

Stating how "the UI" will render it here does not seem to me like a lasting information and since I couldn't find information about the ui in this document otherwise I would remove it.

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
```diff
@@ -7,10 +7,47 @@
...
+    /// Converts given value typed instance to a string with the given <see cref="IFormatProvider"/>
+    /// </summary>
+    /// <param name="value">Value to convert</param>
+    /// <param name="formatProvider">Format provider used to convert the value to string</param>
+    /// <returns>Formatted string representation</returns>
+    internal static string ConvertToString(object value, IFormatProvider formatProvider)
+    {
+        return value switch
+        {
+            DateTime dt => dt.ToUniversalTime().ToString("O", formatProvider),
```
1nf0rmagician, line 45: I am wondering at this point, in #1443 I did not comment on the fact that in the date/time related tests `DateTime` was not covered as I expected it to be covered by default by NewtonsoftJson serialization. Here we have a specific setting for it now. Just want to make sure I shouldn't have added the comment in the last PR 🙈

Also did you intentionally not list DateTimeOffset here?

src/Moryx/Serialization/EntryConvert/EntryValueType.cs
```diff
@@ -71,5 +71,25 @@ public enum EntryValueType
...
\ No newline at end of file
+    Stream,
+    /// <summary>
+    /// Decomposed struct with sub-entries
+    /// </summary>
+    Struct,
+    /// <summary>
+    /// Value type DateTime
+    /// </summary>
+    DateTime,
```
1nf0rmagician, line 82: Should DateTimeOffset have a dedicated value here as well?

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
```diff
@@ -152,6 +209,18 @@ public static object ToObject(Type type, string value, IFormatProvider formatPro
...
+        {
+            result = DateOnly.Parse(value, formatProvider);
+        }
+        else if (type == typeof(TimeOnly))
+        {
+            result = TimeOnly.Parse(value, formatProvider);
+        }
+        else if (type == typeof(TimeSpan))
+        {
+            result = TimeSpan.Parse(value, formatProvider);
```
1nf0rmagician, line 234: DateTime and DateTimeOffsett missing here on purpose?

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
```diff
@@ -67,6 +104,26 @@ public static EntryValueType TransformType(Type propertyType)
...
+        }
+        else if (propertyType == typeof(TimeOnly))
+        {
+            valueType = EntryValueType.Time;
+        }
+        else if (propertyType == typeof(TimeSpan))
+        {
+            valueType = EntryValueType.TimeSpan;
+        }
+        else if (propertyType == typeof(DateTime))
```
1nf0rmagician: If DateTimeOffset is added don't miss here

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
```diff
@@ -252,6 +321,22 @@ public static object ToObject(EntryValueType type, string value, IFormatProvider
...
+
+            case EntryValueType.Time:
+                result = TimeOnly.Parse(value, formatProvider);
+                break;
+
+            case EntryValueType.TimeSpan:
+                result = TimeSpan.Parse(value, formatProvider);
+                break;
+
+            case EntryValueType.DateTime:
```
1nf0rmagician, line 349: If DateTimeOffset is added don't miss here

src/Tests/Moryx.Tests/Serialization/EntryConvertSerializationTests.cs
```diff
@@ -128,30 +129,145 @@ public void ToObject_ByEntryValueType_AlsoUsesParseWithFallback()
...
-        Assert.Multiple(() =>
-        {
-            Assert.That(plainParameterValidation.IsRequired, Is.True ,"Plain parameter should be reuired");
-            Assert.That(requiredParameterValidation.IsRequired, Is.True , "Required parameter should be required");
-            Assert.That(nullableValidation.IsRequired, Is.False, "Nullable parameter should not be required");
-            Assert.That(defaultValueValidation.IsRequired, Is.False, "Default value shuold not be required");
-        });
+
+    [Test]
+    public void EncodesVector3AsStructWithSubEntries()
```
1nf0rmagician: Optionally you could put all the structs into a testcasesource providing the obj, nameof, identifiers and number of subentries

**Edit:** Same test case source could cover that default entries are created for each type as tested in `CreatesDefaultSubEntriesForVector3OnEncodeClass`

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
dbeuchler: Yes we have this specific setting because we want a ISO 8601 date format based on UTC (see the other Datetime conversion parts in the partial). DateTimeOffset is not part of this PR, its part of the tests because its generally supported by Newtonsoft.Json

src/Moryx/Serialization/EntryConvert/EntryValueType.cs
dbeuchler: DateTimeOffset is not part of this PR

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
dbeuchler: DateTimeOffset is not part of this PR, Datetime is in Line 208

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
dbeuchler: DateTimeOffset is not part of this PR

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
dbeuchler: DateTimeOffset is not part of this PR

dbeuchler:
@1nf0rmagician I will check the TestCaseSource and also add Plane and Vector4

src/Moryx/Serialization/EntryConvert/EntryConvert.TypeConversion.cs
1nf0rmagician: Understand, I know its asking more on a PR that is already adding nice functionality, but since we have [ADR005](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/docs/adr/005-datetime-handling.md) it would be so nice if you could also add DateTimeOffset 😬
But I'll approve without, as it is not an issue of this PR that it is missing from the framework in general 😂

dbeuchler:
@1nf0rmagician I added also Plane and Vector4 to EntryConvert and converted the test-cases to `TestCase` and `TestCaseSource`.

It was not the plan of this PR to support DateTimeOffset because this is a bit tricky and from my point of view also not user-friendly (what to show to the user? Date field with picker, Time input and another offset input?

What we could do: 
DateTime values are already converted to UTC during EntryConvert serialization, which means the original DateTimeKind (Local/Unspecified) is lost and cannot be restored on deserialization.

I hadn't realized it was implemented that way. We should reconsider this for MORYX 12. I would actually expect no values ​​to change during a round-trip. We might need to tweak EntryValue a bit more, but in my opinion it shouldn't stay like this.

The same could be applied to DateTimeOffset, where the original timezone offset could be discarded when mapping to UTC. I will implement it like that and also add a table for supports and limitations.

docs/articles/framework/Serialization/entry-convert.md
```diff
@@ -11,14 +11,41 @@ Because the `EntryConvert`-API can be confusing on first sight, we will split it
 * **Deserialize:** The encoded entry tree can be converted into objects after modification by the client with the overloads of `CreateInstance` and `UpdateInstance`
 * **Customization:** The behavior of both encoding and decoding can be customized to specific needs by providing a strategy implementing `ICustomSerialization`
 
+## Supported Types
```
1nf0rmagician, line 14: This is an amazing addition!

1nf0rmagician (commented):
Changes look good to me now, when the ngx-web-framework update is done we could merge this 😊 Do you want to squash some of the fix commits in the middle, when you add the version update for the web framework?

dbeuchler:
> Changes look good to me now, when the ngx-web-framework update is done we could merge this 😊 Do you want to squash some of the fix commits in the middle, when you add the version update for the web framework?

Thansk! I will do that!

dbeuchler:
Done! There are still failing tests, I will fix them this evening

1nf0rmagician (approved):

Issue #469, opened 2024-10-09
Support System.Nullable in ProductStorage


dbeuchler:
**Is your feature request related to a problem? Please describe.**
My product instances often has properties which are filled during production. For example DateTime, double, or integers.
At the moment I must set them to a default-value or MinValue to be stored. `double.NaN` is working in PostgreSQL but not in SQLite. 

The knowledge about the underlying ProductStorage is essential to store the Types and Instances. 

**Describe the solution you'd like**
Supporting System.Nullable in the ProductStorage would help to minimize unexpected problems.

**TODO:** 
- Change Entities to Nullable-Columns by default
- Support System.Nullable in ProductStorage strategies.

PR #1446, opened 2026-08-19
Support IEEE754 special values in FloatColumnMapper

dbeuchler:
`double.NaN`, `double.PositiveInfinity` and `double.NegativeInfinity` are valid IEEE 754 values that can appear on product properties mapped to float columns. PostgreSQL accepts these in `float8` columns, but SQLite throws an exception when writing them to `REAL` columns.

The `FloatColumnMapper` now encodes NaN/+-Infinity as sentinel values before writing to the database and decodes them back when reading. 

The sentinels are created by stepping from `double.MinValue` by bit increment. each step produces the next representable double. Collision with real data is effectively impossible.

**PostgreSQL backward compatibility**

Existing PostgreSQL databases may already contain raw NaN/Infinity values. These are read correctly. On the next save, `HasChanged` detects the mismatch between the raw value in the column and the sentinel-encoded value, triggering a version update that writes the sentinel. No migration needed.

---

Add to docs later (to avoid conflicts): 

> IEEE 754 special values (`NaN`, `PositiveInfinity`, `NegativeInfinity`) are not natively supported by all databases — PostgreSQL accepts them in `float8` columns, but SQLite throws on `REAL` columns. To ensure cross-database compatibility, the mapper encodes these as sentinel values (distinct doubles near `double.MinValue`) on write and decodes them back on read. Existing PostgreSQL databases with raw NaN/Infinity values are handled automatically: the values are read correctly and converted to sentinels on the next save.

---

link #469 because its related

### DIFF
--- a/src/Tests/Moryx.Products.IntegrationTests/FloatColumnMapperTests.cs
+++ b/src/Tests/Moryx.Products.IntegrationTests/FloatColumnMapperTests.cs
@@ -1,0 +1,224 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moq;
+using Moryx.Products.Management;
+using Moryx.Products.Management.Model;
+using NUnit.Framework;
+
+namespace Moryx.Products.IntegrationTests;
+
+[TestFixture]
+public class FloatColumnMapperTests
+{
+    private FloatColumnMapper _mapper;
+
+    [SetUp]
+    public void SetUp()
+    {
+        _mapper = new FloatColumnMapper(typeof(TestObject));
+        _mapper.Initialize(new PropertyMapperConfig
+        {
+            PropertyName = nameof(TestObject.Value),
+            Column = nameof(IGenericColumns.Float1),
+            PluginName = nameof(FloatColumnMapper)
+        });
+    }
+
+    [TestCase(0.0)]
+    [TestCase(1.0)]
+    [TestCase(-1.0)]
+    [TestCase(42.5)]
+    [TestCase(double.MinValue)]
+    [TestCase(double.MaxValue)]
+    [TestCase(double.Epsilon)]
+    [Description("Regular double values are stored in the column and restored exactly after a write-read cycle.")]
+    public void RoundTripsRegularValues(double value)
+    {
+        // Arrange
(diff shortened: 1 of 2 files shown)


src/Tests/Moryx.Products.IntegrationTests/FloatColumnMapperTests.cs
```diff
@@ -0,0 +1,224 @@
...
+        // Arrange
+        var source = new TestObject(double.NegativeInfinity);
+        var columns = CreateColumns();
+
+        // Act
+        _mapper.WriteValue(source, columns);
+
+        // Assert
+        Assert.That(double.IsInfinity(columns.Float1), Is.False);
+    }
```
1nf0rmagician: The asserts in these tests are already covered by the tests above. Instead of keeping these tests I suggest, the helpful description why this test exists is placed in the assert of the tests above for simplicity.

You could argue that each test should have its own assert, which would be fine to, but then the other assert should be removed

src/Tests/Moryx.Products.IntegrationTests/FloatColumnMapperTests.cs
```diff
@@ -0,0 +1,224 @@
...
+        // which gets encoded as sentine
+        var columns = CreateColumns();
+        columns.Float1 = double.NaN;
+        var source = new TestObject(double.NaN);
+
+        // Act
+        var changed = _mapper.HasChanged(columns, source);
+
+        // Assert
+        Assert.That(double.IsNaN(columns.Float1), Is.True);
```
1nf0rmagician: This assert checks a direct assignement from the arrange, thats a bit weird 😅

1nf0rmagician: Same in the test above

1nf0rmagician (changes requested):

src/Tests/Moryx.Products.IntegrationTests/FloatColumnMapperTests.cs
dbeuchler: mhmm understand.. i was writing it to have it explicitly tested but they are coered by the roundtrips, thats correct.

1nf0rmagician (approved):

PR #1447, opened 2026-08-19
Use PropertyAccessor in bindings ReflectionResolver

### LINKED ISSUE #11: Integrate ProperyAccessor in Binding Resolver
The reflection tools offer optimized property access. That could be combined with the platforms binding resolver chain.

dbeuchler:
The ReflectionResolver now uses PropertyAccessor with compiled delegates instead of PropertyInfo.GetValue()/SetValue() for property access, replacing repeated reflection overhead with a direct method invocation. 

The accessor is cached per source type to handle polymorphic objects while still providing fast access on subsequent calls.

No additional tests, the Binding-Tests already check the behavior.

close #11

### DIFF
--- a/src/Moryx/Bindings/ReflectionResolver.cs
+++ b/src/Moryx/Bindings/ReflectionResolver.cs
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0
 
 using System.Reflection;
+using Moryx.Tools;
 
 namespace Moryx.Bindings;
 
 /// <summary>
-/// Resolve binding using reflection
+/// Resolve binding using optimized property access
 /// </summary>
 public class ReflectionResolver : BindingResolverBase
 {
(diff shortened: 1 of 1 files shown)


1nf0rmagician (approved):
While the changes look fine to me, I am wondering, how often is a ReflectionResolver going to be reused usually? The only use case I found, while reading up on the whole topic, is when within an activity we have multiple instruction trying to resolve information from e.g. the product type on it, right? I'm just asking to better understand the impact of this change 😊

dbeuchler:
> While the changes look fine to me, I am wondering, how often is a ReflectionResolver going to be reused usually? The only use case I found, while reading up on the whole topic, is when within an activity we have multiple instruction trying to resolve information from e.g. the product type on it, right? I'm just asking to better understand the impact of this change 😊

resolvers instances are created once and then typically stored. So all tasks with parameters of VisualInstructionParameters. It's used in MeasurementBindings or other places. 

The performance gain is is relatively small, but its mainly about using our own PropertyAccessor instead of raw reflection to stay consistent with the rest of the framework and use our own features.

PR #1448, opened 2026-08-20
Replace resource IL proxy generation with Castle.DynamicProxy

### LINKED ISSUE #768: ResourceProxy does not support explicitly implemented interfaces with events
**Describe the bug**
When explicitly implementing an interface a ResourceProxy cannot be created.
I worked around this limitation implicitly implementing the interface.

@andreniggemann thanks for the report

dbeuchler:
The resource proxy used hand-written IL emit code to generate proxy types at runtime. Probs to @Toxantron 👏 for this masterpiece (I remember the Saturdays debugging IL stuff 💻). This has been running for several years without any problems, but it has some limitations:

- The IL code was hard to read, debug, and maintain. Nobody wants to step through `ILGenerator.Emit` calls to understand why something throws.
- Explicit interface implementations were broken when events were involved (#768).
- Generic methods were not supported (3 years old PR #343)
- Every new feature requires writing more IL by hand.

**Changes:**

- Replaced `ResourceProxyBuilder` (IL emit) with a Castle.DynamicProxy-based implementation
- `ResourceProxy` serves as mixin providing `IResource` base members
- `ResourceInterceptor` handles all call forwarding, events, and resource reference conversion
- Proxy type caching is handled by Castle internally — removed the manual cache
- Explicit interface events now work correctly
- Generic methods on target are now fully supported
- Throw exception on custom delegate types for events
- Enhanced docs

All existing proxy tests pass unchanged. Added new tests for explicit interface implementations, detach behavior and other missing features.

**Performance:**

- The old IL proxies compiled down to direct method calls, so per-call overhead was essentially zero. Castle uses `IInterceptor` which adds indirection through reflection on each call.
- Proxy type generation itself is not an issue — it happens once per resource type and Castle caches the types internally.
- With the MORYX-Benchmarker I cannot identify any performance impact.
  - IL proxies — average: 176 activities/second; Castle proxies — average: 197 activities/second; so the proxy overhead is negligible here
- With 6.5M ops/sec (on an M4 MacbookPro) the Castle proxy is still bloody fast — a single property access takes ~150 nanoseconds, which is irrelevant for any real resource interaction that involves I/O, network, or database.
- On direct call benchmark (loop and measure) I can identify that the IL proxy was around 9x faster
- Possible optimization: Replace MethodInfo.Invoke with compiled delegates (like in PropertyAccessor), but this could be a later improvement. I think we can save 5%-10%, therefore is more mini-optimization.

**The maintainability gain and new features far outweighs the nanoseconds lost per invocation.**

close #768 #343

@andreniggemann FYI (or sec review)


src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+    private void HandleEventAdd(string eventName, Delegate handler)
+    {
+        if (handler is null)
+        {
+            return;
+        }
+
+        lock (_eventDelegates)
+        {
+            _eventDelegates.TryGetValue(eventName, out var existing);
```
1nf0rmagician: I'm not 100% sure yet whether the `ResourceInterceptor` is a per proxy class. If not, couldn't the delegates colide here, if we have two different types with event that have the same name?

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+    public static IResourceProxy Build(Resource target, IReadOnlyList<Type> interfaces, IResourceTypeController typeController)
+    {
+        var mixin = new ResourceProxy(target);
+
+        // Build proxy options with the mixin providing IResource + IResourceProxy
+        var options = new ProxyGenerationOptions();
+        options.AddMixinInstance(mixin);
+        options.BaseTypeForInterfaceProxy = typeof(ResourceProxyBase);
+
+        // Separate primary and additional interfaces
```
1nf0rmagician, line 29: Why, what is the purpose of the primary interface?

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+        ResourceProxy mixin, ResourceInterceptor interceptor)
+    {
+        var targetType = target.GetType();
+        var processedEvents = new HashSet<string>();
+
+        foreach (var iface in interfaces)
+        {
+            foreach (var eventInfo in iface.GetEvents(BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public))
+            {
+                // Skip if already processed (same event on multiple interfaces)
```
1nf0rmagician: Does this skipping work for two explicit interface implementations which share a name?

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+                if (!processedEvents.Add(eventInfo.Name))
+                {
+                    continue;
+                }
+
+                var targetEvent = targetType.GetEvent(eventInfo.Name,
```
1nf0rmagician: Doesn't get this the same object as eventInfo itself?

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        // EventHandler<T>
+        if (handlerType.IsGenericType && handlerType.GetGenericTypeDefinition() == typeof(EventHandler<>))
+        {
+            var argType = handlerType.GetGenericArguments()[0];
+            return CreateGenericForwarder(argType, eventName, interceptor);
+        }
+
+        return null;
```
1nf0rmagician: Is there any expected situation in which this return null happens?

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+
+                var handler = CreateEventForwarder(targetEvent, eventInfo.Name, interceptor);
+                if (handler is null)
+                {
+                    continue;
+                }
+
+                targetEvent.GetAddMethod(nonPublic: true)!.Invoke(target, [handler]);
+                mixin.RegisterEventHandler(eventInfo.Name, handler);
```
1nf0rmagician: A thought to discuss (but not include in this PR): Should we decouple these events similar to what processes and orders do? Currenlty the stacktrace from a resource changed event for example goes from the resource into the database and into the SSE service, which leaves a lot of possibilities to block processing in resources. On the other side it gives better tracability and I don't know how much trouble we get into when the ReadyToWork events to the ProcessEngine would be decoupled 🤔

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+                {
+                    _methodCache[interfaceMethod] = map.TargetMethods[i];
+                    return map.TargetMethods[i];
+                }
+            }
+        }
+
+        // Fallback: find by name and parameter types
+        var paramTypes = interfaceMethod.GetParameters().Select(p => p.ParameterType).ToArray();
+        var resolved = targetType.GetMethod(interfaceMethod.Name,
```
1nf0rmagician, line 146: Why the `NonPublic` BindingFlag here?

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+            invocation.ReturnValue = method.Invoke(_mixin, invocation.Arguments);
+            return;
+        }
+
+        // Event add/remove
+        if (method.IsSpecialName)
+        {
+            if (method.Name.StartsWith("add_"))
+            {
+                HandleEventAdd(method.Name[4..], invocation.Arguments[0] as Delegate);
```
1nf0rmagician: What happens with two explicitely implemented interfaces that share an event name, the cache would override, wouldn't it?

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                return;
+            }
+        }
+
+        // Determine expected argument type from the delegate signature
+        var handlerType = handler.GetType();
+        var expectedArgType = handlerType.IsGenericType
+            ? handlerType.GetGenericArguments()[0]
+            : typeof(EventArgs);
```
1nf0rmagician: I was pondering whether there was any significant performance consideration here indicating that caching the events that have resource event args would be worth a look.  
Then I was wondering whether the benchmarker has a RaiseResourceChanged event during the event processing and whether that would have an influence if e.g. the factory monitor registers to all resource changed events, which all have a resource proxy as an event arg.  
What do you think, would that be something to add to the benchmarker in general? 🤔

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+    #endregion
+
+    #region Resource Reference Conversion
+
+    private IResource ConvertToProxy(IResource instance)
+    {
+        if (instance is null)
+        {
+            return null;
+        }
```
1nf0rmagician, line 237: Worth moving in the typecontroller and using 'as'?

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
```diff
@@ -0,0 +1,302 @@
...
+
+    private static bool IsResourceReference(Type type)
+    {
+        return typeof(IResource).IsAssignableFrom(type)
+               || typeof(IEnumerable<IResource>).IsAssignableFrom(type);
+    }
+
+    private object ConvertResult(object result, Type returnType)
+    {
+        if (result is IEnumerable<IResource> collection and not IResource)
```
1nf0rmagician, line 250: why the and not what would even happen if one would define both on a resource? 😅

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
dbeuchler: This is just how Castle.DynamicProxy API works. CreateInterfaceProxyWithoutTarget(primaryInterface, additionalInterfaces, ...) requires the first argument to be a single Type (the primary interface) and the rest as an array. The primary interface determines the generated proxy type's name and is the "main" type the proxy implements.

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
dbeuchler: Mhmm I thought that I had to fetch them both but the runtime is selecting the correct handler on the target type. Now I removed the name-base and use EventInfo directly from the interface.

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
dbeuchler: No, they are different objects. eventInfo comes from iface.GetEvents() (the interface EventInfo), while targetEvent comes from targetType.GetEvent (the concrete class EventInfo). I thought that we need the target EventInfo  to subscribe on the real target, but I works with the EventInfo from the interface the same way and also fixes the bug that explicit members maybe skipped.

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
dbeuchler: The old implementation also didt properly support two explicit interface events sharing a name. It was also using eventInfo.Name which would either create two separate fields (IL allows same-named fields) or fail at runtime (dont know the behavior. I will add this as a new feature and use ` {type.FullName}.{eventInfo.Name}` as key.

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
dbeuchler: The IL proxy assumed every event is either EventHandler or EventHandler<T> other delegate types were unsupported with an unhandled error during type creation (null-ref?). Here I silently skip them which might be not better. Supporting custom delegates have a high complexity - I will add a `NotSupportedException` here.

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
dbeuchler: NonPublic is needed because explicit interface implementations compile as private methods on the concrete type. Will add a comment

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
dbeuchler: `IResource` could implement `IEnumerable<IResource>`. Without the `and not IResource`, such a resource would match the `IEnumerable<IResource>` branch and get its items proxied — instead of the resource itself. The guard ensures single resources always take the single-resource path some lines below.

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
dbeuchler: GetProxy should stay `Resource` because all code paths guarantee a `Resource` instance. `Proxify` uses `as` but it cannot be changed  because the generic constraint doesnt allow a direct cast (maybe dirty by first casting it to object?).

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
dbeuchler: It could be cached per event key. But GetGenericArguments() on an already-constructed type is lightweight, so the perf-gain would be minimal. Adding a cache for it adds complexity for a micro-optimization. Moving it to the IEnumberable branch and not loading the expectedArgType for IResource is a better optimization.

When the FactoryMonitor registers to all resource-changed events a proxy is created, sure but why do you think this could be problematic? Types are reused, instances are reused, ...

docs/articles/module-resources/resource-management.md
```diff
@@ -25,23 +25,14 @@ For example, the underlying system's structure can be as simple as a single, man
 There could even be an underlying redundant architecture.
 In all cases, the module responsible for mapping `Activities` to `Resources` always sees an enumeration of [IResource](/src/Moryx.AbstractionLayer/Resources/IResource.cs).
 
-Since the architecture of Moryx includes an internal DI-Container for each module, it imposes a restriction on using components from inside the container outside of the [Facade](/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs).
+Since the architecture of MORYX includes an internal DI-Container for each module, it imposes a restriction on using components from inside the container outside of the [Facade](/src/Moryx.Resources.Management/Facades/ResourceManagementFacade.cs).
 Because of that, the resource management applies the proxy pattern to provide access to the resources API while simultaneously hiding the resource instance from the user.
-For more information on the structure of Moryx look into [this article](/docs/articles/framework/index.md).
+For more information on the structure of MORYX look into [this article](/docs/articles/framework/index.md).
```
1nf0rmagician: This is basically a duplicate of what is under the 'Resource Proxies' header

docs/articles/module-resources/resource-proxies.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Interface Discovery
+
+The proxy implements all relevant interfaces of the resource type:
+
+1. Public interfaces derived from `IResource` (excluding `IResource` itself)
+2. Interfaces declared via `[ResourceAvailableAs(typeof(IMyInterface))]` (including non-`IResource` interfaces)
+3. Base interfaces of any already-relevant interface
+
+Derived resource types that do not add new public interfaces reuse the base type's proxy type. For example, if `DerivedResource` extends `SimpleResource` without adding new interfaces, both share the same proxy type.
```
1nf0rmagician, line 26: Is this still the case? Before it was explicit with the proxy linker, but that doesn't exist anymore. Is this an optiomation castle also does under the hood?

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
1nf0rmagician: Ohh waoh thank you, I didn't know that 😅

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
1nf0rmagician: Interesting, good to know. This raises a question for me:  
The old behaviour had a predictable naming convention for types. It was the type name (not interface if I'm not mistaken) of the lowest common denominator in the interface-tree + `_proxy`. Are we sure this convention is not used anywhere and part of the modules 'behaviour' which we would break here?

1nf0rmagician (changes requested):

docs/articles/module-resources/resource-proxies.md
dbeuchler: yesss they are cached in the instance of ProxyGenerator. When CreateInterfaceProxyWithoutTarget is called with the same set of interfaces, Castle returns the cached proxy type rather than generating a new one

docs/articles/module-resources/resource-management.md
dbeuchler: leftover.. removed that

src/Moryx.Resources.Management/Resources/Proxies/ResourceProxyBuilder.cs
dbeuchler: I found some interesting hack in [ResourceActivityBindingResolverFactory.cs](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx.ProcessData.Adapter.ProcessEngine/ResourceActivityBindingResolverFactory.cs)

````cs
// Dirty hack to extract target
var property = proxy.GetType().GetProperties(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
            .FirstOrDefault(p => p.Name == "Target");
var resource = (IResource)property.GetValue(proxy);
````


Replaced that by GetResourcesUnsafe call.

I searched the codebase for IResource.GetType usages out of ResourceManagement. No code depends on the proxy type name. There is no GetType().Name check against proxy type names anywhere. The old type name pattern was never part of any public contract or documented behavior - it was an internal implementation detail of the IL code.

Castle convention: `Castle.Proxies.{PrimaryInterface}Proxy` (e.g. Castle.Proxies.ICellProxy) - hardcoded in Castle internal NamingScope. Castles INamingScope interface is internal, so there is no public API to customize proxy type names. I will add a limitation entry in the table for it. This makes it clear that we have given thought to this matter and how it should be handled.

src/Moryx.Resources.Management/Resources/Proxies/ResourceInterceptor.cs
1nf0rmagician: So, I revisited reflection and cleared up some of the misunderstandings on my part... nvm my comment 😅

1nf0rmagician (approved):

andreniggemann (approved):

PR #1452, opened 2026-08-21
Moved IEEE 754 special handling to EF converter and apply in SqliteProductsContext

dbeuchler:
Move IEEE 754 special value handling from FloatColumnMapper to SqliteProductsContext using an EF Core ValueConverter. 

The previous implementation in the column mapper affected all database providers, even though only SQLite doesn't support these values natively.

Reduces complexity of the ProductStorage


1nf0rmagician (approved):

1nf0rmagician:
Agree, this looks a lot nice and very much like using the context as intended 😁

PR #1454, opened 2026-08-23
Forward ValidationAttribute.ErrorMessage to EntryValidation

### LINKED ISSUE #1450: Forward `ValidationAttribute.ErrorMessage` via `EntryValidation`
We map several validation attributes to the entry validation type which can be applied when processing the entries outside of the system. While the validation attribute in .net allows to provide error messages if an entry violates the conditions given in the validation attribute, we ignore this message and instead rely on generic messages, e.g. in the [ngx-web-framework](https://github.com/PHOENIXCONTACT/ngx-moryx-web/blob/c82203057575888e9ce540c2172824b2d39ad980/projects/ngx-web-framework/entry-editor/input-editor/input-editor.html#L53).  
I'd prefer to keep the generic error messages as a fallback but add error messages from the validation attributes to the entry protocol.

**Additional context**
[ValidationAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute?view=net-10.0)
[EntryValidation](https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/dev/src/Moryx/Serialization/EntryConvert/EntryValidation.cs)

Poseidonas:
Fixes #1450.

`DefaultSerialization.CreateValidation()` walks the validation attributes on a member and reads each one for its bounds, pattern or data type. Everything else on the attribute is dropped, so a message written by the author never leaves the server and the entry editor has only its generic wording to show.

`EntryValidation` now carries an `ErrorMessage`, filled from `ValidationAttribute.ErrorMessage` and left null when no attribute supplied one — which keeps the generic messages as the fallback you asked for rather than replacing them.

**The one decision I had to make**

A member can carry several validation attributes, and there is no obvious winner among their messages. I kept the first one that has a message rather than concatenating, on the grounds that a joined string reads badly in a form field. If you would rather have all of them, or the message belonging to the rule that actually failed, that is a different shape — say the word and I will rework it.

**Verification**

- `dotnet build src/Moryx/Moryx.csproj`: 0 errors.
- Full `Moryx.Tests` suite: **356 passed**, 2 skipped.
- Two tests added to `ValidationTests`: one asserting the message reaches `EntryValidation`, one asserting it stays null for an attribute without a message, so the fallback path is guarded as well. `ValidationDummy` gains one property carrying `ErrorMessage`, and the sub-entry count in `Setup` moves from 8 to 9.
- Both new tests fail to compile against `dev` without the change (`EntryValidation` has no `ErrorMessage`), which is the intended dependency.

`Clone()` copies the new field along with the rest.

I have not touched the consuming side. `ngx-moryx-web` would need to prefer `validation.errorMessage` over its generic string where it renders the message, and I have not looked at what that involves — happy to follow up there if it is wanted.

### DIFF
--- a/src/Moryx/Serialization/DefaultSerialization.cs
+++ b/src/Moryx/Serialization/DefaultSerialization.cs
@@ -145,6 +145,12 @@ public class DefaultSerialization : ICustomSerialization
                     validation.DataType = dataTypeAttribute.DataType;
                     break;
             }
+
+            // The attribute may carry a message explaining what it rejects. Keep
+            // the first one; a member with several attributes has no obvious
+            // winner, and consumers fall back to a generic message when unset.
+            if (validation.ErrorMessage is null && attribute.ErrorMessage is not null)
+                validation.ErrorMessage = attribute.ErrorMessage;
         }
 
         return validation;
--- a/src/Moryx/Serialization/DefaultSerialization.cs
+++ b/src/Moryx/Serialization/DefaultSerialization.cs
@@ -146,11 +146,20 @@ public class DefaultSerialization : ICustomSerialization
                     break;
             }
 
-            // The attribute may carry a message explaining what it rejects. Keep
-            // the first one; a member with several attributes has no obvious
-            // winner, and consumers fall back to a generic message when unset.
-            if (validation.ErrorMessage is null && attribute.ErrorMessage is not null)
-                validation.ErrorMessage = attribute.ErrorMessage;
+            // The attribute may carry a message explaining what it rejects, either
+            // literally or through a resource. FormatErrorMessage resolves both, the
+            // same way DisplayAttribute.GetName does for the display name, and fills
+            // the placeholder with the member's display name where the message has one.
+            //
+            // It is only asked for when the attribute actually declares a message.
+            // Without that guard every member would carry the framework's own English
+            // default, which takes the choice of fallback away from the consumer.
+            //
+            // The first message wins: a member with several attributes has no obvious
+            // winner among them.
+            if (validation.ErrorMessage is null &&
+                (attribute.ErrorMessage is not null || attribute.ErrorMessageResourceName is not null))
+                validation.ErrorMessage = attribute.FormatErrorMessage(attributeProvider.GetDisplayName() ?? string.Empty);
         }
 
         return validation;
--- a/src/Moryx/Serialization/EntryConvert/EntryValidation.cs
+++ b/src/Moryx/Serialization/EntryConvert/EntryValidation.cs
@@ -43,6 +43,15 @@ public class EntryValidation : ICloneable
     public bool IsRequired { get; set; }
 
     /// <summary>
+    /// Message to show when the entry violates one of the rules above, taken from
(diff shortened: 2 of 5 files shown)


src/Moryx/Serialization/DefaultSerialization.cs
```diff
@@ -145,6 +145,12 @@ public virtual EntryValidation CreateValidation(Type memberType, ICustomAttribut
                     validation.DataType = dataTypeAttribute.DataType;
                     break;
             }
+
+            // The attribute may carry a message explaining what it rejects. Keep
+            // the first one; a member with several attributes has no obvious
+            // winner, and consumers fall back to a generic message when unset.
+            if (validation.ErrorMessage is null && attribute.ErrorMessage is not null)
+                validation.ErrorMessage = attribute.ErrorMessage;
```
dbeuchler: Thanks for your first contribution! EntrySerialize is full localizable. The `ErrorMessage` is used for non-localizable error messages. I would prefer also keep the localization consistent and use result of localized `ErrorMessageResourceName` with `ErrorMessageResourceType` like we already do that in DisplayAttribute.

1nf0rmagician: While I did not explicitly say that, it was definitely the intend to cover the localized properties as well (or primarily even) 😅

Poseidonas:
You are both right, and the localized case was the more important half of it. Fixed.

`FormatErrorMessage` is the analogue of the `GetName()` call the display name already goes through, so the resolution now follows the same route the rest of the code takes. Measured against .NET 10 before changing anything:

| attribute | `.ErrorMessage` | `FormatErrorMessage(name)` |
| --- | --- | --- |
| nothing set | `null` | `The Name field is required.` |
| `ErrorMessage` set | the string | the string |
| `ErrorMessageResourceName` + `Type` | **`null`** | the resource, translated |

The third row is what I was reading, so a message given as a resource arrived as nothing at all.

**One decision I would like checked**

`FormatErrorMessage` is only called when the attribute actually declares a message:

```csharp
if (validation.ErrorMessage is null &&
    (attribute.ErrorMessage is not null || attribute.ErrorMessageResourceName is not null))
    validation.ErrorMessage = attribute.FormatErrorMessage(attributeProvider.GetDisplayName() ?? string.Empty);
```

Without that guard, row one applies to every member that never asked for a message, and each one would arrive carrying the framework's own English default. That would take the choice of fallback away from the consumer and hand it a string that ignores the current culture. Leaving it null keeps that decision where it was. If you would rather have the framework default there, it is one condition to remove.

The display name comes from `GetDisplayName()` so a message containing `{0}` gets the same name the entry itself shows, and falls back to an empty string rather than the raw member name when no display attribute is present.

**Tests**

Added a case for the resource path alongside the literal one, with a static class standing in for a generated resource. Worth noting for anyone reading it later: the resource type has to expose a **static property**, not a `const` — I had it as a constant first and every test in the fixture failed with `does not have an accessible static property named ...`.

`Moryx.Tests` is green: 357 passed, 2 skipped.

PR #1457, opened 2026-08-24
Hotfix: Fix GenericEntityMapper / PropertyAccessor / TextColumnMapper

dbeuchler:
https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1364 replaced the serializer call in `GenericEntityMapper.WriteValue` with a manually built `JObject`, but left `HasChanged` on the original `JsonConvert.SerializeObject` path.
Since both are compared against each other, the two paths must produce identical JSON - they no longer did.

- `HasChanged` was always `true`, so every `SaveType` created a new `ProductTypeProperties` version row, even without changes.
- The base properties `Id`, `Name`, `State` and `Identity` ended up in the JSON column as `null`.
- Part links were serialized into the JSON column with `Parent` set, this serializes whole products to json and lead t reference-loop on save.

The manual `JObject` only skipped mapped and read-only properties.
The `DifferentialContractResolver` already skips base properties and references.

Both use `SerializeJson` helper, so the filtering rules live in exactly one place (the contract resolver!).
The concrete type is passed explicitly so `TypeNameHandling.Auto` does not add a redundant `$type` for the root object.

`IsDefaultOrNull` is gone - `JsonSettings.Minimal` uses `DefaultValueHandling.IgnoreAndPopulate`,
so defaults are dropped on write and restored on read anyway.

---

The ReadProperty rewrite in #1364 was not fixing null values, so ConversionAccessor still passed values Convert.ChangeType cannot handle: null values now map to the default of the target type instead of throwing, and Nullable<T> properties are converted via their underlying type.

---

A null complex property was stored as the string "null" in #1364, while a `== null` query looks for SQL NULL, so those products were never found. Nullable structs also came back as their default value instead of null - and the tests never reached the TextColumnMapper at all


src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -191,79 +190,22 @@ public void WriteValue(object source, IGenericColumns target)
...
-            else
-            {
-                jsonObject[prop.Name] = JToken.FromObject(value, JsonSerializer.Create(_jsonSettings));
-            }
-        }
-
-        // Write JSON to the Json - Column
-        _jsonAccessor.WriteProperty(target, jsonObject.ToString(Formatting.None));
+        // Everything else is written as JSON to the JsonColumn
+        _jsonAccessor.WriteProperty(target, SerializeJson(source));
```
1nf0rmagician, line 194: So for my understanding. This line does or does not write the "whole" product type information into  the json column, meaning also fields configured to be placed in specific columns? I know this only reverts the writing to the original so it's no request to change, just want to understand what I though I had already understood🙈

Reading the helpful comment below I understand the proprties should not be placed in the json if they are already mapped to another column. And this was always behaving this way?

1nf0rmagician: Answering my own question, this is the line I should have found last time already

https://github.com/PHOENIXCONTACT/MORYX-Framework/blob/1e4b56e9554b5a89a9a13d728fa71295dab61a19/src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs?plain=1#L50C13-L50C13

It filters all properties mapped to the other columns which makes the custom writing unnecessary

src/Moryx.Products.Management/Plugins/GenericStrategies/GenericEntityMapper.cs
```diff
@@ -191,79 +190,22 @@ public void WriteValue(object source, IGenericColumns target)
...
+    /// </summary>
+    /// <remarks>
+    /// <see cref="WriteValue"/> and <see cref="HasChanged"/> must produce their JSON the very
+    /// same way. Otherwise the comparison in <see cref="HasChanged"/> never matches the stored
+    /// value and every save would create a new version.
+    /// The concrete type is passed explicitly, because <see cref="TypeNameHandling.Auto"/> would
+    /// otherwise add a redundant type name for the root object.
+    /// </remarks>
+    private string SerializeJson(object source)
+        => JsonConvert.SerializeObject(source, source.GetType(), _jsonSettings);
```
1nf0rmagician, line 210: Checking against the code before #1364 the type specification is new, It only removes a redundant info from the json but has no functional changes, correct?

dbeuchler: correct

1nf0rmagician (approved):

PR #1459, opened 2026-08-25
Add additional extensions to retrieve product information from processes

1nf0rmagician:
This is a usual case when populating activity parameters for example, where we want to retrieve information from Task, Recipe, Prodcut Type and maybe Product Instances

### DIFF
--- a/src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
+++ b/src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0
 
 using Moryx.AbstractionLayer.Activities;
+using Moryx.AbstractionLayer.Products;
+using Moryx.AbstractionLayer.Recipes;
 
 namespace Moryx.AbstractionLayer.Processes;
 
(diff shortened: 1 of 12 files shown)


1nf0rmagician:
@dbeuchler please check the "moved" extensions, there are extensions in both packages while only the `IOrderLinkedRecipe` related extensions need to be in `Moryx.ControlSystem`. I want to move the others to keep them group nicely, but that is more of a side change. The added extensions would be most relevant to me.

dbeuchler (changes requested):
Fine for me but we should add unit-tests to Moryx.AbstractionLayer.Tests (`ProcessExtensionsTests`)

1nf0rmagician:
@dbeuchler had to change the name of the moved extensions to `Modify` and `TryModify`. While technically not breaking, it caused ambiguity errors... funny thing I didn't had in mind 😅 
Anyway, I think since the methods are generic anyway the shorter method names are a bit more neat anyway.

Also reused the new extensions in the `SessionExtensions` to reduce duplication

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
```diff
@@ -13,6 +15,164 @@ public static class ProcessExtensions
...
+        {
+            if (process is not ProductionProcess productionProcess)
+            {
+                throw new InvalidOperationException($"Cannot modify an {nameof(ProductInstance)} on a process of type {process.GetType()}");
+            }
+
+            if (productionProcess.ProductInstance is not TInstance instance)
+            {
+                throw new InvalidCastException($"Cannot cast {nameof(ProductionProcess.ProductInstance)} of type "
+                    + $"{productionProcess?.ProductInstance?.GetType()} to {typeof(TInstance)}");
```
dbeuchler: productionProcess cannot be null here. ProductInstance should never be null (also not flagged as nullable). Maybe it is better to use a hard cast here:

````cs
var instance = (TInstance)productionProcess.ProductInstance;
````

This throws an InvalidCastException automatically with a clear message. The only thing you lose is the custom message text, but the default InvalidCastException message from the runtime already includes both the source and target types.

File: src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs [hunk 2]
```diff
@@ -13,6 +15,164 @@ public static class ProcessExtensions
     /// <param name="process">Extended instance of <see cref="IProcess"/></param>
     extension(IProcess process)
     {
+        #region Product Types
```
dbeuchler, line 18: Do we need this single region?

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
```diff
@@ -13,6 +15,164 @@ public static class ProcessExtensions
...
+        /// <![CDATA[
+        /// process.Modify<MyProductInstance>((var instance) => instance.MyProperty = 1);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <exception cref="InvalidCastException">Thrown if the given <paramref name="process"/> does
+        /// not hold a product instance of type <typeparamref name="TInstance"/></exception>
+        /// <exception cref="InvalidOperationException">Thrown if the given <paramref name="process"/>
+        /// is no <see cref="ProductionProcess"/></exception>
+        public TInstance Modify<TInstance>(Action<TInstance> setter) where TInstance : ProductInstance
```
dbeuchler: I dont like `Modify` because it feels like "modifying the process". `ModifyProductInstance` is more straight forward. You get this "ambiguity errors" because of implicit references. Proposal: Keep the name and move the extension in MORYX 12 but:

Why do we need this overhead of this Modify method at all? It is doing the same as `GetProductInstance` and modify the properties.

````cs
process.ModifyProductInstance<SomeInstance>(instance => {
   instance.Foo = 12;
})

// instead of:
var instance = process.GetProductInstance<SomeInstance>();
instance.Foo = 12;
````

The direct getter approach is simpler and more readable because it avoids wrapping every modification in a lambda expression. It makes conditional logic straightforward since you can inspect the instance and decide whether or how to modify it. Debugging is also easier because the call stack remains flat and direct, without additional callback frames to step through. The returned instance can be passed to other methods or used across multiple statements.

In this case there is no reason to use it in a modification callback, there is no transaction and no need to wrap the mutation (e.g. for locking, persistence, notifications, ...)

File: src/Tests/Moryx.AbstractionLayer.Tests/TestData/OtherTestProductInstance.cs [hunk 4]
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2026 Phoenix Contact GmbH & Co. KG
+// Licensed under the Apache License, Version 2.0
+
+using Moryx.AbstractionLayer.Products;
+
+namespace Moryx.AbstractionLayer.Tests.TestData;
+
+public class OtherTestProductInstance : ProductInstance<OtherTestProductType>
+{
+}
```
dbeuchler, line 10: Instead of all the new types, could you not just use the DummyProduct from the TestTools?

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
1nf0rmagician: I appreciate the initial broad categoriation when opening the code file and I see
```csharp
public static class ProcessExtensions
{
    extension(IProcess process)
    {
        Product Types [...]
        
        Product Instances [...]
        
        Activities [...]
    }
}
```

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
1nf0rmagician: > I dont like Modify because it feels like "modifying the process". ModifyProductInstance is more straight forward.

I actually like it because 
```csharp
process.ModifyProductInstance<SomeInstance>(...) // Duplicate mentioning of the fact that we are talking about a product instance

process.Modify<SomeInstance>(...) // Still clear when reading it that we 'modify some instance'
```

> Why do we need this overhead of this Modify method at all?

While I agree we should continue to ask these questions on whether things are still necessary to be maintained, I prefer to keep this for two reasons
1. While the rename is possible with a global Find+Replace even before MORYX 12 migration, removing it would create less obvious migration work and keeping the Framework as stable as possible is also a value we should hold dear
2. For newcomers the modify method gives an indicator in the code which use cases are intended, i.e. "When I recieve a session in my cell, I can Complete, Fail and Trace an activity as well as modify (the instance on) a process" (leaving the getters and setters aside)

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
dbeuchler: I will not block it here but if you have some coffee, you could read https://www.reddit.com/r/csharp/comments/6pmun0/just_say_no_to_c_regions_an_old_article_about/ and https://extractmethod.wordpress.com/2008/02/29/just-say-no-to-c-regions/

In MORYX we reduced regions (in older versions this was massive). Regions nowerdays will be treated as code-smell for the reasons.

src/Moryx.AbstractionLayer/Processes/ProcessExtensions.cs
dbeuchler: > While the rename is possible with a global Find+Replace even before MORYX 12 migration, removing it would create less obvious migration work and keeping the Framework as stable as possible is also a value we should hold dear

`Complete`. `Fail` are methods without a delegate, they are fine. `Trace` is a bit special because it can convert properties from one tracing type to another tracing type (by reflection). 

There is no technical reason for having `Modify` and it introduces complexity which is unecessary for the described reasons: 

> The direct getter approach is simpler and more readable because it avoids wrapping every modification in a lambda expression. It makes conditional logic straightforward since you can inspect the instance and decide whether or how to modify it. Debugging is also easier because the call stack remains flat and direct, without additional callback frames to step through. The returned instance can be passed to other methods or used across multiple statements.

Just a sample of a stacktrace of an exception (the original exception is on line 22): 

````
System.InvalidOperationException : Some exception
   at Moryx.Tests.Serialization.DelegateExceptionTests.<>c.<DelegateException>b__0_0(SomeClass s) in /src/Tests/Moryx.Tests/DelegateTests.cs:line 24
   at Moryx.Tests.Serialization.SomeClass.Modify(Action`1 action) in /src/Tests/Moryx.Tests/DelegateTests.cs:line 34
   at Moryx.Tests.Serialization.DelegateExceptionTests.DelegateException() in /src/Tests/Moryx.Tests/DelegateTests.cs:line 22
````

instead of 

````
System.InvalidOperationException : Some exception
   at Moryx.Tests.Serialization.DelegateExceptionTests.DelegateException() in /src/Tests/Moryx.Tests/DelegateTests.cs:line 22
````

... so three instead of one.

Ultimately, these methods obscure the original location and make debugging more difficult, without there being any technical reason for doing so.

We can introduce them and mark them as obsolete. Then the migration is not a problem and can be done in the next 1,5-2years.

dbeuchler (changes requested):
Could you check the build errors?

dbeuchler (approved):

PR #1460, opened 2026-08-25
Added the Validation logic with the SaveSession Method

holzcodes:
Added workplan validation to the save flow.

Before a workplan is saved, it is now validated for dead ends and isolated nodes. If validation fails, the save operation is rejected and the validation errors are returned to the client.

### DIFF
--- a/src/Moryx.Workplans.Web/WorkplanEditingController.cs
+++ b/src/Moryx.Workplans.Web/WorkplanEditingController.cs
@@ -3,6 +3,7 @@
 
 using System.Drawing;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Moryx.AbstractionLayer.Products.Endpoints;
 using Moryx.AbstractionLayer.Workplans;
@@ -33,7 +34,6 @@ public class WorkplanEditingController : ControllerBase
         var workplans = await _workplans.LoadAllWorkplansAsync();
 
         var recipeSteps = _workplanEditing.AvailableSteps
-
             .Where(step => ModelConverter.ToClassification(step) != WorkplanNodeClassification.Subworkplan)
             .Select(WorkplanStepRecipe.FromStepType);
 
@@ -53,7 +53,6 @@ public class WorkplanEditingController : ControllerBase
 
         return recipeSteps.Concat(subWorkplans).ToArray();
     }
-
 
     [HttpPost("sessions")]
     [Authorize(Policy = WorkplanPermissions.CanEdit)]
@@ -136,6 +135,9 @@ public class WorkplanEditingController : ControllerBase
 
     [HttpPost("sessions/{sessionId}/save")]
     [Authorize(Policy = WorkplanPermissions.CanEdit)]
+    [ProducesResponseType(typeof(WorkplanSessionModel), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(MoryxExceptionResponse), StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(typeof(MoryxExceptionResponse), StatusCodes.Status404NotFound)]
     public async Task<ActionResult<WorkplanSessionModel>> SaveSession(
         [FromRoute] string sessionId,
         [FromBody] WorkplanSessionModel sessionModel)
@@ -154,15 +156,18 @@ public class WorkplanEditingController : ControllerBase
 
         var validation = WorkplanInstance.Validate(
             session.Workplan,
-            ValidationAspect.DeadEnd | ValidationAspect.LoneWolf);
+            ValidationAspect.DeadEnd | ValidationAspect.LoneWolf | ValidationAspect.InfiniteLoop | ValidationAspect.LuckyStreak);
 
         if (!validation.Success)
         {
             var errors = validation.Errors
-                .Select(error => error.Print(session.Workplan))
-                .ToArray();
-
-            return BadRequest(errors);
+                .Select(error => error.Print(session.Workplan));
+     
+            return BadRequest(new MoryxExceptionResponse
+            {
+                Title = "Workplan validation failed",
+                Exception = string.Join(Environment.NewLine, errors)
+            });
         }
 
         await _workplans.SaveWorkplanAsync(session.Workplan);
--- a/src/Moryx.Workplans.Web/WorkplanEditingController.cs
+++ b/src/Moryx.Workplans.Web/WorkplanEditingController.cs
@@ -5,8 +5,10 @@ using System.Drawing;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Moryx.AbstractionLayer.Products.Endpoints;
+using Moryx.AbstractionLayer.Workplans;
 using Moryx.AspNetCore;
 using Moryx.Serialization;
+using Moryx.Workplans.Validation;
 using Moryx.Workplans.Web.Models;
 using Moryx.Workplans.Web.Properties;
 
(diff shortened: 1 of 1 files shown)


src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -128,18 +132,42 @@ private static void UpdateSession(WorkplanSessionModel sessionModel, WorkplanSes
...
+        [FromBody] WorkplanSessionModel sessionModel)
+    {
+        var session = _workplanEditing.OpenSession(sessionId);
+
+        if (session == null)
+        {
+            return NotFound(new MoryxExceptionResponse
+            {
+                Title = Strings.WorkplanEditingController_SessionNotFound
+            });
```
dbeuchler: The NotFound returns a MoryxExceptionResponse object, but the BadRequest returns a string[]. The client has to handle two completely different error shapes from the same endpoint. +. document with `[ProducesResponseType]`

src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -128,18 +132,42 @@ private static void UpdateSession(WorkplanSessionModel sessionModel, WorkplanSes
...
+            session.Workplan,
+            ValidationAspect.DeadEnd | ValidationAspect.LoneWolf);
+
+        if (!validation.Success)
+        {
+            var errors = validation.Errors
+                .Select(error => error.Print(session.Workplan))
+                .ToArray();
+
+            return BadRequest(errors);
```
dbeuchler: Document with `[ProducesResponseType]`

src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -128,18 +132,42 @@ private static void UpdateSession(WorkplanSessionModel sessionModel, WorkplanSes
...
+            {
+                Title = Strings.WorkplanEditingController_SessionNotFound
+            });
+        }
+
+        UpdateSession(sessionModel, session);
+
+        var validation = WorkplanInstance.Validate(
+            session.Workplan,
+            ValidationAspect.DeadEnd | ValidationAspect.LoneWolf);
```
dbeuchler: What about the other `ValidationAspect`s?

src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -51,6 +54,7 @@ public async Task<ActionResult<WorkplanStepRecipe[]>> AvailableSteps()
         return recipeSteps.Concat(subWorkplans).ToArray();
     }
 
+
```
dbeuchler: Fix blank lines

src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -128,18 +132,42 @@ private static void UpdateSession(WorkplanSessionModel sessionModel, WorkplanSes
...
+    [HttpPost("sessions/{sessionId}/save")]
+    [Authorize(Policy = WorkplanPermissions.CanEdit)]
+    public async Task<ActionResult<WorkplanSessionModel>> SaveSession(
+        [FromRoute] string sessionId,
+        [FromBody] WorkplanSessionModel sessionModel)
+    {
+        var session = _workplanEditing.OpenSession(sessionId);
+
+        if (session == null)
```
dbeuchler: You added a null check here but in all other endpoints it returns an NRE. Maybe a ServiceFilter is an option to not have the same code over and over again?

dbeuchler (commented):
Thanks for your contribution! The validation before save is a great addition. I have a few points to check:

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: Done. Changed the validation error response to MoryxExceptionResponse.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: Added ProducesResponseType attributes for the 200, 400 and 404 responses.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: I aded InfiniteLoop and LuckyStreak, so all available validation aspects are now checked.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: Done.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: Should I implement the ServiceFilter in this PR, or would it be better to handle it separately in another ticket?

dbeuchler (changes requested):
I think you introduced a behavior change with this PR.

The HTTP contract changed. New 400/404 response codes where previously only 200 (or 500) was possible. This is fine for me and also a better behavior than 500 but have you checked the UI part? What is the UI doing with 400 and 404?

src/Moryx.Workplans.Web/WorkplanEditingController.cs
dbeuchler: I dont know what's more complicated. Give it a try and maybe it's easier than a Null-Check in every service call. If not, make it in every service-call

seveneleven:
> I think you introduced a behavior change with this PR.
> 
> The HTTP contract changed. New 400/404 response codes where previously only 200 (or 500) was possible. This is fine for me and also a better behavior than 500 but have you checked the UI part? What is the UI doing with 400 and 404?

I really like the idea to validate workplans as early as possible. Adding to Dennis' comment though, this also prevents a workplan from being saved, which could be saved before, which I think should not be done (neither in a future version).
I'd argue, that validation should be done when/before saving, but not preventing it. It might better be handled by the UI and thus leaving the decision to users, if they want to fix an invalid workplan before saving it or not. 
Validity is also something, the UI could display in the overview list.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
```diff
@@ -132,13 +141,43 @@ private static void UpdateSession(WorkplanSessionModel sessionModel, WorkplanSes
...
+            {
+                Title = Strings.WorkplanEditingController_SessionNotFound
+            });
+        }
+
         UpdateSession(sessionModel, session);
+
+        var validation = WorkplanInstance.Validate(
+            session.Workplan,
+            ValidationAspect.DeadEnd | ValidationAspect.LoneWolf | ValidationAspect.InfiniteLoop | ValidationAspect.LuckyStreak);
```
seveneleven: Regarding the 'breaking change' discussion: Why not putting the validation into a separate controller action?
That would also follow SRP and would keep too much logic out of the controller.
Otherwise, why isn't validation moved to the workplans `SaveWorkplanAsync` method to not only have this feature on the REST-API?
While I agree, that validation isn't wrong here, as a user I might want to accept issues with a workplan, but still have it persisted for now and fix them later. 
In this specific use-case, I'd say, it's rather meant to raise awareness instead of an error.

src/Moryx.Workplans.Web/WorkplanEditingController.cs
holzcodes: I did the Null Check in every Service-Call.

1nf0rmagician: #1444 should introduce a first clean implementation of service filters and response object handling which can then be rolled out to all other endpoints 🤗

1nf0rmagician:
> this also prevents a workplan from being saved, which could be saved before, which I think should not be done (neither in a future version).

This is a very valid remark from @seveneleven which leads me to the conclusion, that the validation should be a seperate endpoint. The UI should then have separate calls first to verify and if an issue was detected bring up a prompt clarifying whether the faulty workplan should really be changed.  
Also note, if a faulty workplan is saved it is not indicated as such when selected in the recipes tab in the products UI. This is also an issue to investigate related to this change.

Returned this PR to a draft state and will reassign it at the next opportunity.

PR #1473, opened 2026-08-28
Use proper grammar in exceptions

dbeuchler:
Regarding the MS-guideline, all exceptions shoud use proper grammar and end with period. Also exclamation marks replaced by period to sound gracefully.

https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-proper-grammar

### DIFF
--- a/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
+++ b/src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
@@ -50,7 +50,7 @@ public partial class ProductIdentity : IIdentity
         var match = IdentityPattern().Match(identityString);
         if (!match.Success || !short.TryParse(match.Groups["revision"].Value,
                 NumberStyles.None, CultureInfo.InvariantCulture, out var revision))
-            throw new FormatException($"identityString should consist of <identity>-<revision> instead of {identityString}");
+            throw new FormatException($"identityString should consist of <identity>-<revision> instead of {identityString}.");
 
         return new ProductIdentity(match.Groups["identifier"].Value, revision);
     }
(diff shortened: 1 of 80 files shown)


src/Moryx.AbstractionLayer/Products/ProductIdentity.cs
```diff
@@ -50,7 +50,7 @@ public static ProductIdentity Parse(string identityString)
         var match = IdentityPattern().Match(identityString);
         if (!match.Success || !short.TryParse(match.Groups["revision"].Value,
                 NumberStyles.None, CultureInfo.InvariantCulture, out var revision))
-            throw new FormatException($"identityString should consist of <identity>-<revision> instead of {identityString}");
+            throw new FormatException($"identityString should consist of <identity>-<revision> instead of {identityString}.");
```
andreniggemann: I think this message is misleading without the IdentityPattern

src/Moryx.Drivers.Mqtt/Topics/MqttTopicPrimitive.cs
```diff
@@ -159,7 +159,7 @@ private static string TypeCodeToString(TypeCode type)
             TypeCode.UInt32 => "System.UInt32",
             TypeCode.UInt64 => "System.UInt64",
             TypeCode.String => "System.String",
-            _ => throw new NotImplementedException("This MessageType is not implemented"),
+            _ => throw new NotImplementedException("This MessageType is not implemented."),
```
andreniggemann: Would be worth adding the type here. Otherwise it 's not actionable

src/Moryx.Orders.Management/Implementation/Assignment/OperationAssignment.cs
```diff
@@ -145,7 +145,7 @@ public async Task Restore(IOperationData operationData)
             if (result)
                 continue;
 
-            throw new InvalidOperationException(restoreStep.GetType().Name + " was failed");
+            throw new InvalidOperationException($"{restoreStep.GetType().Name} was failed.");
```
andreniggemann: Shouldn't it be step 'has failed'

andreniggemann (dismissed):
I know you only wanted to add periods, but I took the opportunity to look over exception messages.
If you want you can fix those otherwise I'll create a PR later to target them

andreniggemann (approved):

Issue #1469, opened 2026-08-27
Add a more prominent queue when a new visual instruction is added to the currently opened instructor


1nf0rmagician:
**Is your feature request related to a problem? Please describe.**
Users not experienced with the Visual Instructions UI did oversee the button that appears in the bootom left when a second visual instruction is added to an instructor that already has an open instruction. While we do not automatically switch to the new instruction, a more prominent clue when new instructions pop up in the background would be nice.

**Describe the solution you'd like**
No good idea for how to improve the UX yet unfortunately.

**Related issues/discussions**

dbeuchler:
closed by https://github.com/PHOENIXCONTACT/MORYX-Framework/pull/1475

PR #1475, opened 2026-08-31
Rework of visual instructions: Focus mode and action-bar

### LINKED ISSUE #1470: Consider options to extend the possibilities for applications to decide when instructions should pop to front
Currently, we have the hardcoded behaviour in the Visual Instructions UI, that the current instructions is kept active even if instructions pop up in the background. This is the lowest common denominator we settled upon when initially building the UI, to prevent instructions which require inputs to change during editing because of server activity in the background; especially since this would clear the the entry editor on the instruction.  
However, there are multiple use cases in which a static visual instruction is kept on display providing general information to users. Specific instructions that require active or timely interventions should in these use cases be pushed in the foreground, even if their appearance would be made more notable as requested in #1469.

No favourite solution yet, suggestions for discussion below.

**Describe alternatives you've considered**
1. Feature Extension in the Backend
  a. Add a "push-to-front" option to display and execute methods, which is transmitted to visual instructors
  b. Add a "priority" level to display and execute methods which can be used to layer instructions and make the "push-to-front" behaviour depending on which other instruction is currently active, i.e. only push to front if the current instruction is of lower priority
2. Feature Extension in the Frontend
  a. Add a condition to only stick to the currently active instruction if it contains inputs
  b. Add a configuration option in the UI's settings menu with options for 'Push Newest To Front', 'Push Newest to Back', 'Only Keep Input Instruction In Front'

Each of the described solutions provides a different set of/versitility in its features and some could even be combined if desired. The open question remains which of these is the best fit for requirements and framework consistency.

**Acceptance criteria**
Do not default to another behaviour without further configuration options, instead increase configurability of the systems behaviour.

dbeuchler:
### Focus Mode

Added "FocusMode" to settings dialog for #1470, now it can be decided between 

- Keep input in front: Stay on the current instruction only while editing inputs, otherwise switch to the newest.
- Push Newest To Front: Always switch to the newest instruction when it arrives.
- Keep Current In Front: Always stay on the current instruction, new ones appear in the background.

As far as I remember the WPF UI behavior was always "Push newest" since the inputs was introduced it was "Keep input in front".

I reworked the whole settings dialog

- Added settings for the focus mode
- Centralized settings in InstructionStateService which migrates all settings values automatically to local storage JSON obj. Added todos to remove the legacy keys in the next major

Screenshot 2026-08-31 at 12 07 33

### Design rework

The second change is a design rework. I reduced the whole scss and removed unnecessary things and customizations which not matching the angular material styling. (Dont wonder, I explicitly using rem for fonts here to have a relative font size)

An action-bar was created as component and improved the pagination for multiple instructions for #1469. The pagination now uses the same buttons styles and same size instead of the missused fab-buttons. If there is a newer instruction available, a "move to newest" button is shown.

Screenshot 2026-08-31 at 12 12 07

Screenshot 2026-08-31 at 12 12 25

The whole bar is build as "sticky" means it lays on top of the content to provide this transparency scrolling behavior, so to not just "hide" the content behind it.

Screenshot 2026-08-31 at 12 05 31

Screenshot 2026-08-31 at 12 13 25

Also the results buttons are now more prominent by using raised buttons. 

A thought for the "move to newest" button: Maybe add a ripple effect (like sonar). Just a sample, **not implemented**. 

Screen Recording 2026-08-31 at 12 27 17

close #1470 #1469


src/Moryx.VisualInstructions.Web/app/src/app/app.ts
```diff
@@ -31,66 +30,48 @@ import { MatIconModule } from '@angular/material/icon';
...
   private instructionService = inject(InstructionService);
+  private instructionStateService = inject(InstructionStateService);
   private destroyRef = inject(DestroyRef);
 
-  protected environment = environment;
-  protected clientIdentifier: string = '';
-
-  private readonly cookieName = 'moryx-client-identifier';
-  private readonly cookieLifetimeDays = 365;
+  protected clientIdentifier = this.instructionStateService.instructor;
```
MathoMathiasCamara, line 41: does this clientIdentifier becomes automatically a linkedSignal ? (as it is i am not sure it will be auto updated if you have 2 instructors and you change it.)

MathoMathiasCamara (changes requested):
i like the suggestion.

src/Moryx.VisualInstructions.Web/app/src/app/app.ts
dbeuchler: it is just the signal itself, mapped by reference

MathoMathiasCamara (dismissed):
I like the proposal and the new design

dbeuchler:
Thanks for the approval. Waiting for @1nf0rmagician design review before merge. Discussed improvements for possible (unplanned) additional features: 

1. Provide optional "Preview" as in the old WPF UIs
2. More intelligent swipe logic like a typical image slider
3. Ripple/Sonar effect on "newest" button

1nf0rmagician:
@dacky179, @szThunder fyi

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
```diff
@@ -4,33 +4,147 @@
...
+  private settings: InstructionSettings;
+
+  private _fullscreen;
+  readonly fullscreen;
+
+  private _focusMode;
+  readonly focusMode;
+
+  private _instructor;
+  readonly instructor;
```
1nf0rmagician: We usually define the type on field definitions, isn't this implicit any type this way?

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
```diff
@@ -4,33 +4,147 @@
 */
 
 import { Injectable, signal } from '@angular/core';
+import { environment } from '../../environments/environment';
 
-const FULLSCREEN_KEY = 'workerInstructions.fullscreenEnabled';
+export enum FocusMode {
```
1nf0rmagician: Similar to c# we should stick to one type one file, as long as it is not a private interface, don't you think? 🤔

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
```diff
@@ -4,33 +4,147 @@
...
+
+  private readonly storageKey = 'moryx.visualInstructions.web.settings';
+  private readonly cookieName = 'moryx-client-identifier';
+  private readonly cookieLifetimeDays = 365;
+
+  private readonly legacyFullscreenKey = 'workerInstructions.fullscreenEnabled';
+
+  private readonly defaultSettings: InstructionSettings = {
+    fullscreen: false,
+    focusMode: FocusMode.Input,
```
1nf0rmagician: Actually, I think the current behaviour is `FocusMode.Stay` which I would ste as a default to not introduce a behaviour change. After all, it is a viable use case to want to work through instructions one after another, even if they don't have inputs 🙈

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.ts
```diff
@@ -6,47 +6,62 @@
...
-  instructorName: string;
-}
+  protected selectFocusMode(mode: FocusMode): void {
+    this.selectedFocusMode.set(mode);
+  }
 
+  protected apply(): void {
+    this.instructionStateService.setFocusMode(this.selectedFocusMode());
+    this.instructionStateService.setInstructor(this.selectedInstructor());
+    this.dialogRef.close(true);
```
1nf0rmagician: Rather don't return anything here? The boolean value is never used

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  .radio-label {
+    font-weight: 500;
+  }
+
+  .radio-description {
+    display: block;
+    font-size: 12px;
+    color: var(--mat-sys-on-surface-variant);
+  }
```
1nf0rmagician, line 25: Do you know by chance, whether this matches the styling of the 'Boolean Entry Editor' which also consists of a header and description element or would you argue since this is a radio button a different styling would be alright anyway?

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.scss
```diff
@@ -0,0 +1,31 @@
+.loading-spinner {
+  flex: 1 1;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.dialog-container {
+  width: 100%;
+  overflow: visible;
```
1nf0rmagician: This should be the default anyway

dacky179:
I like the new options 👍 

If you give us the little one, you can expect us to take the hole arm...
Can we push the new instruction to front even if we are not in the instruction module?
The pagination and the control buttons does not have enough contrast in dark mode. maybe you could add something like a white shadow?
Could we get a preview list on the side like notification? maybe configurable?

src/Moryx.VisualInstructions.Web/app/src/app/components/action-bar/action-bar.html
```diff
@@ -0,0 +1,27 @@
...
+    </button>
+    @if (hasNext()) {
+      <button mat-raised-button class="bar-button" (click)="viewNewest.emit()">
+        <mat-icon class="bar-icon">keyboard_double_arrow_right</mat-icon>
+      </button>
+    }
+  </div>
+}
+@if (results().length) {
+  <div class="action-bar-results">
```
1nf0rmagician, line 19: While I see some charm in replacing the 'card with actions' with this action bar, I would argue that we should use the [material toolbar](https://m3.material.io/components/toolbars/guidelines) as a foundation then. There are options with text buttons and for having seperate groups I think we could also find guidance there 🤔

dbeuchler:
> I like the new options 👍

thanks :-) 

> If you give us the little one, you can expect us to take the hole arm... Can we push the new instruction to front even if we are not in the instruction module? 

Not with the current notificiation api (badge is just checking if its an array), but good point to add an issue for this feature request -> https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/1490

> The pagination and the control buttons does not have enough contrast in dark mode. maybe you could add something like a white shadow? 

This is something we can and should not control, its completly from the angular-material theme. One point on my list is to improve the dark-theme a bit, it is in many regions to dark and without contrast.

> Could we get a preview list on the side like notification? maybe configurable?

Already talked about that with @MathoMathiasCamara - maybe if there is time in a followup PR.

1nf0rmagician (commented):
Thanks a lot for the contribution! 

One thing I couldn't see on the screenshots, could you check that 
a. the fullscreen button for media does not overlap with the title (not sure whether the anchoring is correct there)
b. the buttons of the media carousel are still correctly aligned with the new action bar, if there are more than 1 media items on a single instruction

Thank you!

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
dbeuchler: No its defined by the consturctor and not implicit any. Can add types if you want but the result is the same.

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
dbeuchler: I thought Input was the default.. when Stay was the default, I will change that

src/Moryx.VisualInstructions.Web/app/src/app/components/action-bar/action-bar.html
MathoMathiasCamara: what @dbeuchler  did is already close enough to the material toolbar on big screen. The question is should it be a 1 to 1 ? There is also no angular toolbar that matches this.

image

dbeuchler: The original card style was destroyig other styles. I explicitly chose a custom implementation here because the toolbar shown in the screenshots does not exist in Angular Material. `mat-toolbar` is primarily intended for top toolbars, and customizing it for this use case would require more code than the version I have built. Sometimes it makes no sense to use existing components just for the sake of using them, even if alternatives would be simpler.

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
dbeuchler: Added imports and explicitly defined types.

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
dbeuchler: Stay is now the default

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.ts
dbeuchler: leftover from old dialog

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.scss
dbeuchler: Its not the same. The boolean-editor uses just a checkbox value + a mat-hint. This case is explicitly use a different styling because it is more an extending title for me. I tried both, used the hint (which works only on form inputs by the way, so I would have to copy the style from boolean-editor aswell. Writing everything to the title does not look that good either.

src/Moryx.VisualInstructions.Web/app/src/app/dialogs/settings-dialog/settings-dialog.scss
dbeuchler: nop. mat-dialog-content uses auto by default (to make long dialogs scrollable). But overriding it feels wrong. When the instructor-list grows, we now scroll the whole dialog. Removed the whole scss class

dbeuchler:
> b. the buttons of the media carousel are still correctly aligned with the new action bar, if there are more than 1 media items on a single instruction

@1nf0rmagician no, they are aligned as before because they a part of the content. In a follow-up PR we could harmonize the style instead of the custom buttons.

Screenshot 2026-09-08 at 10 30 48

dbeuchler:
> a. the fullscreen button for media does not overlap with the title (not sure whether the anchoring is correct there)

@1nf0rmagician no its not broken, looks the same as before but after a short discussion with @MathoMathiasCamara we identified other UX issues which we put in another PR.

 
Screenshot 2026-09-08 at 11 21 59

src/Moryx.VisualInstructions.Web/app/src/app/services/instruction-state.service.ts
1nf0rmagician: > No its defined by the consturctor and not implicit any

Fascinating, didn't know that. But yeah not it is unified with all other occurances. Thanks!

src/Moryx.VisualInstructions.Web/app/src/app/components/action-bar/action-bar.html
1nf0rmagician: > The original card style was destroyig other styles.

I didn't know that and didn't find that in the description unfortunately, so I was more looking for reasons for the new than reasons against the old

> mat-toolbar is primarily intended for top toolbars, and customizing it for this use case would require more code than the version I have built. 

If you have explored that path I trust you in that this is the easier solution, I just fear that we can discuss again now how big and how widely spaced each of the buttons are, as I don't see a clear guideline other than taste for evaluating those parameters 🙈

1nf0rmagician (approved):
Thanks for taking the time to rework some of the stuff and especially for testing the media stuff as well! 🤗 Want to squash some of the very small commits from the review, whichever way is fine for me?

PR #1483, opened 2026-09-03
Rework module lifecycle components

### LINKED ISSUE #23: Concentrate Module LifeCycle in one place
Currently the life cycle and state machine is spread across different components, even with some redundancies. This complicates refactorings and debugging, as well as unnecessary code.

dbeuchler:
### ModuleLifecycleController

The module lifecycle was spread across `ModuleStarter`, `ModuleStopper`, and `ModuleManagerComponent` with shared mutable state and redundancies. This should be simplified by https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/23.

Single ModuleLifecycleController now owns all lifecycle orchestration (initialize, start, stop, reincarnate). `ModuleManager` delegates to the controller and remains the public API

### Some fixes

Removed redundant `InitializeAsync` call becaus he state machine no-ops redundant initialize calls, but the extra call was unnecessary.

Fixed synchronization on _waitingModules. It was locked by lock but also by semaphore. Now it's completely locked by semaphore. (Added extension for running sync-action in asynchronous context.

State change notification of ServerModule used `GetInvocationList()` + `DynamicInvoke` (reflection-based). Now uses typed `EventHandler<ModuleStateChangedEventArgs>` cast with direct invocation

### StartModulesAsync

`IModuleManager.StartModulesAsync` returns before modules reach running state. This was identified by @1nf0rmagician and tried to fix by #950 which is still WIP. From my point of view this is not correct:

`MoryxHost.StartAsync` awaits `StartModulesAsync` as part of the ASP.NET Core hosted service pipeline. If it blocked until all modules are running, the HTTP pipeline would not start until every module has finished — no endpoints, no UI. Also if modules started in other environments (over Program.cs) the app cannot be fully start because its waiting for the modules.

Module starts are fire and forget via `Task.Run` and their progress is observable through the `ModuleStateChanged` event. In next major, we could make this method synchronous again or rename it to a better name. Also the Module-Starts of Modules which do not have dependencies should be decoupled form this Task. Maybe only the initialization could be part to ensure facades are instantiated.

close #23 #950


src/Moryx.Runtime.Kernel/Modules/Components/ModuleLifecycleController.cs
```diff
@@ -90,60 +127,51 @@ private async Task StartModule(IServerModule module, CancellationToken cancellat
 
     private async Task ExecuteModuleStart(IServerModule module, CancellationToken cancellationToken)
     {
-        // Should be caught by ServerModuleBase but better be safe than sorry
-        try
-        {
-            await module.InitializeAsync(cancellationToken);
-            await module.StartAsync(cancellationToken);
```
andreniggemann, line 97: I think this is the correct way to do it long term, but I guess this could be considered a behavior breaking change.

As an example the FailureState of a module expects Stop or Initialize next.
Currently you could just start the module again. Now you would have to explicitly reincarnate it.

andreniggemann: After talking to @dbeuchler and checking the code again, I am convinced this should not be an issue, 
because Initialize is always called somewhere beforehand

andreniggemann (approved):

src/Moryx.Runtime.Kernel/Modules/Components/ModuleLifecycleController.cs
dbeuchler: The LifeCycleController always calls Initialize first. The fact that Initialize was being called twice was actually a bug that has existed since v1.0 (theres even a ticket for it somewhere internally).

  The call chain looks like this:
  - Reincarnate calls StartAsync -> Initialize
  - StartAllModules also calls Initialize for all modules first
  - and StartAsync does so itself as well

1nf0rmagician:
> IModuleManager.StartModulesAsync returns before modules reach running state. This was identified by @1nf0rmagician and tried to fix by #950 which is still WIP. From my point of view this is not correct

@andreniggemann, @dbeuchler did you have a more in depth discussion about this topic? Because I am still of the opinion, that the module manager should allow to await the start of a module to the point where it is actually started. I do not dispute, that we will fire and forget the task in the Program.cs, the whole contract of `IModuleManager` does not make much sense to me without this ability. While I was fully supportive of the fact that we do not fully implement e.g. the cancellation token handling when moving to the async API initially, as it would have prevented the API breaking changes from being introduced for MORYX 10, I still thought that their intention was to enable this.

I specifically want to discuss the case that, a module gets a cancellation token in its `OnStart` method. At this point I as the module developer expect, that in order to provide a clean behaviour on the cancellation of a module start, I'd need to react to the cancellation of this token. If we assume the current behaviour to be intended however, I would need to store the cancellation token handed to a `IModuleManager.StartModule` and always cancel it before I call `IModuleManager.StartModule` as I could never be sure that the modules indeed have startet, and would need to make sure not to trigger the stop while the start is still running. 

_Note: I only started the discussion here, as this PR closed the WIP PR. I fully support the merging of the components though_

andreniggemann:
For me we need to draw a distinction between the two cases `StartModuleAsync` and `StartModulesAsync`.

@dbeuchler  and I talked about `StartModulesAsync` and I agree with him that, that this should not be awaitable and could be made synchronous.
Awaiting this is most likely a foot gun, we should not expose without a clear use case, that does not exist to my knowledge.
There are concerns about the endpoint of that Task. Is it completed, if all Modules are started or is it already finished, when no more modules want to start (some may have failed them selves, some may have missing dependencies, others might have a StartBehavior aside from Auto configured).

I would say `StartModuleAsync` maybe different, because there's a much clearer use case for me and a clearer user expectation when this should complete.

The point regarding not calling Stop while the Module is still starting is something that should be handled appropriately by the Module StateMachine and not by an outside observer.

1nf0rmagician:
>The point regarding not calling Stop while the Module is still starting is something that should be handled appropriately by the Module StateMachine and not by an outside observer.

Should it? It is a viable option, but it means to take care of it in every module separately 🤔

andreniggemann:
> > The point regarding not calling Stop while the Module is still starting is something that should be handled appropriately by the Module StateMachine and not by an outside observer.
> 
> Should it? It is a viable option, but it means to take care of it in every module separately 🤔

I don’t think that an issue, because basically all modules inherit from ServerModuleBase.

The responsibility needs to be in a central place, so either the module or the LifeCycleManager, because multiple components may issue conflicting requests to start and stop modules, i.e. MoryxHost, LifecycleHooks or the ComandCenter

dbeuchler:
> The responsibility needs to be in a central place,

+1

I already reduced this complexity in this PR

PR #1484, opened 2026-09-03
Unified toolbar style in operation recipes view to other views 

ArthurPxC:
The commands in the operation recipes view, which were previosly placed in a separate side bar on the right, were migrated into the top part of the left side bar. This way the layout is unified to other views like Products, Resources, and so on.

image

The contrast to the background is not part of this issue. Someone needs to provide better background images.

### DIFF
--- a/src/Moryx.Orders.Web/app/src/app/components/operation-recipes/operation-recipes.html
+++ b/src/Moryx.Orders.Web/app/src/app/components/operation-recipes/operation-recipes.html
@@ -1,71 +1,50 @@
 @if (isLoading()) {
-  <mat-progress-bar
-    mode="indeterminate"
-  class="operation-recipe-loading"></mat-progress-bar>
+  <mat-progress-bar mode="indeterminate" class="operation-recipe-loading"></mat-progress-bar>
 }
 <mat-sidenav-container class="operation-recipe-container">
-  <mat-sidenav opened mode="side" position="start">
-    <div class="">
-      <mat-toolbar
-        [style.background-image]="'url(' + operationRecipeToolbarImage + ')'"
-        color="primary"
-        class="operation-recipe-toolbar">
-        <mat-toolbar-row>
-          <a mat-icon-button [routerLink]="['/operations']" mat-icon-button aria-label="Go back to operations">
+  <mat-sidenav opened mode="side" position="start" disableClose="true">
+    <mat-toolbar [style.background-image]="'url(' + operationRecipeToolbarImage + ')'" color="primary"
+      class="operation-recipe-toolbar">
+      <mat-toolbar-row>
+        @if (!isEditMode()) {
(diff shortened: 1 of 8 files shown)


src/Moryx.Orders.Web/app/src/assets/languages/it.json
```diff
@@ -59,7 +59,11 @@
     "WORKPLAN": "Piano di Lavoro",
     "CLASSIFICATION": "Classificazione",
     "PROPERTIES": "Proprietà",
-    "VERSION": "Versione"
+    "VERSION": "Versione",
+    "GO BACK": "Torna indietro",
```
dbeuchler: Key is "GO_BACK"

src/Moryx.Orders.Web/app/src/assets/languages/zh.json
```diff
@@ -59,7 +59,11 @@
     "WORKPLAN": "工作计划",
     "CLASSIFICATION": "分类",
     "PROPERTIES": "属性",
-    "VERSION": "版本"
+    "VERSION": "版本",
+    "GO BACK": "返回",
```
dbeuchler: Key is "GO_BACK"

src/Moryx.Orders.Web/app/src/app/components/operation-recipes/operation-recipes.html
```diff
@@ -1,101 +1,75 @@
...
+              <mat-icon>edit</mat-icon>
+            </button>
+          }
+        } @else {
+          <span class="spacer"></span>
+          <button type="submit" mat-icon-button aria-label="Save the edited resource" (click)="onSave()"
+            [matTooltip]="TranslationConstants.RECIPES.SAVE | translate" matTooltipShowDelay="500">
+            <mat-icon>save</mat-icon>
+          </button>
+          <button type="reset" mat-icon-button aria-label="Cancel Editing" (click)="onCancel()"
```
dbeuchler: You use type=reset here, in all other toolbars we use just normal button. "reset" is normally used on forms to reset values. I would also use button here fore consistency.

dbeuchler (changes requested):
Actually, we wanted to introduce more of these right-hand toolbars in the UIs (see Orders) as a 2nd toolbar. This has already been discussed extensively with @1nf0rmagician and @MathoMathiasCamara. This change is therefore a step backwards. Nevertheless, Recipes differs from Products and Resources. So we don't yet have an overall concept. In order for the toolbar-image combinations to be consistent, we should do it this way as an interim solution.

dbeuchler (approved):

dbeuchler:
@1nf0rmagician please have a final review too

1nf0rmagician (approved):
Thanks for getting back to this, I remember we talked about it in the past. It is now inline with products and resources which is great! 🙂

PR #1487, opened 2026-09-04
fixed the the save on the step properties while the editor is open

holzcodes:
Workplan step properties are not saved when the editor is still open when saving

Steps to reproduce:

Prerequisites
- a saved workplan with a configurable task

1. open workplans module
2. open the workplan
3. click on the configurable task and change its configuration in the entry editor
4. click on save (do not close editor beforehand)
5. open a different module
6. open workplans module again
7. click on the same task and check whether the setting is still changed (which it isn't)

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(rtk ls *)",
+      "Bash(rtk grep *)",
+      "Bash(rtk git *)"
+    ]
+  }
+}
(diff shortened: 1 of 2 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(rtk ls *)",
+      "Bash(rtk grep *)",
+      "Bash(rtk git *)"
+    ]
+  }
+}
```
dbeuchler, line 9: Do not push your personal AI sttings

dbeuchler (changes requested):
Can you maybe explain what if fixes? Maybe an attached bug?

1nf0rmagician:
Reverted to a draft and will reassign at the next possible opportunity.

I added the description of the issue that was missing however, to allow for easier continuation of this topic in future.

PR #1492, opened 2026-09-08
Include resource information in log notifications.

kumarnitin17:
## Summary

Added the resource ID and resource name to resource-related log notifications, so the resource that generated a log can be identified.

## Changes
- Included the resource ID together with the resource name when creating the resource child logger.
- Added the resource information at the end of the notification message.
- Kept the original log message unchanged and added the resource context separately.


src/Moryx.AbstractionLayer/Resources/Resource.cs
```diff
@@ -64,9 +64,9 @@ public abstract class Resource : ILoggingComponent, IResource, IAsyncInitializab
     public IReferences<Resource> Children { get; set; }
 
     /// <inheritdoc />
-    Task IAsyncInitializable.InitializeAsync(CancellationToken cancellationToken)
+    Task IAsyncInitializable.InitializeAsync(CancellationToken cancellationToken)
     {
-        var loggerName = Name?.Replace(".", "_"); // replace . with _ because of logger child structure
+        var loggerName = $"{Id}-{Name}".Replace(".", "_"); // replace . with _ because of logger child structure
```
dbeuchler, line 69: This is breaking, we cannot change the logger name. I have configured every logger specifically in nlog. Using the Id in the logger name makes configuration hard dependent to the database. It would not be possible anymore to deploy generic configuations.

dbeuchler (changes requested):

kumarnitin17:
- Adds instance-level resource properties to `ModuleLogger` without changing the logger name.
This allows log entries created by resources to include additional resource information while keeping the existing NLog logger configuration unchanged.

## Changes

- Added support for instance-level properties in `ModuleLogger`.
- Added a safe `TrySetProperty()` extension for `ILogger`.
- Added the following properties to resource child loggers:
   - ResourceId
   - ResourceName
   - ResourceType
  

- The logger names are not changed.
- The existing message details are kept, and the resource details are added.

## Minimal NLog Configuration 
The resource properties can be included in an NLog target layout as follows:
layout="${message} | Id=*{event-properties:item=ResourceId}*| Name=${event*properties:item=ResourceName} |*Type=${event-properties:item=ResouceType}"
Screenshot 2026-09-16 142313

dbeuchler:
Can you add a hint, when it is ready to be reviewed? (Please follow our coding conventions e.g. keep line until 120chars)

PR #1493, opened 2026-09-08
Add display attributes to `BatchIdentity`

1nf0rmagician:


dbeuchler (changes requested):
Could you add it to the Strings.resx?

src/Moryx.AbstractionLayer/Moryx.AbstractionLayer.csproj
```diff
@@ -12,6 +12,7 @@
 	  <ProjectReference Include="..\Moryx\Moryx.csproj" />
   </ItemGroup>
 
+  <!-- TODO: Make Strings internal in MORYX 12 -->
```
dbeuchler: We cannot make them internal, if they would be internal dotnet cannot resolve them, thats why they are public.

dbeuchler (changes requested):

src/Moryx.AbstractionLayer/Moryx.AbstractionLayer.csproj
1nf0rmagician: What I mean is, use `ResXFileCodeGenerator` instead of `PublicResXFileCodeGenerator` which makes the `Strings.Designer.cs` class internal but still allows the strings to be resolved within the same assembly, it only prevents other projects from referencing strings in this one which I don't think we need for any of the strings in there.

dbeuchler: > strings to be resolved within the same assembly

And thats the Point, within the same assembly. WPF developers know the pain ;-). They must be public if you have Reflection-based consumers when you e.g. use EntryConvert the strings are resolved by a different assembly and this crashes on runtime. Give it a try, maybe Im completly wrong.

1nf0rmagician: Removed, sorry for the back and forth. Apparently I have only added localization to the endpoints so far and hence never stumbled upon the issue. Will keep it in mind for the future though...

dbeuchler (approved):

Issue #1476, opened 2026-09-01
Reconsider ServerModules BehaviourAccess


dbeuchler:
`IBehaviourAccess<T>` and its implementation (ABehaviourAccess, ABehaviourAccess<T>, StartBehaviorAccess, FailureBehaviourAccess) exist only to get/set two enum properties on ManagedModuleConfig. That's 5 classes for something that could be a direct property access.

The only consumer is ModulesController, and even ModuleStarter internally bypasses the abstraction and accesses StartBehaviour directly.

### FailureBehaviour

FailureBehaviour can be set via the CommandCenter and is stored in ManagedModuleConfig, but nothing in the runtime actually evaluates it.

In older framework versions this was used to control module behavior when e.g. a critical ParallelOperations job failed — it would stop the module. That logic was removed at some point but the config flag and UI remained.

Reintroduce failure handling based on this flag or remove FailureBehaviour from the config and the Command Center UI entirely.

andreniggemann:
I support removing FailureBehaviour. Stopping the Module, because one action produced a Critical Error Message does not appear like a common use case.
Demand seems also limited, given that nobody complained, since this feature has been effectively removed with MORYX 6.

PR #1495, opened 2026-09-09
Removed FailureBehaviour from CommandCenter

dbeuchler:
Removed unused FailureBehvior from CommandCenter. 

- Mentioned the removal in #1477. 
- Should be reconsidered in https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/1476

This removes not a feature. It is currently unused in the modules and has no effect.

### DIFF
--- a/src/Moryx.CommandCenter.Web/app/src/common/components/ModuleHeader.tsx
+++ b/src/Moryx.CommandCenter.Web/app/src/common/components/ModuleHeader.tsx
@@ -8,15 +8,13 @@ import Tabs from "@mui/material/Tabs";
 import * as React from "react";
 import { connect } from "react-redux";
 import { Link } from "react-router";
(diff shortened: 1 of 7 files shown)


andreniggemann (commented):
Given that the mentioned issue for reconsideration is not resolved, wouldn't it make more sense to discuss the future of this feature first?

Than we can remove it from the UI and deprecate it in the backend or implement logic that looks at the FailureBehavior.

dbeuchler:
> Given that the mentioned issue for reconsideration is not resolved, wouldn't it make more sense to discuss the future of this feature first?

Thats why the issue exists. We cannot change the behavior of ServerModules in 10 and therefore its a ticket for 12. The UI dropdown does not make any change on the behavior and therefore I removed it from the UI and mentioned the ticket.

andreniggemann (approved):

PR #1496, opened 2026-09-09
Do not inherit `ResourceConstructorAttribute`

1nf0rmagician:
### Summary

ResourceConstructorAttribute must not be derived because a resource constructor can be used in base classes for general resource constructors which however should be hidden in more specific derived classes which in turn could define different resource constructor methods.

### Linked Issues

The issue was found while creating the `MaterialContainer` resource definitions in #1368

### Breaking Changes

This is a behaviour change, the question I honestly cannot answer is, whether this was the intentional behaviour or just an overlooked field on the `AttributeUsageAttribute`.

I'd ask @szThunder, @dbeuchler, @aniggemann and @dacky179, could you please tell me wheether there are any known occasions were the `ResourceConstructorAttribute` was used and intentionally inherited to a derived class? If there are, I'll move this as a proposal for MORYX 12.

### Checklist for Submitter

### Review

**Typical tasks**

- [ ] *Tests* are extended

**Clean code**

### DIFF
--- a/src/Moryx.AbstractionLayer/Resources/Attributes/ResourceConstructorAttribute.cs
+++ b/src/Moryx.AbstractionLayer/Resources/Attributes/ResourceConstructorAttribute.cs
@@ -6,7 +6,7 @@ namespace Moryx.AbstractionLayer.Resources;
 /// <summary>
 /// Attribute to decorate methods that can be used to construct a resource instance
 /// </summary>
-[AttributeUsage(AttributeTargets.Method)]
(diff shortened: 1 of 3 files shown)


dbeuchler (approved):
I also dont see a case where Inherited = true makes sense.

dacky179:
Never used the construcor attribute. If I have a derived class, it doesn’t necessarily have to change the constructor; perhaps only a few options are added. those would lose the resource constructor; but i also see no other way to hide obsolate constructor from base if it would be unconclusive

PR #1512, opened 2026-09-11
Fix duplicate module start race-condition

dbeuchler:
When a module depends on multiple other modules (e.g. `D` depends on `B` and `C`), each dependency completing removes the waiting module from its own waiting list and calls `StartModule`. If `B`and `C` finish around the same time, both threads pick up `D` and start it. The semaphore only protects dictionary access, not the StartModule call afterwards.

Note there is no Test added because I have NO idea how to test this race-condition. It also occurs not every time depending on the systems-performance.

### DIFF
--- a/src/Moryx.Runtime.Kernel/Modules/Components/ModuleLifecycleController.cs
+++ b/src/Moryx.Runtime.Kernel/Modules/Components/ModuleLifecycleController.cs
@@ -14,7 +14,7 @@ internal class ModuleLifecycleController
     private readonly ILogger _logger;
     private readonly ModuleManagerConfig _config;
     private readonly SemaphoreSlim _waitingModulesSemaphore = new(1, 1);
-    private readonly Dictionary<IServerModule, ICollection<IServerModule>> _waitingModules = new();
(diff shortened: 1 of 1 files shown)


ArthurPxC:
Makes sense for me.

ArthurPxC (approved):

andreniggemann (approved):
This would fix this problem.
Just as an idea: Wouldn't it be better to make StartModule resilient against multiple overlapping requests?
This would solve this entire class of problems, because theoretically we could also have issues when somebody from the outside calls IModuleManager.StartAsync in parallel. 
It would have the additional benefit of making it more easily testable, because you can just call StartModuleAsync in parallel multiple times, right?

dbeuchler:
> This would fix this problem. Just as an idea: Wouldn't it be better to make StartModule resilient against multiple overlapping requests? This would solve this entire class of problems, because theoretically we could also have issues when somebody from the outside calls IModuleManager.StartAsync in parallel. It would have the additional benefit of making it more easily testable, because you can just call StartModuleAsync in parallel multiple times, right?

Good idea, moved to new issue: https://github.com/PHOENIXCONTACT/MORYX-Framework/issues/1514

PR #1520, opened 2026-09-18
Add `NullIdentity` and `CombinedIdentity`

1nf0rmagician:
- Adds a `NullIdentity` to cleanly reflect, that an `IIdentifiableObject` can not be identifiable (for a period of time) after all
- Adds `CombinedIdentity` to cleanly reflect, that an `IIdentifiableObject` can hold multiple singular `IIdentity` instances, e.g. if a product instance has a serial number and a mac-adress
- Adds `IdentifierExtensions` to smoothen developer experience with the new feature while using `IIdentifiableObject` and `IIdentity`
- Adds additional documentation for the identity concept


1nf0rmagician:
FYI: @szThunder

src/Moryx.AbstractionLayer/Identity/CombinedIdentity.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </remarks>
+    public static IIdentity From(IEnumerable<IIdentity> identities)
+    {
+        if (identities == null || !identities.Any())
+        {
+            return NullIdentity.Instance;
+        }
+
+        var flat = new List<IIdentity>();
+        foreach (var identity in identities)
```
dbeuchler, line 49: multiple enumerations of `identities`

src/Moryx.AbstractionLayer/Identity/CombinedIdentity.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                flat.Add(identity);
+            }
+        }
+
+        return flat.Count switch
+        {
+            0 => NullIdentity.Instance,
+            1 => flat[0],
+            _ => new CombinedIdentity(flat)
```
dbeuchler, line 69: So a CombinedIdentity.From returns an Null or specific or Combined? Should we move it better to an extension?

src/Moryx.AbstractionLayer/Identity/CombinedIdentity.cs
```diff
@@ -0,0 +1,119 @@
...
+    public IReadOnlyList<IIdentity> GetAll() => _identities;
+
+    /// <summary>
+    /// Checks whether the given <paramref name="candidate"/> can identify this object,
+    /// i.e., all of this' identities are present in the candicate (<code>this</code> ⊆ <paramref name="candidate"/>).
+    /// </summary>
+    public virtual bool MatchedBy(IIdentity candidate) => candidate is CombinedIdentity combinedCandidate
+        ? combinedCandidate.Matches(this) : EqualsSingleIdentity(candidate);
+
+    private bool EqualsSingleIdentity(IIdentity other) => _identities.Count == 1 && _identities.Single().Equals(other);
```
dbeuchler: Instead of Linq.Single you should use indexer here. Just an performance optimization.

src/Tests/Moryx.AbstractionLayer.Tests/CombinedIdentityTests.cs
```diff
@@ -0,0 +1,320 @@
...
+    [Test]
+    public void GetAll_WithSameType_OrderedAlphabeticallyByIdentifier()
+    {
+        // Arrange
+        var z = new BatchIdentity("Z");
+        var a = new BatchIdentity("A");
+        var m = new BatchIdentity("M");
+
+        // Act
+        var sut = CombinedIdentity.From(z, m, a);
```
dbeuchler: what is `sut`? I would prefere a readable name instead of abbreviations. `combinedIdentity` would be readable. In other tests you used `result` which is also okay for me.

src/Tests/Moryx.AbstractionLayer.Tests/IdentityExtensionsTests.cs
```diff
@@ -0,0 +1,317 @@
...
+    [
+        new BatchIdentity("A"),
+        CombinedIdentity.From(new BatchIdentity("A"), new BatchIdentity("B")),
+    ];
+
+    [TestCaseSource(nameof(NonNullIdentityKinds))]
+    public void Matches_WithNullIdentityRequired_ReturnsTrue(IIdentity identitiy)
+    {
+        // requiring nothing is satisfied by any identity
+        Assert.That(identitiy.Matches(NullIdentity.Instance));
```
dbeuchler: Could you generally split `arrange`, `act` and `assert`? What is asserted here? Matches returns true/false? Is there an Is.True missing?

You shold split them to make it readable, show what exactly is tested and most important: make it easier debuggable.

dbeuchler (changes requested):

src/Moryx.AbstractionLayer/Identity/CombinedIdentity.cs
1nf0rmagician: > So a CombinedIdentity.From returns an Null or specific or Combined

Yes, the general idea is that 
```csharp
// Assure the created Identity Types actually match the composition semantically
CombinedIdentity.From(null).GetType() == typeof(NullIdentity)
CombinedIdentity.From(BatchIdentity("A")).GetType() == typeof(BatchIdentity)
CombinedIdentity.From(BatchIdentity("A"), BatchIdentity("A")).GetType() == typeof(CombinedIdentity)
```

>Should we move it better to an extension?

I decided to name the corresponding extension method 'Combine', as it is a lot nicer to read this way. Also the static factory method on the `CombinedCapability` type allows to keep the public ctor hidden which would otherwise allow to circumvent the statement above.

```csharp
// Combining Identities with fluent API
var singleIdentity = new BatchIdentity("A"); // typeof(BatchIdentity)
var combinedIdentity = singleIdentity.Combine(BatchIdentity("B")); // typeof(CombinedIdentity)

// While 'From' does not really make sense, when you already have an identity
var combinedIdentity = singleIdentity.From(BatchIdentity("B")); // typeof(CombinedIdentity)
```

src/Tests/Moryx.AbstractionLayer.Tests/IdentityExtensionsTests.cs
1nf0rmagician: I did another general overhaul of the tests, reducing all the duplicated instantiations. I (wanted to) use the visible split for `arrange`, `act` and `assert` where the tests are not a one-liner and also use the respective comments in these cases

1nf0rmagician:
Let's squash this on merge @dbeuchler

PR #1525, opened 2026-09-22
Draft: runtime possible values

### LINKED ISSUE #1521: Possible Runtime values
I want to be able to control somehow the possible values at runtime. One usage could be like the following

```csharp
public List<string> Values {get; set;}

[MyRuntimePossibleValue(Source = nameof(Values ))]
public string SelectedValue {get; set;}
```
This would make the content of `Values`  populate the  `PossibleValues` of the Selected Value in the UI.

The second use case would like the following:

```csharp
public string Paramater {get; set;}

[AnotherRuntimePossibleValue(Source = nameof(Paramater))]
public string SelectedValue {get; set;}
```
Inside the `AnotherRuntimePossibleValue.Get` method i can use the value of Parameter to fetch for example a resource(s) from the `IContainer` or ``IServiceProvider`.

A custom `PossibleValueAttribute` that offers access to the current instance inside the Parse and Get method. And a `CustomPossibleValue` Serializer that provide the custom `PossibleValueAttribute` with current instance.

**Acceptance criteria**

- [ ] Setting a list property as the source should display the value in the list as possible values for the original property

MathoMathiasCamara:
This provide initial idea on the Runtime Possible Values that is mentioned in #1521. I contains the abstract `RuntimePossibleValuesAttribute` and the `RuntimePossibleValuesSerialization`.

close #1521.


dbeuchler:
@MathoMathiasCamara ready for review?

MathoMathiasCamara:
@dbeuchler, @1nf0rmagician   This is ready for review.

